PR #10, opened 2021-11-23
NSW=1 --> NSW=0 in static, Cf in POTCARs

arosen93:
## Quick Hello
Hi @utf! I'm opening this very minor PR in part to say hello and that I think `atomate2` looks fantastic! I am really excited to start using it. Also a few small things.

## Summary of Proposed Changes
- I added `Cf: Cf` to the list of default pseudopotentials since it's available with the VASP PAW_PBE_54 potentials. It probably won't be relevant to anyone, but why not aim for completeness. You got all the other ones!
- I changed `NSW=1` to `NSW=0` in your `get_incar_updates()` for `StaticSetGenerator()`. Was there a specific reason to use `NSW=1` for the static calculation? One of my concerns is that other codes might look at `NSW` and think that it's doing a relaxation even though it's a static calculation.
    - Take the `DriftErrorHandler()` [in Custodian](https://github.com/materialsproject/custodian/blob/master/custodian/vasp/handlers.py#L679), for example. It checks to see if `EDIFFG >= 0 or NSW == 0`. If this condition is met, it doesn't run the drift error checker; otherwise, it does. However, here, `NSW=1` would make Custodian think it's a relaxation. Granted, Custodian is unlikely to actually do anything here, but it's more the premise that this might influence other codes in unexpected ways.


utf:
Hi @arosen93, great to have someone looking over atomate2. It is obviously still forming so any and all suggestions/PR are very welcome.

Thanks for catching the Cf pseudo potential. As for `NSW = 1` this was a on purpose, as certain calculations require it (e.g., DFPT `LEPSILON` needs `NSW = 1` and won't work with `NSW = 0`). That said, this does cause an incompatibility with custodian and that seems more difficult to fix than just setting `NSW = 1` for certain calculation types.

Before I merge, can you add `NSW = 1` to the custom `lepsilon` settings in `StaticSetGenerator`?

arosen93:
@utf, interesting! I didn't know that about DFPT. I see the thought process there, although my personal opinion is that it's better to set `NSW=1` on an as-needed basis. Just changed the `lepsilon` settings.

arosen93:
I figured that might break something along the way. Sorry for not running tests locally. I'll get the testing stuff set up tomorrow and can fix then. :)

utf:
Ah, the test data is now inconsistent. Should be as simple as editing the INCAR in the `Si_bandstructure/static/inputs/INCAR` file. 

Also, to be consistent, we should probably also adjust NSW in the HSEStaticSetGenerator.

If I have a chance, I can take a look at this later.

arosen93:
Hi again @utf. That should do it now. Tests passed locally. 

I had to change the INCARs in the following:
```
`Si_elastic/elastic_relax...`
`Si_hse_band_structure/hse_static`
`Si_band_structure/static`
```

utf:
For the lint error. The easiest thing to do is to run `pre-commit install` when in the root atomate2 directory. Then, anytime you make a commit, any staged files will get linted automatically.

arosen93:
@utf: I guess I must have run the wrong `pytest` locally. I'm having some trouble catching where `NSW = 1`'s keep appearing in the INCAR files...

PR #12, opened 2021-11-24
Add some missing functionals

arosen93:
- I changed `MBJL` to `MBJ`. I think MBJ is the [supported keyword](https://www.vasp.at/wiki/index.php/METAGGA). Let me know if I'm wrong here.
- I added `RSCAN` and `R2SCAN` to the METAGGAs.
- I renamed `revPBE-vdW` to `vdW-DF` as suggested by the [VASP manual](https://www.vasp.at/wiki/index.php/VdW-DF_functional_of_Langreth_and_Lundqvist_et_al.). Happy to change this back if you think it's important to emphasize the revPBE part.
- I added `BEEF-vdW` since it's commonly used. Not sure what your thoughts are on having defaults set for "third-party" functionals.

### DIFF
--- a/src/atomate2/vasp/schemas/calc_types/run_types.yaml
+++ b/src/atomate2/vasp/schemas/calc_types/run_types.yaml
@@ -54,14 +54,18 @@ HF:
 METAGGA:
   M06L:
     METAGGA: M06L
-  MBJL:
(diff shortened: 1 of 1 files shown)


utf:
Great, thank you very much!

PR #17, opened 2021-11-27
Fix powerups in docs

arosen93:
In the [docs](https://materialsproject.github.io/atomate2/user/codes/vasp.html#modifying-input-sets), the powerups aren't actually applied because they are not done in-place. I've fixed this in the PR (and fixed the formatting in the rest of the `.rst`). 

Apologies that I keep having to go back to fix linting errors. For some reason, I am getting the error below when I try to do any commits, and I haven't figured out why.
```bash
Check Yaml...........................................(no files to check)Skipped
Fix python encoding pragma...........................(no files to check)Skipped
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
autoflake............................................(no files to check)Skipped
black................................................(no files to check)Skipped
blacken-docs.............................................................Passed
isort (python).......................................(no files to check)Skipped
flake8...............................................(no files to check)Skipped
type annotations not comments........................(no files to check)Skipped
rst ``code`` is two backticks............................................Passed
rst directives end with two colons.......................................Passed
rst ``inline code`` next to normal text..................................Passed
mypy.................................................(no files to check)Skipped
error: cannot spawn null: Exec format error
error: gpg failed to sign the data
fatal: failed to write commit object
```


utf:
Fantastic! I made that change (previously modifications were performed in place) but forgot to update the documentation.

PR #23, opened 2021-12-03
LREAL = False in static jobs by default

arosen93:
## Summary
Sets `LREAL = False` in `StaticSetGenerator()` and `HSEStaticSetGenerator()` for more accurate energies.


utf:
Thanks!

PR #28, opened 2021-12-05
Change reciprocal_density type from float to int

arosen93:
This is pretty self-explanatory. Changes the k-point density types to `int`, which is consistent with pymatgen. Just doing this as a PR now so I don't forget.

### DIFF
--- a/src/atomate2/vasp/sets/core.py
+++ b/src/atomate2/vasp/sets/core.py
@@ -199,7 +199,7 @@ class NonSCFSetGenerator(VaspInputSetGenerator):
 
     mode: str = "line"
     dedos: float = 0.005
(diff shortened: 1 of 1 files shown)


utf:
I've never understood this myself. Is there any reason these can't be floats? If not, I think it might be best to update pymatgen instead.

arosen93:
You know, I think you're right here. I didn't really think much of it before.

For reciprocal density (i.e. normalized by volume), pymatgen does `kppa = kppvol * vol * structure.num_sites`. Clearaly, `vol` is a `float`, so there's no need for `kppvol` to be `int`. I'll close this because I do think this is just a nuance with pymatgen.

PR #31, opened 2021-12-10
Add missing LASPH flags on vdW functionals

arosen93:
Very minor changes:
1. I hadn't noticed that you kept `LASPH = True` for the vdW functionals, so I added this flag for BEEF-vdW.
2. optB88-vdW was missing the `LASPH = True` flag, so I added it in. It's mentioned in the [VASP manual](https://www.vasp.at/wiki/index.php/VdW-DF_functional_of_Langreth_and_Lundqvist_et_al.).


arosen93:
@utf, I think the testing is currently broken.

Issue #30, opened 2021-12-09
Feature: Support for different makers in the DoubleRelaxMaker


arosen93:
In the `DoubleRelaxMaker` flow, currently the same `relax_maker` is used for step 1 and step 2. It would be desirable to split this up into two makers (which by default are the same) in case the user wants to do something like a half k-point relaxation for step 1.

I can take care of this modification. This is mainly a reminder for myself.

utf:
Closed with #32

PR #32, opened 2021-12-10
Support two different makers for DoubleRelaxMaker

arosen93:
## Summary
Addresses #30, wherein the `DoubleRelaxMaker` flow can now take in two separate `BaseVaspMaker` args in case the user wants the first or second relax job to have different settings. A common use-case could be to set `user_kpoint_settings` in the first relax maker to be half the usual value. Ideally, there'd be an optional parameter called `half_kpts_first_relax` that could do this automatically, but I had some trouble figuring out the best way to modify the k-point density of the first job in a clean way that accounts for `KSPACING` vs. the other k-point generation schemes.

Edit: In hindsight, I guess the user could always have just defined the flow first and then edited the desired `job.maker.input_set_generator` afterwards... hmm...


utf:
I think this is a reasonable approach to take. Am I ok to merge?

utf:
I actually think `relax_maker1` and `relax_maker2` could be better than  `relax_maker` and `relax_maker2`. Just for style and consistency.

arosen93:
Got it! I'll make that consistent throughout (eg in the tests too).

arosen93:
Also, sorry about the hacky commits here. I'm coding in quick bursts at the airport 😅 I'll get it sorted out and ready to merge soon though.

src/atomate2/vasp/flows/core.py
```diff
@@ -42,12 +42,16 @@ class DoubleRelaxMaker(Maker):
...
+        Maker to use to generate the first relaxation.
+    relax_maker2 : .BaseVaspMaker
+        Maker to use to generate the second relaxation. If not set,
+        then relax_maker2 = relax_maker1.
     """
 
     name: str = "double relax"
-    relax_maker: BaseVaspMaker = field(default_factory=RelaxMaker)
+    relax_maker1: BaseVaspMaker = field(default_factory=RelaxMaker)
+    relax_maker2: BaseVaspMaker = relax_maker1
```
utf, line 54: I'm confused that this works? As I thought relax_maker1 is a class variable of the Maker and not available in global scope. Although maybe this is a data class trick?

At the very least, does this not mean that relax_maker1 and relax_maker2 are the same object. So if you update one then you update the other?

I think the two options would be:
1. Make a deepcopy of the relax_maker1.
2. Set relax_maker2 to None and then use relax_maker1 if relax_maker2 is not set.

utf: When I try just a simple test case, this gives weird results:

```python
from dataclasses import dataclass, field

@dataclass
class Test:
    a: list = field(default_factory=lambda: [1, 2, 3])
    b: list = a
    
Test().a
# Field(name='b',type=<class 'list'>,default=<dataclasses._MISSING_TYPE object at 0x1037cd4f0>,
#       default_factory=<function Test.<lambda> at 0x15a28c310>,init=True,repr=True,hash=None,
#       compare=True,metadata=mappingproxy({}),_field_type=_FIELD)
```

Do the tests definitely work on your system?

utf: Having thought about this a bit more. I think the best behaviour would be just to have two separate makers. I.e., if you update the first relax maker that will have no effect on the second relax maker. This is how nested makers work in all other Flows so probably best to be consistent here.

We could add a utility class method to DoubleRelaxMaker called `from_relax_maker(relax_maker)`. This would allow you to initialise a double relaxation with the same relax_maker used for both relaxations. I think then most of the existing workflows that use the double relax maker should then use this utility function.

arosen93: > When I try just a simple test case, this gives weird results:
> 
> ```python
> from dataclasses import dataclass, field
> 
> @dataclass
> class Test:
>     a: list = field(default_factory=lambda: [1, 2, 3])
>     b: list = a
>     
> Test().a
> # Field(name='b',type=<class 'list'>,default=<dataclasses._MISSING_TYPE object at 0x1037cd4f0>,
> #       default_factory=<function Test.<lambda> at 0x15a28c310>,init=True,repr=True,hash=None,
> #       compare=True,metadata=mappingproxy({}),_field_type=_FIELD)
> ```
> 
> Do the tests definitely work on your system?

Sorry for going AWOL on this one. I had trouble getting the test suite working (have to go back and fix that) and was relying on the CI for what I thought would be a quick fix. Turns out I was a bit off the mark here (only started writing my own `Maker`very recently). Apologies for the lack of clarity.

src/atomate2/vasp/flows/amset.py
```diff
@@ -365,7 +365,7 @@ class HSEVaspAmsetMaker(Maker):
     temperatures: tuple[float, ...] = _DEFAULT_TEMPERATURES
     amset_settings: dict = field(default_factory=dict)
     relax_maker: BaseVaspMaker | None = field(
-        default_factory=lambda: DoubleRelaxMaker(relax_maker1=HSETightRelaxMaker())
+        default_factory=lambda: DoubleRelaxMaker.from_relax_maker(TightRelaxMaker())
```
utf, line 368: This should be: HSETightRelaxMaker

src/atomate2/vasp/flows/core.py
```diff
@@ -79,6 +80,12 @@ def make(self, structure: Structure, prev_vasp_dir: str | Path | None = None):
 
         return Flow([relax1, relax2], relax2.output, name=self.name)
 
+    @classmethod
+    def from_relax_maker(cls, relax_maker: BaseVaspMaker):
```
utf, line 84: Please can you add a docstring?

utf (commented):
Sorry, you might still be editing but noticed a couple of things.

src/atomate2/vasp/flows/core.py
arosen93: Docstring just added!

arosen93:
@utf, yup I'm still editing. I pushed the change because I have some questions. The tests fail due to the LREAL assertion, and I wanted to ask if you could try to help me understand why. Definitely a (useful) learning process!

src/atomate2/vasp/flows/amset.py
arosen93: Thanks! Yup, caught that one too.

tests/vasp/flows/test_core.py
```diff
@@ -15,8 +17,8 @@ def test_double_relax(mock_vasp, clean_dir, si_structure):
 
     # settings passed to fake_run_vasp; adjust these to check for certain INCAR settings
     fake_run_vasp_kwargs = {
-        "relax 1": {"incar_settings": ["NSW", "ISMEAR"]},
-        "relax 2": {"incar_settings": ["NSW", "ISMEAR"]},
+        "relax 1": {"incar_settings": ["NSW", "ISMEAR", "LREAL"]},
```
utf, line 20: The tests are failing because of this line. You are validating on `LREAL` which will compare the written files vs reference files.

The reference files contain `LREAL = Auto` but some of your calculations contain `LREAL = False` so they don't match.

Better would be to remove the check here, and manually check the task document as you are doing below.

arosen93: Thanks for pointing that out. Now, `LREAL` is getting set to `True` somewhere, which is... peculiar... seeing that the default is 'Auto'. I'm digging into that now.

arosen93: Can I check the task doc if the mock vasp run is copying over the INCAR? Wouldn't the `orig_inputs` just be pulled from that INCAR then? Similarly, `.input.parameters` is from the `vasprun.xml` which of course is not from the modified VASP run. 

I'm trying to find out where I can check for the new input settings.

utf: Ah sorry this is getting so complicated. Yeah, I see what you mean. There is no easy way to validate the written inputs without creating a new directory of reference test data.

I guess the best thing to do is to create a new folder of test data and symlink the files that are the same between the two folders (to save space), and just include a new INCAR file with your changes.

Sorry for the hassle!

arosen93: Well, I just feel validated that I understood what was going on! I'll take it!

That's more or less what I figured. I'll give that a go in the morning over here in the states.

arosen93:
@utf, okay the tests finally pass now. However, could I ask you to take care of the symlinking? I am on Windows and am worried that it's going to wreak havoc on things. The only files that need to be kept as an actual file are the `.../inputs/INCAR` in the `Si_double_relax_swaps` subdirectories. The rest aren't used by the test suite.

utf:
Fantastic, thank you!

PR #42, opened 2021-12-29
Custodian FULL_OPT fix

arosen93:
Previously, specifying the FULL_OPT job type for Custodian would never actually run because of a typo (see below). Now fixed.

https://github.com/materialsproject/atomate2/blob/6ef1188d9e1e038a6d964d8e9d9177c8e54727fd/src/atomate2/vasp/run.py#L147-L150


utf:
Thanks!

PR #54, opened 2022-01-17
Create molecule schema

arosen93:
I needed to make a pymatgen `Molecule` schema for my own project and figured it should just be integrated in Atomate2 directly.  Is something like this of interest? I'm not very familiar with Atomate1, so I'm not sure if there is a `Molecule` schema there I should copy, but I just copied the format of the `Structure` schema and adjusted it slightly.

I'll add tests and complete this shortly. Apologies about my other open PR. Even though it's super simple, time got the best of me... will wrap that up soon too.


arosen93:
@utf, this should be good to go, assuming this is how you'd want the code organized.

utf:
Looks great. Thanks very much.

PR #65, opened 2022-01-30
[WIP] Add mag_density to VASP output doc

arosen93:
This PR adds a `mag_density` attribute to the VASP `CalculationOutput` document. This can sometimes be more useful to query by than `total_mag` because the former is invariant to the # of unit cells.

### DIFF
--- a/src/atomate2/vasp/schemas/calculation.py
+++ b/src/atomate2/vasp/schemas/calculation.py
@@ -307,6 +307,10 @@ class CalculationOutput(BaseModel):
     transition: str = Field(
         None, description="Band gap transition given by CBM and VBM k-points"
     )
+    mag_density: float = Field(
+        None,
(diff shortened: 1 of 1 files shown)


arosen93:
I just need to point the test to the following data directory: `atomate2/tests/test_data/vasp/magnetic_run/`. I'm not sure the ideal way to do that given the current way Atomate2 is set up. Please let me know.

https://github.com/arosen93/atomate2/blob/d613db0b3dfcd64282be8bdc7e9d52a14309cceb/tests/vasp/schemas/test_calculation.py#L86-L95

arosen93:
Thanks! That was easy!

utf:
Is this still a WIP?

arosen93:
Thanks for asking! Nope. My bad!

utf:
Great thanks very much!

PR #66, opened 2022-01-30
Store efermi in CalculationOutput

arosen93:
This PR stores the `efermi` from the `vasprun.xml`. It is otherwise always left at `None` because it wasn't assigned in `electronic_output`.

Example:
```python
from atomate2.vasp.schemas.calculation import CalculationOutput
from pymatgen.io.vasp.outputs import Vasprun, Outcar

# run this in, e.g. one of the test double-relax runs
c = CalculationOutput.from_vasp_outputs(Vasprun('vasprun.xml.gz'),Outcar('OUTCAR.gz'))
```

### DIFF
--- a/src/atomate2/vasp/schemas/calculation.py
+++ b/src/atomate2/vasp/schemas/calculation.py
@@ -382,6 +382,7 @@ class CalculationOutput(BaseModel):
             bandstructure = vasprun.get_band_structure(efermi="smart")
             bandgap_info = bandstructure.get_band_gap()
             electronic_output = dict(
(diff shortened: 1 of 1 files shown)


arosen93:
On a side-note, in [this failed test](https://github.com/materialsproject/atomate2/runs/5000245327?check_suite_focus=true#step:7:55) the efermi from `Vasprun.efermi` is 6.23352981 but from `vasprun.get_band_structure(efermi='smart')` is 5.9249. Pretty big difference! Is this normal, @utf?

PR #71, opened 2022-02-05
Fix a few typos in the docs

arosen93:
Just cleaning up a few things in the documentation:
- Fixes some spelling mistakes I previously introduced in the VASP docs
- Added `docs_build/*` to `.gitignore` because that's the destination for the docs build folder when following the developer instructions 
- Fixed the ORCiD link (thanks for the credit!)


utf:
This is great. Thanks!

PR #75, opened 2022-02-07
Fix test suite for pymatgen vasprun fix

arosen93:
Updates test suite with correct energies so that pymatgen can be updated. Fixes (and can close) the error with #72.


utf:
Fantastic. Thank you very much.

PR #83, opened 2022-03-09
potcar spec kwarg

jmmshn:
## Unused potcar_spec kwarg

It looks like there is an `potcar_spec` kwarg here
https://github.com/materialsproject/atomate2/blob/e7bfdb93cc3748c3465685fffd6efac8eba7257b/src/atomate2/vasp/sets/base.py#L73
which might be useful for live debugging, but `write_vasp_input_set` cannot use it since it breaks the input set creation here:
https://github.com/materialsproject/atomate2/blob/f319382ef47f4d4385d131cb2f9edefc24af83bf/src/atomate2/vasp/files.py#L171

This should allow you to write input files but it still looks like the KPOINTS file is missing.

Is this the expected behavior?
```
from atomate2.vasp.sets.core import RelaxSetGenerator
from atomate2.vasp.files import write_vasp_input_set

mgo_structure = Structure(
    species=["Mg", "O"],
)
gen = RelaxSetGenerator()
write_vasp_input_set(mgo_structure, input_set_generator=gen, potcar_spec=True)

### creates inputfile but missing KPOINTS
```

### DIFF
--- a/src/atomate2/vasp/files.py
+++ b/src/atomate2/vasp/files.py
@@ -146,6 +146,7 @@ def write_vasp_input_set(
     input_set_generator: VaspInputSetGenerator,
     from_prev: bool = False,
     apply_incar_updates: bool = True,
+    potcar_spec: bool = False,
     clean_prev: bool = True,
     **kwargs,
 ):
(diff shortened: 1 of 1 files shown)


jmmshn:
Oh nvm I was not aware of this VASP flag
https://www.vasp.at/wiki/index.php/KSPACING

utf:
Thanks! Good spot.

utf:
I just noticed you didn't add docstrings for the new kwargs. Would you be able to add them in please?

PR #92, opened 2022-03-18
Replace monty which with shutil which

arosen93:
## Summary
`monty.os.path.which` is deprecated and will be removed in v2023. Replaced with `shutil.which`.


utf:
Thanks!

PR #94, opened 2022-03-27
Fix `calculate_deformation_potentials()`

janosh:
I was looking through the code trying to get a feel for the new atomate. 😄 

Noticed the default `ibands` value in `calculate_deformation_potentials()` will raise

```py
TypeError: 'NoneType' object is not iterable
```

https://github.com/materialsproject/atomate2/blob/c3d3fcd3e6edb8bb069cb0c23193ce1f356ed6b7/src/atomate2/vasp/jobs/amset.py

Happy to fix as part of this PR. Should `ibands` become a required arg?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # atomate2
 
-<a href="https://github.com/materialsproject/atomate2/actions?query=workflow%3Atesting"><img alt="code coverage" src="https://img.shields.io/github/workflow/status/materialsproject/atomate2/testing?label=tests"></a>
-<a href="https://codecov.io/gh/materialsproject/atomate2/"><img alt="code coverage" src="https://img.shields.io/codecov/c/gh/materialsproject/atomate2"></a>
-<a href="https://pypi.org/project/atomate2"><img alt="pypi version" src="https://img.shields.io/pypi/v/atomate2?color=blue"></a>
-<img alt="supported python versions" src="https://img.shields.io/pypi/pyversions/atomate2">
+[![code coverage](https://img.shields.io/github/workflow/status/materialsproject/atomate2/testing?label=tests)](https://github.com/materialsproject/atomate2/actions?query=workflow%3Atesting)
+[![code coverage](https://img.shields.io/codecov/c/gh/materialsproject/atomate2)](https://codecov.io/gh/materialsproject/atomate2)
+[![pypi version](https://img.shields.io/pypi/v/atomate2?color=blue)](https://pypi.org/project/atomate2)
+![supported python versions](https://img.shields.io/pypi/pyversions/atomate2)
 
 **👉 [Full Documentation][docs] 👈**
 
@@ -70,7 +70,7 @@ run_locally(bandstructure_flow, create_folders=True)
 ```
 
 In this example, we run execute the workflow immediately. In many cases, you might want
-to perform calculations on several materials simulatenously. To achieve this, all
+to perform calculations on several materials simultaneously. To achieve this, all
 atomate2 workflows can be run using the [FireWorks] software. See the
 [documentation][atomate2_fireworks] for more details.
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -1,9 +1,9 @@
 # atomate2
 
-<a href="https://github.com/materialsproject/atomate2/actions?query=workflow%3Atesting"><img alt="code coverage" src="https://img.shields.io/github/workflow/status/materialsproject/atomate2/testing?label=tests"></a>
-<a href="https://codecov.io/gh/materialsproject/atomate2/"><img alt="code coverage" src="https://img.shields.io/codecov/c/gh/materialsproject/atomate2"></a>
-<a href="https://pypi.org/project/atomate2"><img alt="pypi version" src="https://img.shields.io/pypi/v/atomate2?color=blue"></a>
-<img alt="supported python versions" src="https://img.shields.io/pypi/pyversions/atomate2">
+[![code coverage](https://img.shields.io/github/workflow/status/materialsproject/atomate2/testing?label=tests)](https://github.com/materialsproject/atomate2/actions?query=workflow%3Atesting)
+[![code coverage](https://img.shields.io/codecov/c/gh/materialsproject/atomate2)](https://codecov.io/gh/materialsproject/atomate2)
+[![pypi version](https://img.shields.io/pypi/v/atomate2?color=blue)](https://pypi.org/project/atomate2)
+![supported python versions](https://img.shields.io/pypi/pyversions/atomate2)
```
utf, line 6: Is there a `[` missing at the start of this line?

README.md
```diff
@@ -125,7 +125,6 @@ A full list of all contributors can be found [here][contributors].
 [changelog]: https://materialsproject.github.io/atomate2/user/changelog.html
 [installation]: https://materialsproject.github.io/atomate2/user/install.html
 [contributing]: https://materialsproject.github.io/atomate2/user/contributing.html
-[contributors]: https://materialsproject.github.io/atomate2/user/contributors.html
```
utf, line 128: Why remove the contributors page?

src/atomate2/common/schemas/cclib.py
```diff
@@ -297,7 +297,7 @@ def cclib_calculate(
 
     if method in cube_methods and not cube_file:
         raise FileNotFoundError(
-            f"A cube file must be provied for {method}. Returning None."
+            f"A cube file must be provide for {method}. Returning None."
```
utf, line 300: I think it should actually be "provided"

setup.cfg
```diff
@@ -0,0 +1,8 @@
+[isort]
```
utf, line 1: I'm loathe to introduce a new config file. Is there is a reason for defining these in `setup.cfg` as well as in `.pre-commit.yaml`?

utf (commented):
Great catch about `calculate_deformation_potentials`. The code should not add the `--bands` to the command line runner if `ibands` is `None`.

And thanks for your other edits! I just had a couple of questions.

setup.cfg
janosh: I definitely understand your sentiment. I don't like config files either, especially not root-level ones. The reason for this file is that CLIs and IDEs don't read config from `.pre-commit-config.yaml`. So if you enabled format on save in your editor or you use `flake8` or `mypy` or `isort` directly instead of through `pre-commit run mypy|isort|...`, they would run with their default settings and make changes that would be reverted by `pre-commit`. E.g. try `flake8 src` in a terminal. It complains about 100 of lines too long because it doesn't know about the line length set in `.pre-commit-config.yaml`.

Like I said, I understand your sentiment so if you think this is insufficient reason, I'll of course delete the file.

src/atomate2/common/schemas/cclib.py
janosh: Good catch.

README.md
janosh: Oops, that was by mistake. Got a bit late last night.

README.md
janosh: No, the last badge is not wrapped by a link, hence no leading bracket.

setup.cfg
janosh: It does work the other way round btw. Settings here will also be picked up when running these tools through pre-commit. So we could remove the config from `.pre-commit-config.yaml` to keep a SSOT.

utf: Yeah, that would be preferable!

janosh:
I think this is ready.

utf:
Brilliant, thanks very much.

PR #101, opened 2022-03-29
Quick start guide

janosh:
This is still a WIP. I was reading the [Quick Start](https://materialsproject.github.io/atomate2/user/index.html#quick-start) guide which is very well done! Props for starting out with good docs! 👍 

I think just showing this code snippets could lead to misunderstanding for new users though. People might just run `pip install atomate2` and try to run the snippet, not knowing that it requires

- setting `PMG_VASP_PSP_DIR` in `~/.pmgrc.yaml` to a directory with VASP POTCAR files
- telling `atomate2` where to find your VASP binary
- and having a MongoDB ready to accept incoming connections.

Encountering unexpected errors early on can be discouraging so maybe best to list these requirements immediately after the snippet?

```python
from atomate2.vasp.flows.core import RelaxBandStructureMaker

# construct a rock salt MgO structure
mgo_structure = Structure(
    species=["Mg", "O"],
)

# make a band structure flow to optimise the structure and obtain the band structure
bandstructure_flow = RelaxBandStructureMaker().make(mgo_structure)

# run the job
run_locally(bandstructure_flow, create_folders=True)
```

### DIFF
--- a/docs/src/user/install.rst
+++ b/docs/src/user/install.rst
@@ -157,8 +157,8 @@ Create a conda environment
 
 .. note::
 
-   Make sure to create a Python 3.7+ environment as recent versions of atomate2 only
-   support Python 3.7 and higher.
+   Make sure to create a Python 3.8+ environment as recent versions of atomate2 only
+   support Python 3.8 and higher.
 
 We highly recommended that you organize your installation of the atomate2 and the other
 Python codes using a conda virtual environment. Some of the main benefits are:


utf:
I think it's definitely a good idea to note that some setup will be required to run the script!

README.md
```diff
@@ -69,6 +71,14 @@ bandstructure_flow = RelaxBandStructureMaker().make(mgo_structure)
 run_locally(bandstructure_flow, create_folders=True)
 ```
 
+Before the above code can run successfully, you'll need to
+
+- set `PMG_VASP_PSP_DIR` in `~/.pmgrc.yaml` to a directory with VASP POTCAR files,
+- tell atomate2 through `ATOMATE2_VASP_CMD=/path/to/vasp_std` where to find your VASP binary,
+- have a MongoDB instance ready to accept job output.
+
+See the [installation] steps for details how to set all of this up.
```
janosh, line 80: How's this? The flow from this line to the next paragraph is a bit awkward maybe? But wanted to keep it short.

utf: Could you remove the `ATOMATE2_VASP_CMD=/path/to/vasp_std` bit or change it to `export ATOMATE2_VASP_CMD=/path/to/vasp_std`.

Thanks!

utf:
Great, thanks!

arosen93:
Okay, so feel free to ignore because nobody asked for my opinion :P 

1. I liked the cleaner, minimalist look of the original README better because it shows off how simple it is to use atomate2 and nothing more. The README is meant to get the passer-by excited about using the code, not necessarily get up and running with it. I think just having text that said "To set up Atomate2 to run workflows like this, see the installation instructions." After all, one day (hopefully soon) Atomate2 will not just be VASP-focused, and other users won't care about those details.
2. The paragraph on using FireWorks is nice, but I know many people are overwhelmed by it. After all, you can run multiple jobs simultaneously by using `run_locally` and a job scheduler. FireWorks is also mentioned in the "Tutorials" section.

None of this is important. Feel free to keep everything as-is.

janosh:
> Okay, so feel free to ignore because nobody is asking for my opinion here, but... :P

I hereby issue a standing invitation to for all future occasions to please share your opinion! 😆 

> After all, one day (hopefully soon) Atomate2 will not just be VASP-focused, and other users won't care about those details.

That's a fair point which I hadn't considered. Though as long as the example is VASP-focused, I guess the setup instructions need to be as well? Once there's a 2nd example for another code, it could have its own setup bullet points.

In any case, I'm okay with removing again. Maybe just a link to `[installation]` like @arosen93 suggests is enough. I was definitely expecting something like that when reading the docs and thinking this can't be all to start running jobs.

PR #104, opened 2022-03-31
Refactor `test_copy_vasp_outputs` to use builtin `tmp_path` + `monkeypatch` fixtures

### LINKED ISSUE #103: Use `pytest` `tmp_path` fixture
Boy am I glad to see you're using `pytest`. 😄 

I noticed you wrote your own `tmp_dir` fixture.

https://github.com/materialsproject/atomate2/blob/614ace0ee657759091b6e39296614b07b42140c8/tests/conftest.py#L46-L58

`pytest` ships with a bunch of builtin fixtures including [`tmp_path`](https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path) which gives a `pathlib.Path` (and [`tmpdir`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.tmpdir.tmpdir) which gives a `py.path.local` though that's scheduled for deprecation).

Happy to submit a PR to use the builtin (unless I'm missing some special requirement).

janosh:
Closes #103.

### DIFF
--- a/tests/vasp/test_files.py
+++ b/tests/vasp/test_files.py
@@ -9,11 +9,14 @@ import pytest
         ({"additional_vasp_files": ("CHGCAR",)}, ("POSCAR", "INCAR", "CHGCAR")),
     ],
 )
-def test_copy_vasp_outputs_static(vasp_test_dir, tmp_dir, copy_kwargs, files):
+def test_copy_vasp_outputs_static(
+    vasp_test_dir, tmp_path, monkeypatch, copy_kwargs, files
+):
     from pathlib import Path
 
     from atomate2.vasp.files import copy_vasp_outputs
 
+    monkeypatch.chdir(tmp_path)
     path = vasp_test_dir / "Si_band_structure" / "static" / "outputs"
     copy_vasp_outputs(src_dir=path, **copy_kwargs)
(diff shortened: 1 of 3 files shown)


tests/vasp/test_files.py
```diff
@@ -9,11 +9,14 @@
...
 )
-def test_copy_vasp_outputs_static(vasp_test_dir, tmp_dir, copy_kwargs, files):
+def test_copy_vasp_outputs_static(
+    vasp_test_dir, tmp_path, monkeypatch, copy_kwargs, files
+):
     from pathlib import Path
 
     from atomate2.vasp.files import copy_vasp_outputs
 
+    monkeypatch.chdir(tmp_path)
```
utf, line 19: Ah so this fixture doesn't automatically cd into the directory? To be honest, this seems like my implementation is actually better as it saves boilerplate?

janosh: I think it's often a good thing to be explicit about changing a test function's context. Since it's just 1 line, maybe acceptable amount of boilerplate? But feel free to close.

utf:
Thanks for flagging this, but I think I prefer it with the current setup!

PR #106, opened 2022-04-01
Bump GitHub actions to v3

janosh:
The typos fixed in `{changelog,install}.rst` in 17e9070 will conflict with #105. Happy to remove those changes from this PR before merging but wanted to bring them up.

### DIFF
--- a/src/atomate2/settings.py
+++ b/src/atomate2/settings.py
@@ -39,10 +39,10 @@ class Atomate2Settings(BaseSettings):
         "vasp_std", description="Command to run standard version of VASP."
     )
     VASP_GAMMA_CMD: str = Field(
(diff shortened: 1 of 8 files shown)


src/atomate2/settings.py
```diff
@@ -79,7 +79,7 @@ class Atomate2Settings(BaseSettings):
     VASP_RUN_BADER: bool = Field(
         False,
         description="Whether to run the Bader program when parsing VASP calculations."
-        "Requires the bader command to be on the path.",
+        "Requires the bader CLI to be on the path.",
```
utf, line 82: I think "executable" would be even better here?

janosh: Sounds good.

utf:
Great, thanks!

PR #112, opened 2022-04-07
Add AUTO_ISPIN to settings

arosen93:
The `auto_ispin` flag is really useful for reducing the computational cost of calculations with negligible magnetic moments. It's so important in fact that I think it should be added to the settings file so that any of the existing jobs/flows can be run with this parameter if the user desired.

This PR adds `AUTO_ISPIN` as a setting with a default of `False`.


utf:
Unfortunately, I don't think this will work quite as you expect. Because the VaspInputSetGenerator objects are created and serialised fully when the jobs are _made_ and not _run_, this means only the atomate2.yaml on the local machine (not where the calculations are run) will affect the auto_ispin setting.

One option would be to add an `override_auto_ispin` option to `write_vasp_input_set`, as I did for input set overrides, because this function is not called until the job is actually run, so it will use the settings on the running machine. https://github.com/materialsproject/atomate2/blob/7065ce387a4efefaaf96e12f7402e640ea0fee9a/src/atomate2/vasp/files.py#L181-L182
Although I can imagine this getting quite messy. E.g., what happens if someone wants to override `auto_kspacing` next. So if we are to go down his route it should be a more general solution.

The alternative is to add a new powerup `use_auto_ispin` that will automatically set auto_ispin for all VASP calculations. E.g. this code should be complete:

```python
def use_auto_ispin(
    flow: Job | Flow | Maker,
    value: bool = True,
    name_filter: str | None = None,
    class_filter: type[Maker] | None = BaseVaspMaker,
):
    """
    Update the auto_ispin setting of any VaspInputSetGenerators in the flow.

    Alternatively, if a Maker is supplied, the auto_ispin of the maker will be updated.

    Note, this returns a copy of the original Job/Flow/Maker. I.e., the update does not
    happen in place.

    Parameters
    ----------
    flow : .Job or .Flow or .Maker
        A job, flow or Maker.
    value : bool
        The value of auto_ispin to set.
    name_filter : str or None
        A filter for the name of the jobs.
    class_filter : Maker or None
        A filter for the VaspMaker class used to generate the flows. Note the class
        filter will match any subclasses.
    """
    dict_mod_updates = {"input_set_generator->auto_ispin": value}

    updated_flow = deepcopy(flow)
    if isinstance(updated_flow, Maker):
        updated_flow = updated_flow.update_kwargs(
            {"_set": dict_mod_updates},
            name_filter=name_filter,
            class_filter=class_filter,
            dict_mod=True,
        )
    else:
        updated_flow.update_maker_kwargs(
            {"_set": dict_mod_updates},
            name_filter=name_filter,
            class_filter=class_filter,
            dict_mod=True,
        )
    return updated_flow
```

What do you think?

arosen93:
Oh yeah, I totally didn't think about that! Thanks for catching it. 

My alternate suggestion was going to be a powerup, and I think your suggestion sounds great!

utf:
Yeah, I obviously didn't think about that before either, since the input set generator uses `SETTINGS.SYMPREC`. Will think about how best to handle that...

utf:
Hey @arosen93, just wondering if you'll address the above powerup in this PR or whether I should close it?

arosen93:
@utf, thanks for the reminder about this. If you want to include the powerup, feel free to close this PR and open one with your code snippet since you wrote it after all. It's not a must-need for me at the moment, so it might take a few days for me to get to it otherwise.

utf:
Ok! I added it in f3b4724

arosen93:
Thanks!

PR #116, opened 2022-04-12
fixed gzipping of files like `WSWQ.1`

jmmshn:
## Summary

The current behavior will copy all files with a suffix `WSWQ.1` `WSWQ.2` to a single `WSWQ.gz` file since `path.with_suffix` just replaces the suffix.
I think gzipping should always keep the original suffix.

Also think the line in `find_and_filter_files` is a typo.

### DIFF
--- a/src/atomate2/utils/file_client.py
+++ b/src/atomate2/utils/file_client.py
@@ -375,7 +375,7 @@ class FileClient:
             Overwrite gzipped file if it already exists.
         """
         path = self.abspath(path, host=host)
-        path_gz = path.with_suffix(".gz")
+        path_gz = path.parent / (f"{path.name}.gz")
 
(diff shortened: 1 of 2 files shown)


src/atomate2/utils/file_client.py
```diff
@@ -375,7 +375,7 @@ def gzip(
             Overwrite gzipped file if it already exists.
         """
         path = self.abspath(path, host=host)
-        path_gz = path.with_suffix(".gz")
+        path_gz = path.parent / (f"{path.name}.gz")
```
utf, line 378: This looks good, but you don't need the brackets around f"{path.name}.gz"

jmmshn: Ok removed

utf:
Great, thanks!

PR #117, opened 2022-04-13
non-radiative workflow

jmmshn:
# Nonradiative workflow


src/atomate2/vasp/schemas/defect.py
```diff
@@ -0,0 +1,377 @@
...
+        None, description="The energies of the defect (supercell) in charge state (q2)."
+    )
+
+    distorted_calcs_dirs: List[List[str]] = Field(
+        None,
+        description="Directories of distorted calculations, stored as seperate lists for each charge state",
+    )
+
+    @classmethod
+    def from_distorted_calcs(
```
utf, line 186: As mentioned above, it will be more efficient to specify the information you actually need. Which to me looks like just the structure, energy and directory. I'd also add the uuids for easier book keeping.

src/atomate2/vasp/sets/defect.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@dataclass
+class AtomicRelaxSetGenerator(VaspInputSetGenerator):
+    """Class to generate VASP atom-only relaxation input sets."""
+
+    def __post_init__(self):
+        """Initialize the class."""
+        super().__post_init__()
+        self.use_structure_charge = True
```
utf, line 18: You don't need to do all of this. Can just set the class attribute directory. E.g.,

```suggestion
        use_structure_charge: bool = True
```

utf (commented):
Thanks for this @jmmshn. It's looking in a really nice state. I have a few edits but hopefully they should all be quick!

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,214 @@
...
+            static_maker=self.static_maker,
+            prev_vasp_dir=dir2,
+            add_name=f"q={charge_state2}",
+        )
+
+        deformations1.append_name(f" q={charge_state1}")
+        deformations2.append_name(f" q={charge_state2}")
+
+        ccd_job = get_ccd_from_task_docs(
+            deformations1.output, deformations2.output, struct1, struct2
```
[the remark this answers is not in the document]
jmmshn: Is cluster-side performance ever going to be a bottle-neck?
I think doing it this way allow for one extra layer of validations and a bit more clear code.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -0,0 +1,164 @@
...
+
+    Parameters
+    ----------
+    ref : pymatgen.core.structure.Structure
+        pymatgen structure corresponding to the ground (final) state
+    distorted : pymatgen.core.structure.Structure
+        pymatgen structure corresponding to the excited (initial) state
+    static_maker : atomate2.vasp.jobs.core.StaticMaker
+        StaticMaker object
+    distortions : tuple
```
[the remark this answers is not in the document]
jmmshn: The distortion is given as a percentage of the distance between the relaxed states.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -0,0 +1,164 @@
...
+        cur_dir = Path.cwd()
+        fd_doc = FiniteDiffDocument.from_directory(cur_dir)
+        gzip_files(cur_dir, force=True)
+        return fd_doc
+
+    def update_incar(self):
+        """Update the INCAR."""
+        incar = Incar.from_file("INCAR")
+        incar.update({"ALGO": "None", "NSW": 0, "LWAVE": False, "LWSWQ": True})
+        incar.write_file("INCAR")
```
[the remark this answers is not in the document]
jmmshn: Done

src/atomate2/vasp/schemas/defect.py
```diff
@@ -0,0 +1,377 @@
...
+__all__ = ["CCDDocument", "WSWQDocument", "FiniteDiffDocument"]
+
+
+class WSWQDocument(BaseModel):
+    """WSWQ document schema."""
+
+    nspin: int = Field(None, description="Number of spins channels")
+    nkpoints: int = Field(None, description="Number of k-points")
+    nbands: int = Field(None, description="Number of bands")
+    data: List[List[List[List[float]]]] = Field(
```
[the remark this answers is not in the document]
jmmshn: Since I'm always calculating rates I don't see a need for the phase information for now. So I'm just gonna do the absolute value.

src/atomate2/vasp/sets/defect.py
jmmshn: oops yeah this had other logic in it before.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -0,0 +1,155 @@
...
+    static_maker.make(ref)
+
+    distorted_structures = ref.interpolate(distorted, nimages=sorted(distortions))
+    # add all the distorted structures
+    for i, d_struct in enumerate(distorted_structures):
+        static_job = static_maker.make(d_struct, prev_vasp_dir=prev_vasp_dir)
+        suffix = f" {i+1}" if add_name == "" else f" {add_name} {i}"
+        static_job.append_name(f"{suffix}")
+        jobs.append(static_job)
+        outputs.append(static_job.output)
```
utf, line 67: Similar to what I did in the electron phonon workflow, can you also add some metadata that will get written to a json file that includes the distortion value, charge state, and reference structure? Any json files will automatically get parsed by the TaskDocument and they're always useful to have in the database later on.

E.g., see what I did here: https://github.com/materialsproject/atomate2/blob/525d0ff78e880fc9c17aa8a9597ba810444681e2/src/atomate2/vasp/jobs/elph.py#L167-L178

Using the `update_maker_kwargs` is necessary in case someone uses a flow maker instead of a job maker for the static calculations, even if the static calculation is a job in this version of the workflow. One reason for this could be you want to do a gamma point only calculation first and then 2x2x2 afterwards. All you'd need to do is replace the `static_maker` with a flow maker and nothing else in the workflow has to change.

src/atomate2/vasp/schemas/defect.py
```diff
@@ -0,0 +1,379 @@
...
+        WSWQDocument
+            WSWQDocument object.
+        """
+        # TODO make the pymatgen code automatically determine if the object is complex or absolute value
+        data = np.abs(wswq.data)
+        return cls(
+            nspin=wswq.nspin,
+            nkpoints=wswq.nkpoints,
+            nbands=wswq.nbands,
+            data=data.tolist(),
```
utf, line 87: Ah, because you renamed matrix_elements to data this will fail. I'm happy for you to rename data back to matrix elements to make this simpler and match the attributes in the pymatgen class!

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,224 @@
...
+from atomate2.vasp.schemas.defect import CCDDocument
+from atomate2.vasp.sets.core import StaticSetGenerator
+from atomate2.vasp.sets.defect import AtomicRelaxSetGenerator
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_DISTORTIONS = (-1, -0.15, -0.1, -0.05, 0, 0.05, 0.1, 0.15, 1)
+
+DEFECT_RELAX_GENERATOR = AtomicRelaxSetGenerator(use_structure_charge=True)
```
utf, line 28: You can remove the use_structure_charge here because its specified in the input set itself.

src/atomate2/vasp/sets/defect.py
```diff
@@ -0,0 +1,56 @@
...
+            A dictionary of updates to apply.
+        """
+        return {
+            "IBRION": 2,
+            "ISIF": 2,
+            "EDIFFG": -0.05,
+            "LREAL": False,
+            "NSW": 99,
+            "LCHARG": False,
+            "ENCUT": 700,
```
utf, line 53: Do we want to drop this down to 500?

tests/vasp/flows/test_defect.py
```diff
@@ -0,0 +1,126 @@
...
+    mock_vasp(ref_paths, fake_run_vasp_kwargs)
+
+    si_defect = Structure.from_file(
+        test_dir / "vasp" / "Si_config_coord" / "relax_q1" / "inputs" / "POSCAR"
+    )
+    INCAR_UPDATES = {
+        "KSPACING": 1,
+    }
+
+    def update_calc_settings(flow):
```
utf, line 104: Why do you need this additional function? Why not just use `update_user_incar_settings(flow, incar_updates=INCAR_UPDATES)` directly?

utf (commented):
Hey, some more little bits I noticed

src/atomate2/vasp/flows/defect.py
jmmshn: I just wantted to be explicit here incase anything upstream changes since this is the most important setting

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,224 @@
...
+DEFECT_STATIC_GENERATOR = StaticSetGenerator(
+    user_incar_settings={
```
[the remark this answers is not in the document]
jmmshn: done

src/atomate2/vasp/flows/defect.py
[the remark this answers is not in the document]
jmmshn: Oh OK, that make sense I forgot the other files!
I'll make these changes

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,224 @@
...
+    charges
+        A list of charges on the structure
+
+    Returns
+    -------
+    dict
+        A dictionary with the two structures with the charge states added.
+
+    """
+    structs_out = [structure.copy() for _ in charges]
```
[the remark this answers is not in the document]
jmmshn: but that might be confusing for people looking at this.
I think we should have the decorated function and undecorated function look similar whenever possible.

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,224 @@
...
+        ----------
+        structure
+            A structure.
+        charge_state1
+            The reference charge state of the defect.
+        charge_state2
+            The excited charge state of the defect
+
+        """
+        name = f"{self.name}: {structure.formula}({charge_state1}-{charge_state2})"
```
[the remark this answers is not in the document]
jmmshn: oops, done

src/atomate2/vasp/jobs/defect.py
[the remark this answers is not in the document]
utf: Oh I just mean in the docstring

jmmshn:
The serialization problems are resolved but requires a new `pymatgen` release with this PR: https://github.com/materialsproject/pymatgen/pull/2501

I think the only thing that remains is the psuedo potential stuff for the tests.

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,224 @@
...
+        flow = self.ccd_maker.make(
+            structure=structure,
+            charge_state1=charge_state1,
+            charge_state2=charge_state2,
+        )
+        ccd: CCDDocument = flow.output
+
+        dirs0 = ccd.distorted_calcs_dirs[0]
+        dirs1 = ccd.distorted_calcs_dirs[1]
+        mid_index0 = len(self.ccd_maker.distortions) // 2
```
utf, line 229: I think it could be nice to store the undistorted index in the CCDDocument itself. Might be useful later on and then you could use it directly here?

src/atomate2/vasp/jobs/defect.py
```diff
@@ -0,0 +1,182 @@
...
+    outputs = []
+
+    # add the static job for the reference structure
+    static_maker.make(ref)
+
+    distorted_structures = ref.interpolate(distorted, nimages=sorted(distortions))
+    # add all the distorted structures
+    for i, d_struct in enumerate(distorted_structures):
+        static_job = static_maker.make(d_struct, prev_vasp_dir=prev_vasp_dir)
+        suffix = f" {i+1}" if add_name == "" else f" {add_name} {i}"
```
utf, line 75: Any reason why one is `i+1` and the other is just `i`?

src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Parameters
+        ----------
+        structure
+            A structure.
+        charge_state1
+            The reference charge state of the defect.
+        charge_state2
+            The excited charge state of the defect
+
```
utf, line 208: Can remove this blank line at the end of the docstring.

src/atomate2/vasp/schemas/defect.py
```diff
@@ -0,0 +1,357 @@
...
+        cls, directory: str | Path, **kwargs
+    ) -> FiniteDifferenceDocument:
+        """
+        Read the FintieDiff file.
+
+        Parameters
+        ----------
+        directory : str | Path
+            Path to the FintieDiff directory.
+        ref_dir : str
```
utf, line 48: This docstring doesn't match the function signature.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -0,0 +1,182 @@
...
+        static_job.append_name(f"{suffix}")
```
utf, line 76: Similar to what I did in the electron phonon workflow, can you also add some metadata that will get written to a json file that includes the distortion value, charge state, and reference structure? Any json files will automatically get parsed by the TaskDocument and they're always useful to have in the database later on.

E.g., see what I did here: https://github.com/materialsproject/atomate2/blob/525d0ff78e880fc9c17aa8a9597ba810444681e2/src/atomate2/vasp/jobs/elph.py#L167-L178

Using the `update_maker_kwargs` is necessary in case someone uses a flow maker instead of a job maker for the static calculations, even if the static calculation is a job in this version of the workflow. One reason for this could be you want to do a gamma point only calculation first and then 2x2x2 afterwards. All you'd need to do is replace the `static_maker` with a flow maker and nothing else in the workflow has to change.

tests/vasp/flows/test_defect.py
```diff
@@ -0,0 +1,127 @@
+import numpy as np
+
+from atomate2.vasp.flows.defect import ConfigurationCoordinateMaker
```
utf, line 3: I think if you move these imports inside the tests themselves it should fix the psuedopotential issue.

Kind of annoying, I don't know how to fix it permanently.

tests/vasp/schemas/test_defect.py
```diff
@@ -0,0 +1,89 @@
+from collections import defaultdict
+
+import numpy as np
+
+from atomate2.vasp.schemas.defect import CCDDocument, sort_pos_dist
```
utf, line 5: And move these imports too.

utf (commented):
Thanks @jmmshn, basically there now. Just some docstring fixes and metadata to add now.

For the pseudpotential issue, I think it could be that you need to ensure all imports of atomate2 functions/classes happens inside the test functions themselves. Otherwise the monkeypatching doesn't work. I added comments on the places you need to edit.

jmmshn:
@utf  Thanks for the additional comments!  The `json` stuff will be super helpful although I might tweak them once things go into production.
The pseudopotential testing thing still breaks though

utf:
I think failing tests are now just to do with the pymatgen version.

jmmshn:
Ok, all tests passing now!

utf (commented):
Hi @jmmshn. There are just a few minor unresolved comments to fix. If you click on the "Files changed" tab and scroll through the files you can see all the remaining comments.

Once those are done I will merge!

jmmshn:
@utf 
So I rewrote that part a little bit:
- The finite diff job should recycle the vasp kwargs from the static jobs
- Now it is just defined as a job and note a maker since we are recycling the vasp kwargs

https://github.com/materialsproject/atomate2/pull/117/files#diff-c9c36f118f994616adb2a3d6c859184ae30b50a6ba07cd6357aaad435e10db04R246-R258

jmmshn:
https://github.com/materialsproject/atomate2/blob/main/pyproject.toml#L90-L91
These two settings changed during the recent packaging update.  Are we sticking with 88 now?

utf:
Thanks for this!

PR #121, opened 2022-04-22
Add codespell pre-commit hook and format markdown links + code fences

janosh:
[`codespell`](https://github.com/codespell-project/codespell) catches typos with a bias against false positives both in code and in commit messages (requires `pre-commit install -t commit-msg`).


utf:
Thanks!

PR #133, opened 2022-05-04
Defect Workflow

jmmshn:
## Defect workflow

The main goal here is to calculate the formation energy diagram.
Steps:
- Run one bulk calculation first this gets us the LOCPOT file for finite-size corrections and also allows us to see that the calculations settings make sense (these can be big calculations so it's good to see one finish and then replicate the settings to additional calculations)
- Spawn a series of defect charge state calculations
- Collect them into summary documents
- Maybe compute the formation energy here

Tasks:
- [x] Basic dynamic wf for defect
- [x] Figure out what to store as additional information for task documents
- [x] ~~Compute the formation energy as part of the workflow~~ (not feasible since you need the phase diagram)
- [x] Store the `DefectEntry` as part of the workflows: the formation energy can be constructed with a list of defect entries and the phase diagram (which is needed for the chemical potentials).
- [x] Tests Cleanup and documentation


src/atomate2/vasp/flows/defect.py
```diff
@@ -0,0 +1,331 @@
...
+    """
+
+    name: str = "formation energy"
+    relax_maker: BaseVaspMaker = field(
+        default_factory=lambda: RelaxMaker(
+            input_set_generator=DEFECT_RELAX_GENERATOR,
+        )
+    )
+
+    def make(self, defect_gen: DefectGenerator, sc_mat: NDArray | None = None):
```
utf, line 79: `defect_gen` and `sc_mat` could be output references if you were using this maker in the middle of flow (likely to happen since you'll want to do a structural relaxation etc first).

For that reason, it might make sense to put almost all of this logic inside `perform_defect_calculations`.

utf (commented):
Hi @jmmshn.

Thanks very much for this workflow. This is going to be very useful.

I have some initial thoughts I want to run by you.

1. I think it could be better to remove the bulk relaxation and supercell generation from the `FormationEnergyMaker`. Instead, I propose that these functions be handled by another `Maker` higher up than this one.  Just because, if you want to run substitutions, vacancies, and interstitials, you'll end up relaxing the structure 3 times which is quite wasteful.

2. This might be controversial but I don't like the way the defect generators work. I know you are working from the pymatgen implementations but since you are rewriting them anyway, this could be time for a more radical change. My major problems are that 
   - You need a structure to initialise them. This means you need to know the structure at the time you make the defect generator but if you're going to do a relaxation or whatever, then you won't have this information.
   - You have no control over the charge states and species that are considered.

  Adding the ability to customise charge states/species could come later. But, I think it would be better to make the defect generators initialisable without a structure. Similar to how the new input set generators work. You only need the structure when you come to actually do the generating. The defect generators then become a container for the defect generation settings (defect type, oxidation states, species etc) and are a natural way to configure which defects to run.

If you decide to go down this route, then I'd also propose that your `FormationEnergyMaker` accept a DefectGenerator and structure as arguments only.

What do you think?

jmmshn:
Thanks for the suggestions!
I made changes according to point 1. 

I also agree with point 2 so I think the setup for the generators should be:
- the class should only define the behavior of the sc_generation
- options on how charge states are generated (either all charge states up until the canonical ICSD derived values or test charge states in a window around the canonical ICSD value)

I will make these changes and ping this thread once they are done.

src/atomate2/vasp/flows/defect.py
jmmshn: I think everything in the flow is wrapped in a job now.

jmmshn:
> This might be controversial but I don't like the way the defect generators work. I know you are working from the pymatgen implementations but since you are rewriting them anyway, this could be time for a more radical change. My major problems are that
> 
>     * You need a structure to initialise them. This means you need to know the structure at the time you make the defect generator but if you're going to do a relaxation or whatever, then you won't have this information.
> 
>     * You have no control over the charge states and species that are considered.

So since the `Defect_Generator` only creates UC representations of the defect, I think it should always be creates with the structure.  If additional relaxations need to be done (i.e. relax the supercell or if you have pre-relaxed defect cell) we can build in short-circuit mechanisms later.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,28 +30,127 @@
 
 logger = logging.getLogger(__name__)
 
+################################################################################
```
utf, line 33: Do you mind removing these headers?

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,28 +30,127 @@
...
+        default_factory=lambda: RelaxMaker(
+            input_set_generator=DEFECT_RELAX_GENERATOR,
+            task_document_kwargs={"store_volumetric_data": ["locpot"]},
+        )
+    )
+
+    def __post_init__(self):
+        """Post-initialization."""
+        self.relax_maker = self.relax_maker.update_kwargs(
+            update={
```
utf, line 97: One issue with this, is that any user settings will be overridden. I know if a user changes LVHAR this could break downstream things, however, if a user has specifically overridden something then we should defer to them as they might have a particular reason for doing so.

Instead, you could add these settings to the `DEFECT_RELAX_GENERATOR`, and that will also make this class a bit cleaner.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,28 +30,127 @@
...
+            class_filter=RelaxMaker,
+            dict_mod=True,
+        )
+
+    def make(
+        self,
+        defect_gen: DefectGenerator,
+        dielectric: float | NDArray,
+        bulk_sc_dir: str | Path | None = None,
+        sc_mat: NDArray | None = None,
```
utf, line 119: Can you make these names more descriptive. E.g.:

- `defect_gen` -> `defect_generator`
- `bulk_sc_dir` -> `bulk_supercell_dir`
- `sc_mat` -> `supercell_matrix`

I know these are a bit verbose but that is just the python way.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,28 +30,127 @@
...
+    ):
+        """Make a flow to calculate the formation energy diagram."""
+        jobs = []
+
+        if bulk_sc_dir is None:
+            get_sc_job = bulk_supercell_calculation(
```
utf, line 125: Currently this logic is quite hard to follow as `bulk_supercell_calculation` and `get_supercell_from_prv_calc` do several things.

Could you break it up into:
1. Getting the supercell matrix.
2. Getting the bulk_supercell_dir.

Basically the logic would then be very explicit and look like this. Before running everything else:

If bulk_dir provided:
- Add job to extract supercell (and check supercell matches provided supercell if not None).
- Set supercell_mat = supercell_extract.output
- Leave bulk_dir as is.

If bulk not provided:
- Add job to calculate supercell matrix.
- Add job to calculate bulk
- Set supercell_mat = get_supercell.output
- Set bulk_dir = bulk_job.output.dir_name

Afterwards use same logic to submit jobs.

That way, if we want to reuse the code to generate the supercell matrix that will be easy to do. At the moment the supercell logic is tied up with generating a bulk calculation.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,373 @@
...
+from atomate2.vasp.sets.defect import AtomicRelaxSetGenerator
+
+_logger = logging.getLogger(__name__)
+
+################################################################################
+# Default settings                                                            ##
+################################################################################
+
+DEFAULT_DISTORTIONS = (-1, -0.15, -0.1, -0.05, 0, 0.05, 0.1, 0.15, 1)
+DEFECT_INCAR_SETTINGS = {
```
utf, line 50: These all seem to be duplicates?

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,373 @@
...
+        name_counter[defect.name] += 1
+    return Response(output=output, replace=defect_q_jobs)
 
-logger = logging.getLogger(__name__)
 
+@job
+def run_all_charge_states(
+    defect: Defect,
+    relax_maker: RelaxMaker | None = None,
+    sc_mat: NDArray | None = None,
```
utf, line 222: This is now required, right?

utf, line 222: I think this could be a required argument too.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,373 @@
...
+
+    Parameters
+    ----------
+    defect:
+        A defect object representing the defect in a unit cell.
+    relax_maker:
+        A RelaxMaker object to use for the atomic relaxation.
+        If None, the default will be used (see DEFAULT_RELAX_MAKER).
+    sc_mat:
+        The supercell matrix. If None, the code will attempt to create a
```
utf, line 239: This is now changed right?

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,373 @@
...
+            defect_entries.append(def_ent)
+            fnv_plots[int(qq)] = plot_data
 
+        output["results"][defect_name] = dict(
+            defect=defect, defect_entries=defect_entries, fnv_plots=fnv_plots
+        )
+    return output
+
+
+################################################################################
```
utf, line 370: Can you remove this header?

src/atomate2/vasp/schemas/defect.py
```diff
@@ -18,6 +18,10 @@
 __all__ = ["CCDDocument", "WSWQ", "FiniteDifferenceDocument"]
 
 
+class DefectCalcSummary(BaseModel):
```
utf, line 21: Is this needed any more?

src/atomate2/vasp/sets/defect.py
```diff
@@ -44,13 +43,16 @@ def get_incar_updates(
...
             "ISIF": 2,
             "EDIFFG": -0.05,
             "LREAL": False,
             "NSW": 99,
-            "LCHARG": False,
             "ENCUT": 500,
             "LAECHG": False,
             "NELMIN": 6,
+            "ALGO": "Normal",
+            "NCORE": 4,
```
utf, line 55: Shouldn't set this here. It will be system specific.

src/atomate2/vasp/sets/defect.py
```diff
@@ -44,13 +43,16 @@ def get_incar_updates(
...
+            "GGA": None,
```
utf, line 56: I don't think this should be here either? We want to set GGA/use the GGA tag from the previous calculation

utf (commented):
Hi Jimmy, this is really coming along. I put a few more comments.

I have a general comment about the test files. There seem to be a lot unnecessary files included in the tests folders, (PROCARs, REPORTS, XDATCAR, WAVECAR, std_err.txt, OSIZCAR, etc). I thought the tool used to generate the test folders normally removes these extra files. Did you add them back? If so, could you remove any that aren't necessary for the tests? It seems like only LOCPOT is needed?

src/atomate2/vasp/files.py
```diff
@@ -189,3 +189,17 @@ def write_vasp_input_set(
...
+
+
+def cp_contcar_to_poscar(allow_bad_contcar: bool = True, **rename_files_kwargs):
+    """Check that the CONTCAR file can be parsed as a Structure.
+
+    Check for the unzipped CONTCAR file only.
+    """
+    try:
+        Structure.from_file("CONTCAR")
+    except Exception:
```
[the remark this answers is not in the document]
jmmshn: When the CONTCAR is empty we get:

```
ValueError: Empty POSCAR
```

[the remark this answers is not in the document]
jmmshn: However, the protection here should be "never copy CONTCAR to POSCAR unless you are absolutely sure the CONTCAR is good".
The empty CONTCAR is the most common problem case, but if there was file corruption or if there was never a CONTCAR in the first place they all result in different errors `IndexError` and `FileNotFoundError` so I think a general exception here is might be OK. 

The point of this function should be to _hopefully_ save some time for the user if a CONTCAR was provided during restart for example.

src/atomate2/vasp/files.py
```diff
@@ -99,7 +99,7 @@ def copy_vasp_outputs(
         rename_files(files_to_rename, allow_missing=True, file_client=file_client)
 
     if contcar_to_poscar:
-        rename_files({"CONTCAR": "POSCAR"}, file_client=file_client)
+        cp_contcar_to_poscar(allow_bad_contcar=True, file_client=file_client)
```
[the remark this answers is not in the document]
jmmshn: 1. I'm specifically thinking about restarts if you want to restart from a previously failed directory this check is required.
2. This is basically a situation that burned me a few times doing long-running defect calculations.  If I'm doing (Start calculation)->(hit wall time)->(move CONTCAR to POSCAR and restart) each time you restart VASP will create a new empty CONTCAR and there is the possibility of losing progress if you are doing this "by hand".
3. This was all done with restarting in mind. The additional stuff I added to `RelaxMaker` is meant to deal with this 
4. I added the kwarg after changing it to `check_contcar`

pyproject.toml
```diff
@@ -25,14 +25,15 @@ classifiers = [
 ]
 requires-python = '>="3.8"'
 dependencies = [
-    'pymatgen>=2022.3.22',
+    'pymatgen@git+https://github.com/materialsproject/pymatgen@master',
```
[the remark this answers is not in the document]
jmmshn: Bumped version in #155

src/atomate2/vasp/flows/defect.py
jmmshn: OK, I would still like to add some kind of checking and warning mechanism here, that might go in the different PR.

src/atomate2/vasp/flows/defect.py
jmmshn: Removed

src/atomate2/vasp/flows/defect.py
jmmshn: I made the following PR to jobflow that will allow me to do some paranoia checks on the supplied Maker.
https://github.com/materialsproject/jobflow/pull/177
It is getting a bit complicated here because for long-running HSE calculations I found that doing a PBE relaxation helps save a lot of time and tends a avoid charge sloshing.  So I am now feeding the maker a DoubleRelaxMaker with PBE followed by HSE.

src/atomate2/vasp/jobs/defect.py
jmmshn: technically this function did not need to change since the `sc_mat` will always be supplied from a previous calculation.
The default `sc_mat = None` behavior is just the default behavior of the defects packages.

src/atomate2/vasp/jobs/defect.py
jmmshn: removed

src/atomate2/vasp/sets/defect.py
jmmshn: OOPs

src/atomate2/vasp/schemas/defect.py
jmmshn: This is a placeholder for the final output schema. I have commented this out.

pyproject.toml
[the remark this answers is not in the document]
jmmshn: @utf this needed updates and packaging for the `pymatgen-analysis-defects` and updates to `pymatgen` those are out now.

src/atomate2/vasp/flows/defect.py
jmmshn: So I commented this out and will be added back after the PR is made.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,27 +33,160 @@
...
+        default_factory=lambda: RelaxMaker(
+            input_set_generator=PBE_GEN,
+            task_document_kwargs={"store_volumetric_data": ["locpot"]},
+        )
+    )
+
+    def __post_init__(self):
+        """Check the calculation settings."""
+        if self.validate_maker:
+            check_relax_maker(self.relax_maker)
```
utf, line 149: I think these checks should be removed. If people want to use wrong settings they should be able to. In my opinion we should remove this and just add a warning in the docstring to say that these options should be set on any input makers. We don't currently check that other settings work for other workflows and if we start going down that route then it will result in quite convoluted code. Especially when you can have nested flows it starts getting very difficult to check properly and the checks could be quite brittle. For example, what happens if you want to do an initial quick relaxation without LVHAR, and turn on the required settings afterwards. This will currently result in an error.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,27 +33,160 @@
...
-DEFECT_INCAR_SETTINGS = {
-    "ISMEAR": 0,
-    "LWAVE": True,
-    "SIGMA": 0.05,
-    "KSPACING": None,
-    "ENCUT": 500,
-}
-DEFECT_KPOINT_SETTINGS = {"reciprocal_density": 64}
+# Default relax should use: PBE -> HSE to save computation time
+PBE_GEN: AtomicRelaxSetGenerator = AtomicRelaxSetGenerator(
```
utf, line 58: In general, I'd like to avoid inheriting from subclasses of the input set generator. This is because it makes it very difficult to work out which settings have been changed as you have to go through all subclass layers.

I think we could add two new defect input sets, one for PBEsol and one for HSE. I suggest they are called `ChargeStateRelaxSetGenerator` and `HSEChargeStateRelaxSetGenerator` to be consistent with naming of the other input sets. They should live in `atomate2/vasp/sets/defect.py` file.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,27 +33,160 @@
...
+    kpts=((0.25, 0.25, 0.25),),
+    kpts_shift=(0, 0, 0),
+    kpts_weights=[1],
+)
+
+SPECIAL_KPOINT_GAMMA = Kpoints(
+    comment="special k-point",
+    num_kpts=1,
+    style=Kpoints_supported_modes.Reciprocal,
+    kpts=((0.25, 0.25, 0.25), (0.25, 0.25, 0.25)),
```
utf, line 52: Duplicated 1/4 1/4 1/4 point. Should (0, 0, 0) be in there?

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,367 @@
...
 from atomate2.utils.path import strip_hostname
 from atomate2.vasp.files import copy_vasp_outputs
-from atomate2.vasp.jobs.core import StaticMaker
+from atomate2.vasp.jobs.core import RelaxMaker, StaticMaker
 from atomate2.vasp.run import run_vasp
 from atomate2.vasp.schemas.defect import CCDDocument, FiniteDifferenceDocument
 from atomate2.vasp.schemas.task import TaskDocument
+from atomate2.vasp.sets.defect import AtomicRelaxSetGenerator
+
+_logger = logging.getLogger(__name__)
```
utf, line 43: Can just be called `logger` to be consistent with elsewhere in the codebase.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,367 @@
...
+
+    summary_d = dict(
+        uc_structure=uc_structure,
+        sc_entry=relax_output.entry,
+        sc_struct=relax_output.structure,
+        sc_mat=sc_mat.tolist(),
+        dir_name=relax_output.dir_name,
+        uuid=relax_job.uuid,
+    )
+    # waiter = wait_for_dict(summary_d)
```
utf, line 168: Unneeded.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,367 @@
...
+            defect_entries.append(def_ent)
+            fnv_plots[int(qq)] = plot_data
+
+        output["results"][defect_name] = dict(
+            defect=defect, defect_entries=defect_entries, fnv_plots=fnv_plots
+        )
+    return output
+
+
+# For Configuration-Coordinate-Diagram (CCD) Calculations
```
utf, line 366: You don't need these section comments!

utf (commented):
Hi Jimmy. Apologies for the delay getting back to this PR. I added some more comments. Also, there are some unresolved comments from the last review. I think if you go on the "Files Changed" tab you can see them more clearly!

Thanks again.

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,27 +33,160 @@
...
+
+DEFECT_STATIC_GENERATOR: StaticSetGenerator = StaticSetGenerator(
+    user_kpoints_settings=SPECIAL_KPOINT,
 )
 
+CCD_DEFAULT_DISTORTIONS = (-1, -0.15, -0.1, -0.05, 0, 0.05, 0.1, 0.15, 1)
+
+
+# Define RelaxMaker Errors
+def check_relax_maker(maker: Maker):
```
utf, line 97: Change name to `check_defect_relax_settings`

src/atomate2/vasp/flows/defect.py
```diff
@@ -22,27 +33,160 @@
...
+        "LHFCALC": True,
+        "PRECFOCK": "Fast",
+        "LASPH": True,
+        "LDAU": False,
+    },
+    user_kpoints_settings=SPECIAL_KPOINT,
+)
+
+HSE_DOUBLE_RELAX = DoubleRelaxMaker(
+    relax_maker1=RelaxMaker(input_set_generator=PBE_GEN),
```
utf, line 82: Remember to add WAVECAR to copy additional files kwarg.

jmmshn:
@utf I reworked how `DefectGenerator`s behave since I found that I was dragging around tolerances a lot in the upstream code.
That did cause some problems for this workflow since the old way had all the useful information in one nice MSONable object.
So now I just pass in a list of defects with the stipulations that they were all created for the same unit-cell structure.

src/atomate2/vasp/files.py
```diff
@@ -27,11 +27,13 @@ def copy_vasp_outputs(
     src_host: str | None = None,
     additional_vasp_files: Sequence[str] = tuple(),
     contcar_to_poscar: bool = True,
+    check_contcar: bool = False,
     file_client: FileClient | None = None,
 ):
     """
     Copy VASP output files to the current directory.
 
+
```
utf, line 36: Remove extra newline

src/atomate2/vasp/sets/defect.py
```diff
@@ -4,15 +4,39 @@
...
-    """Class to generate VASP atom-only relaxation input sets."""
+class ChargeStateRelaxSetGenerator(VaspInputGenerator):
+    """Generator for atomic-only relaxation for defect supercell calculations.
 
-    use_structure_charge: bool = True
+    Since the defect cells are assumed to be large, we will use only a single k-point.
+    """
+
+    def __post_init__(self):
+        self.use_structure_charge = True
```
utf, line 38: Please don't use post_init for this! Just use the normal dataclass variable. If people want to override this for any reason they won't be. Similarly for the other places you do this!

pyproject.toml
```diff
@@ -25,14 +25,15 @@ classifiers = [
...
-    'pymatgen>=2022.7.8',
+    'pymatgen>=2022.7.19',
     'custodian>=2019.8.24',
     "pydantic",
     "monty",
     'jobflow>=0.1.5',
     "PyYAML",
     "numpy",
     "click",
+    "pymatgen-analysis-defects>=2022.8.24",
```
utf, line 36: Can you add this as an optional dependency. Basically, pymatgen-analysis-defects relies on dscribe which is rather heavy.

utf (commented):
Some other things I noticed!

jmmshn:
@utf this requires a bump to job flow to be finished. (the check and warning thing)

utf:
Ok I've bumped jobflow version in main

jmmshn:
Hi, @utf I think I address all the issues.
So the current setup allows for granular control of which Defects to run but not which charge state.
I will make an update to the defects package that allows the user to override the automatically determined charge states of a given defect so that `Defect.get_charge_states()` returns the user-defined quantities instead of the automatically determined ones.  But that should not affect this PR in any way.

BTW thanks again for the review the code was greatly improved with your help!!

src/atomate2/vasp/flows/defect.py
```diff
@@ -4,45 +4,217 @@
...
+        if vis.incar["ISIF"] != 2:
+            raise ValueError("ISIF should be 2")
+
+    recursive_call(maker, func=check_func, class_filter=RelaxMaker, nested=True)
+
+
+@dataclass
+class FormationEnergyMaker(Maker):
+    """Maker class to help calculate of the formation energy diagram.
+
```
utf, line 105: Can you add a detailed description of what this flow does, what the various options are etc.

src/atomate2/vasp/flows/defect.py
```diff
@@ -4,45 +4,217 @@
...
+    def __post_init__(self):
+        """Check the calculation settings."""
+        if self.validate_maker:
+            check_defect_relax_maker(self.relax_maker)
+
+    def make(
+        self,
+        defects: list[Defect],
+        dielectric: float | NDArray,
```
utf, line 136: Dielectric should be optional, in which case corrections aren't calculated.

src/atomate2/vasp/jobs/defect.py
```diff
@@ -3,27 +3,324 @@
...
+            "entry": charged_output.entry,
+            "dir_name": charged_output.dir_name,
+            "uuid": charged_relax.uuid,
+        }
+    add_flow = Flow(jobs, output=outputs)
+    return Response(output=outputs, replace=add_flow)
+
+
+@job
+def collect_defect_outputs(
```
utf, line 249: This returns quite a complicated results object. Can you:
1. Create a new schema for it.
2. Add a `from_objects` or something class method that generates schema from the calculation data. This should rely only on the pymatgen objects themselves rather than the calculation directories (although it can include the calculation directories as inputs so they get stored in the schema as metadata). This function can then be used by a future builder.
3. Add another classmethod that accepts a list of directories etc, that loads the files and calls the classmethod from point 2.
4. In this job, just call the classmethod from point 3 and return the schema.

E.g., this should follow how we do it in the ElectronPhonon workflow (see [here](https://github.com/materialsproject/atomate2/blob/0fe95429145d7b2a0988bc4e3a36b924516f9711/src/atomate2/vasp/flows/elph.py#L167-L189) and [here](https://github.com/materialsproject/atomate2/blob/0fe95429145d7b2a0988bc4e3a36b924516f9711/src/atomate2/vasp/jobs/elph.py#L194)) and Phonopy workflow (see [here](https://github.com/materialsproject/atomate2/blob/0fe95429145d7b2a0988bc4e3a36b924516f9711/src/atomate2/vasp/flows/phonons.py#L314-L336) and [here](https://github.com/materialsproject/atomate2/blob/0fe95429145d7b2a0988bc4e3a36b924516f9711/src/atomate2/vasp/jobs/phonons.py#L193)).

For these classmethods, it is much better to be explicit with what arguments are required rather than relying on a dictionary with the data which is much harder to document and reproduce. When we eventually add a builder/create the CP2K defect workflow, this will reduce the code duplication significantly.

utf (commented):
Hi @jmmshn.

This is looking like its nearly done. In light of the new defect capabilities in #157, I've had a think about the `collect_defect_outputs` function. I think we should move to the Schema  paradigm for this job, which I've outlined in the comments.

The only other major thing is refactoring how the supercell matrix and bulk supercell dir are extracted like we discussed before. Currently these are done in the same function but this is quite messy and hard to follow.

Once this is merged we should have a chat with @nwinner to decide how to integrate with CP2K.

jmmshn:
@utf so something seems off with the MPRester tests `test_retrieve_structure_from_materials_project` I was running into a similiar problem with the defects package and queries just sporatically fail for no reason.
I think it best to spoof that the API call for testing.

utf:
Hi @jmmshn, sorry about that. We are looking into it. 

Please let me know when this is ready for re-review and any major changes

jmmshn:
This is ready for review.

The main change I made in addition to the ones discussed in the comments is:
The workflow now only accepts a single defect.  This makes the output schema significantly cleaner but you lose the ability to launch a bunch of defects at once without thinking due to the repeated bulk SC calculation.  However, that is still possible by linking many submissions with an initial bulk SC calculation.

jmmshn:
@utf just reminder that this is ready to merge.

mattmcdermott:
@jmmshn Hoping to try this out soon :) thank you for your contributions!!

tests/vasp/flows/test_defect.py
```diff
@@ -127,3 +148,89 @@ def test_nonrad_maker(mock_vasp, clean_dir, test_dir, monkeypatch):
...
+            parameters from PSCTR are:
+            VRHFIN =Si: s2p2
+            NELECTRONS =4.0
+            END of PSCTR-controll parameters
+        """
+        ps = PotcarSingle(pot_head, symbol="Si")
+        ps.keywords["NELECTRONS"] = 4.0
+        return ps
+
+    monkeypatch.setattr(PotcarSingle, "from_symbol_and_functional", fake_Si)
```
jmmshn, line 176: new monkey patch is here

.pre-commit-config.yaml
```diff
@@ -26,7 +26,7 @@ repos:
   rev: 5.10.1
   hooks:
   - id: isort
-- repo: https://gitlab.com/pycqa/flake8
+- repo: https://github.com/PyCQA/flake8
```
jmmshn, line 29: gitlab repo was deleted

jmmshn:
Also, I just found out that I have to press the submit review button for you to see my replies!! sorry about that

jmmshn:
So me not knowing you have to press the "submit review button" really messed this up eh?
@utf  Let me know if you need to start over from scratch b/c all the comments just got dumped at the same time and it's really hard to follow now.
Sorry again about the mixup!!

jmmshn:
Refactored to general defect workflow for different packages. #213

PR #134, opened 2022-05-05
Molecular dynamics job

mjwen:
Add basic MD job, based on 

- `MPMDSet` in pymatgen
- `MDFlow` in atomate1


src/atomate2/vasp/jobs/core.py
```diff
@@ -529,6 +531,40 @@ def make(
...
+        Keyword arguments that will get passed to :obj:`.should_stop_children`.
+    write_additional_data : dict
+        Additional data to write to the current directory. Given as a dict of
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    # TODO expose basic setting of INCAR parameters to user
```
utf, line 563: Please don't put these controls in the job maker, instead put them in the input set generator to be consistent with the other input sets.

Also, I'd advise against exposing the basic incar settings as those are easily overriden using `user_incar_settings`. Instead you could add a several "presets", specified by a string, e.g., "nvt", "npt", "long" or whatever you think is useful

src/atomate2/vasp/sets/core.py
```diff
@@ -775,6 +777,86 @@ def get_kpoints_updates(
...
+            # "MAXMIX": 20,
+            # "BMIX": 1,
+            # "NSIM": 4,
+            # "ADDGRID": True,
+        }
+
+        # use VASP default ENCUT
+        update["ENCUT"] = None
+
+        if defaults["ISPIN"] == 1:
```
utf, line 849: Not sure what defaults is here?

Instead of this, you might want to set `auto_ispin: bool = True` in this dataclass. See the docstring for that setting here: https://github.com/materialsproject/atomate2/blob/6445c2b67289abf9adc5a314aed0ae3488373a1d/src/atomate2/vasp/sets/base.py#L257

And this is an example of where I've turned it on: https://github.com/materialsproject/atomate2/blob/6445c2b67289abf9adc5a314aed0ae3488373a1d/src/atomate2/vasp/sets/core.py#L528

tests/vasp/jobs/test_core.py
```diff
@@ -193,3 +193,29 @@ def test_transmuter(mock_vasp, clean_dir, si_structure):
...
+
+    # settings passed to fake_run_vasp; adjust these to check for certain INCAR settings
+    fake_run_vasp_kwargs = {"transmuter": {"incar_settings": ["ISMEAR", "NSW"]}}
+
+    # automatically use fake VASP and write POTCAR.spec during the test
+    mock_vasp(ref_paths, fake_run_vasp_kwargs)
+
+    # generate transmuter job
+    job = MDMaker(
+        transformations=["SupercellTransformation"],
```
utf, line 215: I guess you don't need these options?

utf (commented):
Hey @mjwen, thanks for the PR. I realise this is a WIP but here are some initial comments.

mjwen:
thanks for the suggestions @utf ! The npt, nvt, ... are great suggestions; will definitely do it.  

I've got another questions for you. For MD run, I believe a user would want all ionic steps being parsed and stored. 
- seems `TaskDocument.from_directory()` will only parse the last step. What's a good way to processed? Modifying `TaskDocument` or creating a new class (probably subclassing) like `TaskDocument`?
- All the ionic steps are potentially a large amount of data, so should we store them in the extra output store by default?

utf:
I think the full ionic information is parsed. For each ionic step you get the energy, forces, and structure - do you need anything else? E.g., it is here: https://github.com/materialsproject/atomate2/blob/39c0508c1884f49ef246417bee0214991989b3f9/src/atomate2/vasp/schemas/calculation.py#L334

You're right that this will likely be huge for an MD simulation. One option would be to parse a [pymatgen Trajectory](https://pymatgen.org/pymatgen.core.trajectory.html#pymatgen.core.trajectory.Trajectory) object instead and store this in the data store. However, this shouldn't be the default situation.

If you're happy with a trajectory object, my advice would be:
1. Add a `store_trajectory` option to Calculation.from_vasp_files which parses the Trajectory, adds it to the list of `vasp_objects` and sets `ionic_steps` in the CalculationOutput to `None`.
2. Note that atomate2 already includes an [enum for trajectories](https://github.com/materialsproject/atomate2/blob/39c0508c1884f49ef246417bee0214991989b3f9/src/atomate2/vasp/schemas/calculation.py#L72)
3. Add the pymatgen Trajectory class to the list of objects that are automatically parsed into the data store in all vasp jobs: https://github.com/materialsproject/atomate2/blob/39c0508c1884f49ef246417bee0214991989b3f9/src/atomate2/vasp/jobs/base.py#L28
4. In the MDMaker, override the `make` function and enable the `store_trajectory` option. I did a similar thing to enable storing band structures here: https://github.com/materialsproject/atomate2/blob/39c0508c1884f49ef246417bee0214991989b3f9/src/atomate2/vasp/jobs/core.py#L204-L215 The `original` part of this line is very important `return super().make.original(self, structure, prev_vasp_dir)` otherwise the overriden function will return a job rather than actually running vasp.

Hope that makes sense, I can clarify further if not.

utf:
Also, did you see this page in the documentation on how to write VASP tests?

https://materialsproject.github.io/atomate2/dev/vasp_tests.html

mjwen:
> I think the full ionic information is parsed. For each ionic step you get the energy, forces, and structure - do you need anything else? 

I did a quick run of the MD job but did not find the `ionic_steps` in the output. After a closer look, it seems custodian changes `IBRION=0` to `IBRION=1` and then somehow the ionic_steps are now shown (not sure whether this is the reason). Anyway, can you point me to the place to config the behavior of custodian? 

Besides energy, forces, and structure at each ionic step, I may want other properties, like stress (if ISIF gives it) and partial charges (via bader analysis?) in the long run.  

I agree that pymatgen Trajectory may be a good option to store the data. I think I get the idea of how to make it happen and will let you know if I need more help.

utf:
> I did a quick run of the MD job but did not find the ionic_steps in the output. After a closer look, it seems custodian changes IBRION=0 to IBRION=1 and then somehow the ionic_steps are now shown (not sure whether this is the reason). Anyway, can you point me to the place to config the behavior of custodian?

There isn't any logic in the task document for storing ionic_steps depending on the value of IBRION. It should always be stored. Maybe you weren't looking in the right place. They should be in `TaskDocument.calcs_reversed[0].output.ionic_steps`.

And I've no idea about the custodian changing IBRION! You may have to look in the source code or ask on the issues page.

utf:
> Besides energy, forces, and structure at each ionic step, I may want other properties, like stress (if ISIF gives it) and partial charges (via bader analysis?) in the long run.

Good point! I double checked and these are the keys that are stored for each ionic step:
- e_fr_energy
- e_wo_entrp
- e_0_energy
- forces
- stress
- electronic_steps
- structure

So that should cover most things.

Bader might be a little more difficult because you'd have to continuously run Bader as the calculation is progressing as the CHGCAR at the end only has the data for the last step.

utf:
To make this clearer, I added IonicStep and ElectronicStep schemas in abecd04 and c3c5a0b.

mjwen:
> Bader might be a little more difficult because you'd have to continuously run Bader as the calculation is progressing as the CHGCAR at the end only has the data for the last step.

Yeah, it seems a second run is needed to get Bader for each ionic step. For now, I do not need it and let's leave it for a second. 

> To make this clearer, I added IonicStep and ElectronicStep schemas in [abecd04](https://github.com/materialsproject/atomate2/commit/abecd04e5bc61699c88abdd6b83a07f195065f4e) and [c3c5a0b](https://github.com/materialsproject/atomate2/commit/c3c5a0b7bef64db4e9c0ba0dc7ece0e370764df7).

Nice, this is great! 

Now that we have these nice schemas, instead of using pymartgen Trajectory, I think maybe we want to create a atomate2 schema for trajectory. Simply,
```
class Trajectory(BaseModel):

    ionic_steps: List[IonicStep] = Field(None, description="...") 
```
will do it. 

This will make it cleaner that we simply store ionic_steps in a different store than using a pymatgen Trajectory object.  Also, the pymatgen Trajectory has lots of additional stuff and may [not be the cleanest object](https://github.com/materialsproject/pymatgen/pull/2515) to store the data.

utf:
Unfortunately, creating a Trajectory `BaseModel` won't work. This is because only MSONable classes can be stored in the data store. All pydantic models will get converted to a dict (with no additional metadata) so it is impossible for jobflow to recognise them and store them in the data store.

mjwen:
I've come up with a way to deal with this. But before proceeding, I'd like to hear what you think. 

I've created a `Trajectory` BaseModel, and added it as a field `trajectory` of `CalculationOutput`. Its purpose is to replace `ionic_steps` in `CalculationOutput`. By default, `CalculationOutput.trajectory` is None and everything is the same as before. If instead we set `store_trajectory = True` in `CalculationOutput.from_vasp_outputs()`, `ionic_steps` will be set to None, and `trajectory` will be set to store the ionic steps. In addition, the trajectory will be stored in additional data store because `trajectory` is added to `_DATA_OBJECTS_`. 

This is basically what we've discussed before, with the only difference being that we use a Trajectory BaseModel instead of a pmg Trajectory class. The good side is that we store the same `ionic_steps`, just in different places. The ugly side is that now we have `ionic_steps` and `trajectory` in `CalculationOutput` for the same info. If you think it is OK, how about not using `trajectory` at all and let `ionic_steps` be stored in the additional store by default? Any downside of doing this? 

This is the major issue for now; and of course there are other TODOs and the tests I need to do before you want to review it.

mkhorton:
> This will make it cleaner that we simply store ionic_steps in a different store than using a pymatgen Trajectory object. Also, the pymatgen Trajectory has lots of additional stuff and may https://github.com/materialsproject/pymatgen/pull/2515 to store the data.

Just adding a note to say I'm keeping an eye on the pymatgen PR -- I am absolutely ok with merging in edits to this to make it cleaner or more useful for your workflows. As far as I know, current uses of this class are minimal (we do use it in the MP API to return the `Trajectory` associated with a geometry optimization, but nothing downstream uses this and it is constructed on-the-fly from the relevant task doc, so changes would be ok).

utf:
> If we want to make it to serve our purpose---storing all ionic steps info, we need to replicate a lot of schema already existing in atomate2. More specifically, we need to implement/copy [IonicStep](https://github.com/materialsproject/atomate2/blob/55ac4bc0fb8635fe333883ed74ff8d1f06130e5c/src/atomate2/vasp/schemas/calculation.py#L309) and its fields (e.g. [ElectronicStep](https://github.com/materialsproject/atomate2/blob/55ac4bc0fb8635fe333883ed74ff8d1f06130e5c/src/atomate2/vasp/schemas/calculation.py#L317)) to pymatgen. These schemas will all become attributes of the pymatgen Trajectory class

I'm not sure I follow this step? It seems like there is a `frame_properties` attribute on the Trajectory and you could just dump everything other than the structure there?

mjwen:
@utf Ready for review. 

In addition to adding the MDSetGenerator and MDMaker, major addition/changes include: 

- a pymatgen Trajectory is used to store ionic steps information for an MD job by default; no change for other jobs 
- to determine the status of an MD job, check whether the number of ionic steps is equal to NSW in INCAR, instead of checking the convergence of the last electronic and ionic steps 
- do not check max_force for an MD job

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+        # LREAL: True -> AUTO
+        updates.update(
+            {
+                "ENCUT": None,  # None to use VASP default
+                "TEBEG": self.start_temp,
+                "TEEND": self.end_temp,
+                "NSW": self.nsteps,
+                "POTIM": self.time_step,
+                "ISPIN": 2 if self.spin_polarized else 1,
+                "EDIFF_PER_ATOM": 0.00001,
```
utf, line 881: We have moved away from EDIFF_PER_ATOM in atomate2. It is reasonable to set a lower EDIFF here than the default but ultimately, I worry EDIFF_PER_ATOM could result in very large EDIFFS for large supercells.

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+                "MAXMIX": 20,
+                "NELM": 500,
+                "NSIM": 4,
+                "ISYM": 0,
+                "IBRION": 0,
+                "NBLOCK": 1,
+                "KBLOCK": 100,
+                "PREC": "Normal",
+                "LDAU": False,
+                "ADDGRID": True,
```
utf, line 899: We have also removed ADDGRID from most VASP calculations based on the VASP developers advice.

src/atomate2/vasp/schemas/calculation.py
```diff
@@ -414,6 +416,7 @@ def from_vasp_outputs(
         outcar: Outcar,
         locpot: Optional[Locpot] = None,
         elph_poscars: Optional[List[Path]] = None,
+        store_ionic_steps: bool = True,
```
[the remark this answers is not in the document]
mjwen: done.

src/atomate2/vasp/schemas/task.py
```diff
@@ -187,27 +190,47 @@ class OutputSummary(BaseModel):
     )
 
     @classmethod
-    def from_vasp_calc_doc(cls, calc_doc: Calculation) -> "OutputSummary":
+    def from_vasp_calc_doc_and_trajectory(
```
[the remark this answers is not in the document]
mjwen: done.

src/atomate2/vasp/schemas/task.py
```diff
@@ -187,27 +190,47 @@ class OutputSummary(BaseModel):
...
+        cls, calc_doc: Calculation, traj: Union[Trajectory, None] = None
+    ) -> "OutputSummary":
         """
-        Create a summary of VASP calculation outputs from a VASP calculation document.
+        Create a summary of VASP calculation outputs from a VASP calculation document
```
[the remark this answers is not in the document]
mjwen: done.

src/atomate2/vasp/sets/core.py
mjwen: done.

src/atomate2/vasp/sets/core.py
mjwen: agreed. EDIFF_PER_ATOM could be really bad for large cells, especially in MD. Removed it. I believe the default EDIFF=1e-5 in `BaseVaspSet.yaml` is good enough for MD runs.

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+        dict
+            A dictionary of updates to apply.
+        """
+        updates = self._get_ensemble_defaults(self.ensemble)
+
+        # Based on pymatgen.io.vasp.sets.MPMDSet. Changes include:
+        # LREAL: True -> AUTO
+        updates.update(
+            {
+                "ENCUT": None,  # None to use VASP default
```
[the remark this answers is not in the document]
mjwen: Yes, it is taken from pmg [MPMDSet](https://github.com/materialsproject/pymatgen/blob/1f1056cab8b9358f4d1ac2012f2a5732cd29bcf1/pymatgen/io/vasp/sets.py#L2657). 

I agree with you that this may not be good. The [MVLNPTMDSet](https://github.com/materialsproject/pymatgen/blob/1f1056cab8b9358f4d1ac2012f2a5732cd29bcf1/pymatgen/io/vasp/sets.py#L2676) uses a slightly larger value, 1.5*ENCUT, but still may not be ideal since the ENCUT changes from material to material. I prefer to set it to a const value, but smaller than that in the BaseVaspSet.yaml, like 520.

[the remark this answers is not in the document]
utf: Yeah, I think 520 is good.

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+    spin_polarized
+        Whether to do spin polarized calculations. The VASP `ISPIN` parameter.
+    """
+
+    ensemble: str = "nvt"
+    start_temp: float = 300
+    end_temp: float = 300
+    nsteps: int = 1000
+    time_step: int = 2
+    spin_polarized: bool = False
```
utf, line 838: Can you remove this parameter and set `auto_ispin: bool = True`. This means, if you're starting from a previous calculation, and that one has no magnetic moments, then ISPIN will be disabled automatically.

If you aren't starting from a previous calculation it is quite dangerous to set spin polarised to False by default.

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+        # ADDGRID: True -> False
+        # EDIFF_PER_ATOM: 0.00001, removed in favor of EDIFF
+        updates.update(
+            {
+                "ENCUT": None,  # None to use VASP default
+                "TEBEG": self.start_temp,
+                "TEEND": self.end_temp,
+                "NSW": self.nsteps,
+                "POTIM": self.time_step,
+                "ISPIN": 2 if self.spin_polarized else 1,
```
utf, line 882: Remove this and see note about `auto_ispin` above.

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+                "LCHARG": False,
+                "LPLANE": False,
+                "LWAVE": True,
+                "ISMEAR": 0,
+                "NELMIN": 4,
+                "LREAL": "Auto",
+                "BMIX": 1,
+                "MAXMIX": 20,
+                "NELM": 500,
+                "NSIM": 4,
```
utf, line 893: This is the default so can be removed. Same with a few others. So you can remove:

- NSIM
- LSCALU
- BMIX
- NBLOCK

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,129 @@ def get_kpoints_updates(
...
+                "ISYM": 0,
+                "IBRION": 0,
+                "NBLOCK": 1,
+                "KBLOCK": 100,
+                "PREC": "Normal",
+                "LDAU": False,
```
utf, line 899: Why is DFT+U disabled? Shouldn't we just use the default U values from pymatgen? In that case just remove this tag and they will be applied automatically!

utf (commented):
Thanks @mjwen, this is nearly there I think. I just spotted a few more edits.

davidwaroquiers:
Hello @mjwen and @utf , this is really nice work. Just my 2 cents, discussing with @gpetretto, one thing that may often happen in a MD job is that you need to continue it (you most probably cannot do e.g. 50000 steps in the timelimit). From what we have seen, it is not clear whether the velocities are passed down when you restart. Actually, we think that they are not and the MD job would restart with random or 0 velocities. What do you think ?
Thanks,
Best

mkhorton:
This is a timely comment @davidwaroquiers; I'm actually at a workshop on checkpoint/restart today... I think we're at a point where we need to establish an ideal pattern in atomate2 for how best handle checkpointed states and/or continuation jobs, @utf and I have had some offline discussions about this. Perhaps better to discuss in a separate issue?

I haven't reviewed this PR in detail, does it already have logic for continuation jobs included?

src/atomate2/vasp/sets/core.py
mkhorton: From a code hygiene perspective, @munrojm and I were discussing recently that there's a big advantage to having all calculations map cleanly to a pymatgen input set, since this greatly simplifies the task validation logic. If it's at all possible to inherit the settings from a specific input set, rather than copying manually, I think this would be very useful.

mjwen:
@mkhorton No, this PR does not include any logic to continue a job. I vaguely remember you guys were discussing the need to continue a job; @jmmshn was also participating? It would be great to have a specific issue where we can discuss this in greater detail.

src/atomate2/vasp/sets/core.py
mjwen: done

src/atomate2/vasp/sets/core.py
mjwen: done

src/atomate2/vasp/sets/core.py
mjwen: done. I also removed `LREAL=AUTO` and `ISMEAR=0`. For LREAL, the default setting in the yaml files have `AUTO`. For ISMEAR, I believer it is reasonable to follow the default yaml files.

src/atomate2/vasp/sets/core.py
mjwen: @mkhorton Just to make sure I understand, by `simplifies the task validation logic`, are you referring to the emmet building process or something else? And can you expand a bit on this...

mjwen:
@utf Besides removing some default params, I've also removed `nph` ensemble as an option, which is not that common and I don't think MP folks have a good sense on what good parameters should be for it.

src/atomate2/vasp/sets/core.py
utf: Also, @mkhorton could you clarify what you mean by "inherit the settings from a specific input set, rather than copying manually"?

utf:
Hi @mkhorton @davidwaroquiers @mjwen. Agreed that we need a proper discussion about restarting. This is something @jmmshn has opinions about too.

I have some ideas for how we could do it but each with their own tradeoffs. I'd be interested to hear if there were any good ideas from your workshop @mkhorton.

I will create a separate issue to discuss this further.

src/atomate2/vasp/sets/core.py
```diff
@@ -800,6 +808,123 @@ def get_kpoints_updates(
...
+        updates.update(
+            {
+                "ENCUT": 520,
+                "TEBEG": self.start_temp,
+                "TEEND": self.end_temp,
+                "NSW": self.nsteps,
+                "POTIM": self.time_step,
+                "LCHARG": False,
+                "LPLANE": False,
+                "LWAVE": True,
```
utf, line 880: Does this write the wavecar after each step? Could this be very expensive due to IO time of writing supercell wave functions?

davidwaroquiers:
> Hi @mkhorton @davidwaroquiers @mjwen. Agreed that we need a proper discussion about restarting. This is something @jmmshn has opinions about too.
> 
> I have some ideas for how we could do it but each with their own tradeoffs. I'd be interested to hear if there were any good ideas from your workshop @mkhorton.
> 
> I will create a separate issue to discuss this further.

Hi @utf, for abinit (PR #81 ), we have restart capabilities so we would indeed be very interested to discuss this and provide some of our ideas.

utf:
Great, thanks so much for this @mjwen!

PR #137, opened 2022-05-23
[WIP] VASP/Phonopy Phonon Workflow

JaGeo:
## Summary

This is a work in progress implementation of a Phonon Workflow with VASP and Phonopy. It is based on the work and experiences of @ab5424 in his Master's thesis (for atomate).

I'm still a long way from finishing the workflow, but I believe I've identified the most critical classes that need to be developed and have an idea regarding data transfer.

Of course, feedback is welcome now, but I believe it requires some more changes from my side before someone of you invests too much time.


JaGeo:
As I know that @gpetretto and @davidwaroquiers are also working on transferring workflows from abiflows to atomate2, there might be already some ideas about a Document for PhononData. If so, I would be happy to hear about it.

utf:
Thanks very much for this Janine!

I'll aim to take a proper look later this week. 

Would you mind installing the pre-commit? This will automatically raise (and most likely fix) the linting errors. To install you can run the following in the root directory:

```
pip install pre-commit
pre-commit install
```

For this first time, you can run the pre-commit on all files:
```
pre-commit run --all
```

Afterwards it will get run automatically whenever you commit something.

JaGeo:
Thanks. Will do. As I said, it really still needs some days of work and testing but it probably helps to get some suggestions with regard to general design now.

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,163 @@
...
+        jobs.append(displacements)
+
+        # perform the phonon displacement calculations
+        vasp_displacement_calcs = run_phonon_displacements(displacements.output,
+                                                           phonon_maker=self.phonon_displacement_maker,
+                                                           **self.run_phonon_displacements_kwargs)
+        jobs.append(vasp_displacement_calcs)
+
+        # Computation of BORN charges
+        if self.born_maker is None:
```
gpetretto, line 136: I have two comments about this point: 
1) Maybe you already planned to update this later, but for how it is written now the BECs are always calculated. However there are good reasons to skip this step (e.g. metals). It would be good that `None` means that BECs are not calculated.
2) with the `born_maker: BaseVaspMaker = field(default_factory=StaticSetGenerator)` it will enter this `if` only if born_maker is explicitly set to `None`. Also the default of the of born_maker will not include lepsilon. An option would be to define the born_maker:
`born_maker: BaseVaspMaker = field(default_factory=lambda: StaticSetGenerator(lepsilon=True))`

JaGeo: Thank you for pointing this out. I will add this option by using `born_maker: BaseVaspMaker |None = field(default_factory=lambda: StaticSetGenerator(lepsilon=True))` similar to what is done for the `bulk_relax_maker`. I was actually thinking about doing this but somehow did not completely implement this option ...

JaGeo:
As mentioned, I have added one of the suggestions by Guido and the pre-commit flags. I am happy to hear more. Otherwise, I hope that I can refine the workflow in the coming weeks. 
In parts of the workflow, I am using directories to access the files to create the output data. I guess it would be much cleaner if I would transfer forces and born data as outputs. This would be something, where I would appreciate an opinion.

JaGeo:
Switched to transfer via outputs for now. I think this should simplify extending the workflow to other codes as well.

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+    supercells = phonon.supercells_with_displacements
+
+    displacements = []
+    for cell in supercells:
+        displacements.append(get_pmg_structure(cell))
+    return displacements
+
+
+@job(output_schema=PhononBSDOSDoc)
```
utf, line 146: I think phonon band structures and density of states should be stored in the data store (for large objects). You can do this automatically using:

```suggestion
from pymatgen.phonon.bandstructure import PhononBandStructure
from pymatgen.phonon.dos import PhononDos

@job(output_schema=PhononBSDOSDoc, data=[PhononDos, PhononBandStructure])
```

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,409 @@
...
+    #     print("From Phonopy")
+    #     print(borns)
+    #     print(epsilon)
+    #
+    #     print("From atomate2")
+    #     print(epsilon_static)
+    #     print(born)
+    if born is not None:
+        if not conventional:
+            primitive = [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]
```
utf, line 230: I think phonopy now has tools to determine the primitive matrix automatically?

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,409 @@
...
+        imaginary_modes=imaginary_modes,
+    )
+
+    return phonon_doc
+
+
+@job
+def run_phonon_displacements(
+    displacements,
+    phonon_maker: BaseVaspMaker = None,
```
utf, line 313: I think it would be useful to include some metadata with these displacements. The metadata will automatically get added to the VASP task document. The metadata I have in mind is the index of the displacement, the original (non-supercell) structure, and the bulk supercell structure. This will help reconstructing the phonon documents later a bit easier!

You can see how I did this in the electron phonon workflow. Specifically, this function: https://github.com/materialsproject/atomate2/blob/9a7c3ccdadbbc617eb164e6ae1ff801aae2d1c31/src/atomate2/vasp/jobs/elph.py#L121

And the metadata is included using code such as this:

```python
        # write details of the electron phonon temperature and structure elph_info.json
        # file. this file will automatically get added to the task document and allow
        # the elph builder to reconstruct the elph document. note the ":" is
        # automatically converted to a "." in the filename.
        info = {
            "displacement_number": i,
            "original_structure": original_structure,
            "supercell_structure": supercell_structure,
        }
        elph_job.update_maker_kwargs(
            {"_set": {"write_additional_data->elph_info:json": info}}, dict_mod=True
        )
```

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,409 @@
...
+    "run_phonon_displacements",
+    "generate_frequencies_eigenvectors",
+    "PhononDisplacementMaker",
+]
+
+
+def get_phonon_object(
+    displacement, min_length, structure, sym_reduce, symprec, conventional
+):
+    transformation = CubicSupercellTransformation(min_length=min_length)
```
utf, line 49: Rather than having this here (where `CubicSupercellTransformation` will get called multiple times at different parts of the workflow), could it be refactored so that the supercell matrix is calculated once at the beginning of the flow and then passed directly to the various other jobs throughout.

utf (commented):
Hi @JaGeo, apologies its taken some time to look at this in detail.

Overall it looks great! This is exactly along the lines of what I envisioned. I definitely think using the outputs directly (rather than files) is a better approach. I don't think the amount of data transfer will be too bad.

I have two main comments:
1. We need to be careful about relying on objects inside of Flow Maker make functions. Any function that actually requires the objects should be wrapped in a job.
2. I think it would be useful to refactor the `generate_frequencies_eigenvectors` so that most of the work is done inside the pydantic model.

I've gone into a lot more detail in the code review.

If anything isn't clear, I'd be happy to discuss on here/email/zoom. Thanks again for this. I'm very excited to have a phonon workflow in atomate2.

JaGeo:
Thank you very much for the detailed review. I will try to improve the code according to these suggestions. I hope I can manage to find time in the next 1-2 weeks :).

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+    phonon_displacement_maker: BaseVaspMaker = field(
+        default_factory=PhononDisplacementMaker
+    )
+    generate_frequencies_eigenvectors_kwargs: dict = field(default_factory=dict)
+
+    def make(
+        self,
+        structure: Structure,
+        prev_vasp_dir: str | Path | None = None,
```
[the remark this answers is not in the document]
JaGeo: I agree with you. I am also not completely happy with the CubicSupercellTransformation a the moment. It prefers similar lattice parameters over 90 degree angles and I am not really sure this is optimal for these kind of phonon computations.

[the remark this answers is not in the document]
utf: ASE has a better scheme for generating a cubic supercell. See the function here https://wiki.fysik.dtu.dk/ase/ase/build/tools.html#ase.build.find_optimal_cell_shape

And the explanation here: https://wiki.fysik.dtu.dk/ase/tutorials/defects/defects.html#algorithm-for-finding-optimal-supercell-shapes

Perhaps we could use these ase functions instead?

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    # TODO: add some unit conversion factors
+    #  to easily use other codes to compute phonons?
+    name: str = "phonon"
+    sym_reduce: bool = True
+    symprec: float = SETTINGS.SYMPREC
+    displacement: float = 0.01
+    min_length: float = 20.0
+    conventional: bool = False
```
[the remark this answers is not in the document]
JaGeo: Agreed but then we should probably also make the computation of the band structure optional?

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,226 @@
...
+        vasp_displacement_calcs = run_phonon_displacements(
+            displacements.output,
+            phonon_maker=self.phonon_displacement_maker,
+            **self.run_phonon_displacements_kwargs,
+        )
+        jobs.append(vasp_displacement_calcs)
+
+        # Computation of BORN charges
+        if self.born_maker is not None:
+            born_job = StaticMaker(input_set_generator=self.born_maker).make(
```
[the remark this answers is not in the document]
JaGeo: Thanks! Understood!

src/atomate2/vasp/flows/phonons.py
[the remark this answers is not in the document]
JaGeo: I think I would need to test this in much more detail but it does look very useful especially with access to the optimality measure.

src/atomate2/vasp/flows/phonons.py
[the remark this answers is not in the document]
utf: Potentially? Although it might be possible to use [the AUTO option for PRIMITIVE_AXIS](https://phonopy.github.io/phonopy/setting-tags.html#primitive-axes-or-primitive-axis) to automatically align the band structure whatever input structure you use?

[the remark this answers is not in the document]
JaGeo: Yes, true.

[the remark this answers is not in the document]
JaGeo: Yes, I know this option. Need to check.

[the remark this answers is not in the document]
JaGeo: I would like to be able to use the kpath options from pymatgen. I am not sure phonopy uses the same convention for the primitive cell.

src/atomate2/vasp/jobs/phonons.py
JaGeo: Yes, agreed. I will indeed change this.

utf:
The changes you've made already look great! Happy to help diagnose any bugs should they come up.

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,418 @@
...
+                "factor": 14.399652,
+            }
+    phonon.produce_force_constants(forces=set_of_forces)
+
+    # Currently, the setting of the born
+    # charges does not work without this! I am not sure why!
+    # I don't see it
+    # I think I am missing something in the code
+    phonon.save("phonopy.yaml")
+    phonon = load("phonopy.yaml")
```
JaGeo, line 250: @utf : I currently have a problem here. I can only get correct bandstructures with NAC correction when I save the phonon object and then reload. There should be another way but I currently don't find the bug. If you have any idea, I am happy to hear.

JaGeo: I am attaching minimal example where it does not happen. I can currently not reproduce this bug outside of the workflow. 
[minimal_phonopy.zip](https://github.com/materialsproject/atomate2/files/8879780/minimal_phonopy.zip)

utf: I'm afraid I can't see what is going wrong!

Can you confirm that you're using the same phonopy version outside of the workflow?

JaGeo: I will confirm this and check again but thanks already for having a look!

JaGeo: I have upgraded to the latest phonopy version. It is still happening. I think I will continue with this file creation part for now and try to finalize the document part first. 
phonon_band_structure-01

JaGeo: This is the script that I am using for the workflow. I might need to try it in a completely new environment to make sure this is also not the reason for this problem.
[start_workflow.zip](https://github.com/materialsproject/atomate2/files/8880814/start_workflow.zip)

JaGeo:
Open points: 
- [ ] Code is still a bit messy, needs some refactoring 
- [ ] Cleanup of documentation, making sure every important parameter can be somehow adapted
- [ ] Need to check kpath creation and correspondence with created cells again
- [ ] decision what we would like to save in the document (should some more fields live in a data store for large data?)
- [ ] write unit tests
- [ ] make tests on real systems with real supercell sizes
- [ ] potentially create another way to determine supercell matrices that prefers 90 degree angels

JaGeo:
@utf in case you have some time in the next weeks, I would be happy about Feedback on the document part.

I think we need to store force constants, born charges etc. to easily develop workflows for Grüneisen parameters/quasi-harmonic approximation.

src/atomate2/vasp/jobs/phonons.py
utf: It seems like the workflow is using the same symprec throughout but it is quite high. I wonder if it will fix things if you use 1e-5?

JaGeo: I will try! Thanks!

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,329 @@
...
+    supercell_matrix: np.array,
+    displacement: float,
+    sym_reduce: bool,
+    symprec: float,
+    use_standard_primitive: bool,
+    kpath_scheme: str,
+    code: str,
+    displacement_data: dict[str, list],
+    total_energy: float,
+    epsilon_static: Matrix3D = None,
```
utf, line 174: I wouldn't define all of these optional arguments here. Firstly, because it is duplication of default parameters and could result in the defaults getting out of sync with the document model. And secondly because if you want to add a new option to the document model you will have to remember to add it again here.

Instead just use  **kwargs and then change the function contents to something like:
```
    phonon_doc = PhononBSDOSDoc.from_forces_born(
         structure=structure,
         supercell_matrix=supercell_matrix,
         displacement=displacement,
         sym_reduce=sym_reduce,
         symprec=symprec,
         use_standard_primitive=use_standard_primitive,
         kpath_scheme=kpath_scheme,
         code=code,
         displacement_data=displacement_data,
         total_energy=total_energy,
         **kwargs
     )
```

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    create_thermal_displacements: bool = True
+    generate_frequencies_eigenvectors_kwargs: dict = field(default_factory=dict)
+    kpath_scheme: str = "seekpath"
+    code: str = "vasp"
+
+    def make(
+        self,
+        structure: Structure,
+        prev_vasp_dir: str | Path | None = None,
```
utf, line 152: Instead of using `born_manual`, `epsilon_static_manual`, and `supercell_matrix` as dataclass fields, can you add `born`, `epsilon_static` and `supercell_matrix` to the make function arguments instead. And only calculate born etc if these are set to `None`.

utf:
> I think we need to store force constants, born charges etc. to easily develop workflows for Grüneisen parameters/quasi-harmonic approximation.

I think its a good idea to store the force constants too. Maybe there could be a flag to disable storing them if people are want to be very space conscious or e.g., if you are only doing convergence calculations.

src/atomate2/vasp/jobs/phonons.py
JaGeo: Thank you. I agree with this. I would however also add `born`, `epsilon_static` and `full_born` to the explicit keywords. They will be set in most of the cases.

src/atomate2/vasp/jobs/phonons.py
JaGeo: You were correct. It was the symprec... I have now introduced a second symprec setting. 

I will try to further clean up the code in the next days and check the kpath setting.

JaGeo:
I have cleaned up the output document:
image

@utf , please let me know what you think about the new structure.

I will then proceed to fulfill the other open points on my todo list above. I will also provide a detailed check of the structure/kpath correspondence here as I think this is the only part where something else could go massively wrong. Then, I will think about useful unit tests.

JaGeo:
I am currently performing some more quantitative tests for ZnS. I personally think that the convergence parameters are quite strict (maybe too strict?) and I would not expect any major differences to the Abinit results for now. I will post the comparison here. 

Then, I will add more documentation and further clean up the code and I would ask you to review again.

JaGeo:
I have now performed a large-scale test with a supercell larger than 15 Angstrom in each direction for ZnS (mp-10695). I think it compares extremely well to the phonon band structure from Abinit and DFPT. 

**MP**
image

**This workflow**
image

JaGeo:
I am now at the stage where I would need to write the unit tests and adapt requirements. I would be happy if another code review could be done by you @utf before so that I would not need to change the structure of the unit tests afterwards.

src/atomate2/vasp/files.py
```diff
@@ -177,7 +177,7 @@ def write_vasp_input_set(
     vis = input_set_generator.get_input_set(
         structure, prev_dir=prev_dir, potcar_spec=potcar_spec
     )
-
+    print(vis)
```
utf, line 180: Remove unused code.

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+        Whether to reduce the number of deformations using symmetry.
+    symprec : float
+        Symmetry precision to use in the
+        reduction of symmetry to find the primitive/conventional cell
+        (use_primitive_standard_structure, use_conventional_standard_structure)
+        and to handle all symmetry-related tasks in phonopy
+    displacement: float
+        displacement distance for phonons
+    min_length: float
+        min length of the supercell that will be build
```
utf, line 65:
```suggestion
        min length of the supercell that will be built
```

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+        with this workflow. Otherwise, we must rely on seekpath
+    use_conventional_standard_structure: bool
+        this will enforce to start the phonon computation
+        from the conventional standard structure
+        according to Setyawan, W., & Curtarolo, S. (2010).
+        High-throughput electronic band structure calculations:
+        Challenges and tools. Computational Materials Science,
+        49(2), 299-312. doi:10.1016/j.commatsci.2010.05.010.
+        We will however use seekpath and primitive structures
+        as determined by from phonopy to compute the phonon band structure
```
utf, line 88: Maybe these should be combined into a single parameter "use_symmetrize_structure", which accepts "primitive", "conventional", or None.

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+    bulk_relax_maker : .BaseVaspMaker or None
+        A maker to perform a tight relaxation on the bulk.
+        Set to ``None`` to skip the
+        bulk relaxation
+    static_energy_maker : .BaseVaspMaker or None
+        A maker to perform the computation of the DFT energy on the bulk.
+        Set to ``None`` to skip the
+        static energy computation
+    generate_phonon_diplsacements_kwargs: dict
```
utf, line 97:
```suggestion
    generate_phonon_displacements_kwargs: dict
```

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+    kpath_scheme: str = "seekpath"
+    code: str = "vasp"
+
+    def make(
+        self,
+        structure: Structure,
+        prev_vasp_dir: str | Path | None = None,
+        born_manual: List[Matrix3D] | None = None,
+        full_born: bool = True,
+        epsilon_static_manual: Matrix3D | None = None,
```
utf, line 155: Can you remove the "manual" from epsilon and born kwargs?

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        Make flow to calculate the phonon properties.
+
+        Parameters
+        ----------
+        structure : .Structure
+            A pymatgen structure.
+        prev_vasp_dir : str or Path or None
+            A previous vasp calculation directory to use for copying outputs.
+        born_manual: Matrix3D
```
utf, line 168: Is it possible to remove the `full_born` parameter and just figure out with the full or reduced born has been passed by checking the length of the born array. E.g., using some pseudocode like:

```python
if len(born) == len(structure):
    full_born = False
```

This will have to be put inside the job where the born is actually used as the structure might not be available otherwise.

Also, if you go down this route, you can specify that either the reduced or full born description can be supplied, in the born parameter docstring.

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+            self.supercell_matrix = None
+        else:
+            supercell_job = get_supercell_size(
+                structure,
+                self.min_length,
+                self.prefer_90_degrees,
+                **self.get_supercell_size_kwargs,
+            )
+            jobs.append(supercell_job)
+            self.supercell_matrix = supercell_job.output
```
utf, line 238: I think we shouldn't set `self.supercell_matrix` and instead this could just be:
```suggestion
        if supercell_matrix is None:
            supercell_job = get_supercell_size(
                structure,
                self.min_length,
                self.prefer_90_degrees,
                **self.get_supercell_size_kwargs,
            )
            jobs.append(supercell_job)
            supercell_matrix = supercell_job.output
```

And later on in this function replace `self.supercell_matrix` with just `supercell_matrix`

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,326 @@
...
+            phonon_maker=self.phonon_displacement_maker,
+        )
+        jobs.append(vasp_displacement_calcs)
+
+        # Computation of BORN charges
+        if self.born_maker is not None and (
+            born_manual is None or epsilon_static_manual is None
+        ):
+            born_job = self.born_maker.make(structure)
+            jobs.append(born_job)
```
utf, line 273: Can you update this to subsume the additional born logic later on in the function. Note this also assumes you rename born_manual and epsilon_static_manual, and are able to remove the full_born parameter:


```suggestion
        if self.born_maker is not None and (
            born_manual is None or epsilon_static_manual is None
        ):
            born_job = self.born_maker.make(structure)
            jobs.append(born_job)
			
            # I am not happy how we currently access "born" charges
            # This is very vasp specific code
            epsilon_static = born_job.output.calcs_reversed[0].output.epsilon_static
            born = born_job.output.calcs_reversed[0].output.outcar["born"]
            born_run_job_dir = born_job.output.dir_name
            born_run_uuid = born_job.output.uuid
        else:
            born_run_job_dir = None
            born_run_uuid = None
```

src/atomate2/vasp/jobs/phonons.py
utf: Ok that sounds good. Although please see my comment above about full_born, I think we should be able to detect this automatically.

src/atomate2/vasp/schemas/phonons.py
```diff
@@ -0,0 +1,474 @@
...
+    """
+    Collection of all data produced by the phonon worklow
+    """
+
+    structure: Structure = Field(
+        None,
+        description="Structure of Materials Project.",
+    )
+
+    ph_bs: PhononBandStructureSymmLine = Field(
```
utf, line 114: Can you rename just for clarity. E.g., to `phonon_bandstructure`

src/atomate2/vasp/schemas/phonons.py
```diff
@@ -0,0 +1,474 @@
...
+        None,
+        description="Phonon band structure object.",
+    )
+
+    ph_dos: PhononDos = Field(
```
utf, line 119: And similarly to `phonon_dos`.

src/atomate2/vasp/schemas/phonons.py
```diff
@@ -0,0 +1,474 @@
...
+        None,
+        description="Phonon density of states object.",
+    )
+
+    free_energy_list: List[float] = Field(
```
utf, line 124: You can remove the list, so it is just `free_energy` or `free_energies`

src/atomate2/vasp/schemas/phonons.py
```diff
@@ -0,0 +1,474 @@
...
+
+    born: List[Matrix3D] = Field(
+        None,
+        description="born charges as computed from phonopy. Only for symmetrically different atoms",
+    )
+
+    epsilon_static: Matrix3D = Field(
+        None, description="The high-frequency dielectric constant"
+    )
+    dft_phononwebsite: dict = Field(
```
utf, line 156: Do you have an idea of the size of this data? Is it very big?

utf (commented):
Hi @JaGeo, apologies for the delay looking at this PR, I was on annual leave and I then I just started a new job this week.

This is looking in great shape. I left some additional comments. It is basically tidying at this point, with a few renamings. Once those are done I think the last thing are the tests.

Thanks again!

JaGeo:
Thanks for the review! And, no worries :) . I can simply switch to another task if something needs more time. Congrats again on the new position!  
I will try to implement the changes in the coming weeks and then start with the unit tests. I will also try to use it for some projects soon to make sure it is also tested on a large number of different structures. @ab5424 tested the initial implementation on many structures but we changed many things due to relying on atomate2 defaults.

src/atomate2/vasp/flows/phonons.py
JaGeo: Thank for the suggestion.

src/atomate2/vasp/schemas/phonons.py
JaGeo: Will check this! Thanks for the hint!

src/atomate2/vasp/flows/phonons.py
JaGeo: Will remove this and the next kwargs as they are not further used in the code.

src/atomate2/vasp/flows/phonons.py
JaGeo: Will add the following and test if it works: 
`if len(born) == len(structure):
    full_born = True`

src/atomate2/vasp/flows/phonons.py
JaGeo: I agree and it would also fix the bug that I had overlooked so far ...

src/atomate2/vasp/flows/phonons.py
JaGeo: I updated it. Moved the static run after this whole born logic as well.

src/atomate2/vasp/schemas/phonons.py
JaGeo: Done

src/atomate2/vasp/schemas/phonons.py
JaGeo: Done

src/atomate2/vasp/schemas/phonons.py
JaGeo: Done

JaGeo:
I will start with the tests during this week. I currently, however, have problems with other tests when switching to a newer version of pymatgen. I guess this might be partially solved when the new defect workflow (https://github.com/materialsproject/atomate2/pull/133) is merged.

pyproject.toml
```diff
@@ -55,7 +55,8 @@ tests = [
 ]
 strict = [
     "pydantic==1.9.0",
-    "pymatgen==2022.7.8",
+    "pymatgen==2022.8.23",
+    "pymatgen-analysis-defects==2022.8.24",
```
utf, line 59: Please remove this! This will be an option dependency.

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,318 @@
...
+        with this workflow. Otherwise, we must rely on seekpath
+
+        "conventional" will enforce to start the phonon computation
+        from the conventional standard structure
+        according to Setyawan, W., & Curtarolo, S. (2010).
+        High-throughput electronic band structure calculations:
+        Challenges and tools. Computational Materials Science,
+        49(2), 299-312. doi:10.1016/j.commatsci.2010.05.010.
+        We will however use seekpath and primitive structures
+        as determined by from phonopy to compute the phonon band structure
```
utf, line 90: Use rst list formatting syntax instead:
```suggestion
        allowed strings: "primitive", "conventional", None
        
        - "primitive" will enforce to start the phonon computation
          from the primitive standard structure
          according to Setyawan, W., & Curtarolo, S. (2010).
          High-throughput electronic band structure calculations:
          Challenges and tools. Computational Materials Science,
          49(2), 299-312. doi:10.1016/j.commatsci.2010.05.010.
          This makes it possible to use certain k-path definitions
          with this workflow. Otherwise, we must rely on seekpath
        - "conventional" will enforce to start the phonon computation
          from the conventional standard structure
          according to Setyawan, W., & Curtarolo, S. (2010).
          High-throughput electronic band structure calculations:
          Challenges and tools. Computational Materials Science,
          49(2), 299-312. doi:10.1016/j.commatsci.2010.05.010.
          We will however use seekpath and primitive structures
          as determined by from phonopy to compute the phonon band structure
```

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,318 @@
...
+            supercell_matrix=supercell_matrix,
+            phonon_maker=self.phonon_displacement_maker,
+        )
+        jobs.append(vasp_displacement_calcs)
+
+        # Computation of static energy
+        if self.static_energy_maker is not None and total_dft_energy is None:
+            static_job = self.static_energy_maker.make(structure=structure)
+            jobs.append(static_job)
+            total_energy = static_job.output.output.energy
```
utf, line 269:
```suggestion
            total_dft_energy = static_job.output.output.energy
```

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,318 @@
...
+            static_run_job_dir = static_job.output.dir_name
+            static_run_uuid = static_job.output.uuid
+        else:
+            total_energy = total_dft_energy
```
utf, line 273: And then you can remove this line

src/atomate2/vasp/flows/phonons.py
```diff
@@ -0,0 +1,318 @@
...
+
+            # I am not happy how we currently access "born" charges
+            # This is very vasp specific code
+            epsilon_static = born_job.output.calcs_reversed[0].output.epsilon_static
+            born = born_job.output.calcs_reversed[0].output.outcar["born"]
+            born_run_job_dir = born_job.output.dir_name
+            born_run_uuid = born_job.output.uuid
+        else:
+            epsilon_static = epsilon_static
+            born = born
```
utf, line 290: Remove these lines

src/atomate2/vasp/jobs/base.py
```diff
@@ -133,7 +133,6 @@ def make(self, structure: Structure, prev_vasp_dir: str | Path | None = None):
 
         if "from_prev" not in self.write_input_set_kwargs:
             self.write_input_set_kwargs["from_prev"] = from_prev
-
```
utf, line 136: Can you add this back in?

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,447 @@
...
+    born_run_uuid=None,
+    static_run_job_dir: str | Path | None = None,
+    static_run_uuid=None,
+    optimization_run_job_dir: str | Path | None = None,
+    optimization_run_uuid=None,
+    # combine serval of these options
+    npoints_band: int = 100,
+    kpoint_density_dos: int = 7000,
+    tol_imaginary_modes: float = 1e-5,
+    tmin=0,
```
utf, line 224: Please see other comment about moving these optional arguments into **kwargs.

src/atomate2/vasp/schemas/phonons.py
```diff
@@ -0,0 +1,469 @@
...
+        )
+
+    @staticmethod
+    def get_kpath(
+        structure: Structure, kpath_scheme: str, symprec: float, **kpath_kwargs
+    ):
+        """
+            get high-symmetry points in k-space
+            Parameters
+        -   ---------
```
utf, line 442: Weird formatting?

Also there should be a blank line between the first line and the Parameters line

utf (commented):
Hi @JaGeo. I fixed the issue with pymatgen, so if you pull in the changes on main that will hopefully fix your problem. I also left some final comments which would be useful to address before doing any testing.

Thanks!

JaGeo:
Thanks! Will further refine it.

pyproject.toml
JaGeo: I will.

JaGeo: I updated the non-strict installation part as well. I hope this is fine.

src/atomate2/vasp/flows/phonons.py
JaGeo: Thanks. Will also change all other side effects of this change. I should probably run some of the automatic systems already now to spot such things. I wanted to wait until the unit tests are implemented as I always suspect potential problems.

src/atomate2/vasp/flows/phonons.py
JaGeo: In Germany, we call this "betriebsblind". Thanks.

src/atomate2/vasp/jobs/base.py
JaGeo: Yes, of course.

src/atomate2/vasp/jobs/phonons.py
JaGeo: Yes, will do. I am just sometimes a bit worried that people don't see through these **kwargs as it complicates usage in my opinion. Not sure everyone immediately understands the Structure.from_file(...) in pymatgen for example.

src/atomate2/vasp/schemas/phonons.py
JaGeo: Uups.

src/atomate2/vasp/jobs/phonons.py
utf: Yes, I also have this worry. Many people do not understand it. Perhaps you could add an explanation in the docstring pointing people to this other method?

Also, as this is a common practice in atomate2, we could even have a dedicated page in the docs about how to use atomate2 effectively, with some common design principles.

src/atomate2/vasp/flows/phonons.py
utf: That's a great word!

src/atomate2/vasp/jobs/phonons.py
JaGeo: Both suggestions sound like a great idea. 

I was also thinking about whether there should be a GUI at some point for people to build larger workflows by drag and drop or similar. You might have thought about it already and it would also be a massive amount of work ... but it's really hard to get into the usage workflows at the moment. Or we add even more tutorials...

utf: I think in the short term more tutorials is definitely needed. I was even thinking about adding a tutorial for most substantial workflows. E.g., we could add a workflow on how to do phonons using atomate2. It is more work from the developer side but it would definitely increase adoption!

As for a GUI, I know @mkhorton has a great vision for this, and its also something I'd really want to do. I'm not sure how to make it happen, specially regarding funding for the project. Realistically, it will need an experienced web developer working for ~1 year minimum. 

I actually think this could be something which a spin-off company would be well suited for. Create a web application that could be licensed out to industry to fund the development (academia could be free/get a discount). But I'm aware this doesn't fit into the ethos of MP.

JaGeo: I agree that this will be challenging. Maybe, we could also think about a template for developing your own workflow and mention the common steps in workflow development with important classes/methods for this? This is how I would start implementing something like this with a new co-worker as well. I don't think that we can expect typical chemists/physicits to navigate new and complex codes very well. 

I am really happy about the test documentation.

JaGeo: @davidwaroquiers I would love to join the discussion as it actually fits one of my broader and long-term interests.

src/atomate2/vasp/schemas/phonons.py
JaGeo: I decided to simply write a file to the last folder including this information so that it does not enlarge the database. Currently, this is only a dict and it has to be of the same size as the bandstructure objects.

JaGeo:
To test all potential paths through this workflow, I will add 4 more sets of test data. Unfortunately, this is necessary due to the options for kpath settings and cell definitions. Also, I need to test a compound different from Si to actually have useful BORN charges.

JaGeo:
I have added the last test. I found some issues in the normalization of the free energy for non-primitive cells that I have corrected and I have decided to change the input from total_dft_energy_per_cell to total_dft_energy_per_formula unit so that different cell definitions should be easier to handle.

I will probably now go through the typical workflow checks and then we could re-run workflows here.

If someone else would like to test this workflow even more, I would be happy about it. From my side, it will take a few weeks to run several (10-20) MP structures and compare to the results from DFPT and Abinit. Band structure of Si at least agrees with the MP within 2 cm-1.

JaGeo:
Would be ready for another round of review and the automatic tests from side at least.

JaGeo:
Thanks. Stupid mistake from my side...

phonopy is okay as an requirement?

utf:
Please could you add it as an optional dependency for now? And then feel free to edit the GitHub actions to also install that optional dependency set.

JaGeo:
Thanks. Yes, I think the latter is the issue. It's already an optional requirement.

JaGeo:
Sorry... Added it to the wrong workflow ....

JaGeo:
Had to add seekpath to the optional requirements [phonons] as well. Everytime I install phonopy, I forget to install seekpath ...

JaGeo:
Tests have passed. I would reply to any further comments later today or next week.

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,398 @@
...
+    epsilon_static: Matrix3D = None,
+    born: Matrix3D = None,
+    **kwargs,
+):
+    """
+    Analyze the phonon runs and summarize the results.
+
+    Parameters
+    ----------
+    structure: Structure object
```
utf, line 248: Same issue with missing description.

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,398 @@
...
+    displacement_data: dict
+        outputs from displacements
+    total_dft_energy: float
+        total dft energy in eV per cell
+    epsilon_static: Matrix3D
+        The high-frequency dielectric constant
+    born: Matrix3D
+        Born charges
+    kwargs:
+        additional arguments that are passed
```
utf, line 272: Say where they are to `PhononBSDOSDoc.from_forces_born`.

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,398 @@
...
+    """
+    Run phonon displacements.
+
+    Note, this job will replace itself with N displacement calculations.
+
+    Parameters
+    ----------
+    displacements
+    structure: original structure for meta data
+    supercell_matrix: supercell matrix for meta data
```
utf, line 310: Missing types and descriptions.

utf (commented):
Hi @JaGeo, a few more observations.

I'm really happy how this PR turned out. Thanks for all your edits. Personally, I think once you've made these changes we can merge this and then do the more substantial testing afterwards.

JaGeo:
Thanks. Will go through all of the suggested changes today! 
I am currently trying to setup everything on another computer so that I can hopefully test more compounds faster. I still sometimes have weird issues with custodian and VASP for extremely simple systems that I need to resolve ...

src/atomate2/vasp/jobs/phonons.py
```diff
@@ -0,0 +1,398 @@
...
+    "get_supercell_size",
+    "generate_phonon_displacements",
+    "run_phonon_displacements",
+    "generate_frequencies_eigenvectors",
+    "PhononDisplacementMaker",
+]
+
+
+@job
+def structure_to_primitive(structure: Structure, symprec: float):
```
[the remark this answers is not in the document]
JaGeo: I think it is probably fine to use the defaults here. Will move the functions, rename and change to default symprec.

JaGeo:
I hope I added all suggestions. I made one more change on the tests. I realized on the weekend that I had several useless statements in there ... 

We could rerun the tests again. I hope that I changed the installation configurations correctly. I will be really happy if you don't need to restart the automatic tests by hand every time I make a mistake ...

JaGeo:
Yeah, of course, more typos. I usually use the read-out-loud function everywhere to find such things... Now, I think it should be fine ;).

src/atomate2/common/jobs.py
```diff
@@ -6,25 +6,46 @@
...
-    Symmetrize a structure.
+    Job that creates a standard primitive structure.
 
     Parameters
     ----------
-    structure : .Structure
-        A structure.
-    symprec : float
-        The symmetry precision.
+        structure: Structure object
```
utf, line 19: I don't think these should be indented an extra level.

utf:
>  I will be really happy if you don't need to restart the automatic tests by hand every time I make a mistake ...

I think once this is merged GitHub won't require me to do it manually any more, since you will officially be a contributors. 🤞 

One final comment. Are you happy if I merge once that is resolved?

src/atomate2/common/jobs.py
JaGeo: Corrected.

JaGeo:
Yes, I would be happy if this is merged. I will then run several test structures, use it in my research and start a new pull-request to fix potential bugs in case they show up.

src/atomate2/common/jobs.py
JaGeo: And, tests have to be restarted, of course. Sorry.

utf:
Thanks very much @JaGeo. This is an exemplar workflow!

JaGeo:
Thank you so much

Issue #135, opened 2022-05-06
Feature suggestion: harmonic phonon workflow?


JaGeo:
I have worked on a phonon workflow including phonopy and the finite displacement method. I would like to make it available in atomate2. 

Would there be interest?

JaGeo:
#137

PR #138, opened 2022-05-23
Fix f-orbital DOS properties

arosen93:
Addresses #135. Because `get_element_spd_dos()` in `pymatgen` does not parse DOS data for f-orbitals, the schema for DOS properties breaks for such elements. This PR only tabulates DOS properties for s-, p-, and d-orbitals. 

Assuming a test case is desired, I'll need to run a calculation with f-orbitals first. Still working remotely, so I might need a day or two for that.


utf:
Thanks, this is ok.

arosen93:
Thanks! If any related issue still crops up, please ping me!

PR #141, opened 2022-06-06
update IO classes to latest upstream pymatgen

rkingsbury:
## Summary

This PR updates the contents of `src/atomate2/common/sets.py`, which defines the interface for `InputSet` and `InputGenerator`, to match the latest pymatgen. Originally I believe this file was added to `atomate2` before that interface was finalized and merged.

We could also remove this file and import the classes directly from pymatgen if desired; I thought there might be a specific desire to have a local copy in the `atomate2` repo to facilitate rapid iteration or something.


utf:
This is fantastic, thanks so much @rkingsbury.

> We could also remove this file and import the classes directly from pymatgen if desired; I thought there might be a specific desire to have a local copy in the atomate2 repo to facilitate rapid iteration or something.

The only reason for the duplication is that I was using the classes before your PR was merged in pymatgen. I think it would be best if we used the pymatgen input set and input set generator classes, since they are stable and merged now.

utf:
Great, thank you!

PR #142, opened 2022-06-08
Configure CI to use `setup-python`'s built-in `pip` caching

janosh:
`setup-python` has [built-in `pip` dependency caching](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching).

Note caching was not working before as it [used the hash of non-existent `setup.py` as the cache key](https://github.com/materialsproject/atomate2/compare/main...janosh:ci-deps-caching?expand=1#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023dL53).

Equivalent PR to pymatgen: https://github.com/materialsproject/pymatgen/pull/2401

Also moves `codespell` pre-commit config to `setup.cfg`. I noticed `setup.cfg` was emptied in b82138e so I tried adding it to `pyproject.toml` instead but found that's not supported yet. Tracking issue: https://github.com/codespell-project/codespell/issues/2055

Curious why did you decide to [keep `setup.cfg` around](https://stackoverflow.com/a/34111495)?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,6 +26,8 @@ jobs:
       - uses: actions/setup-python@v3
         with:
           python-version: '3.10'
+          cache: pip
+          cache-dependency-path: pyproject.toml
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/docs_manual.yml
+++ b/.github/workflows/docs_manual.yml
@@ -20,6 +20,8 @@ jobs:
       - uses: actions/setup-python@v3
         with:
           python-version: '3.10'
+          cache: pip
+          cache-dependency-path: pyproject.toml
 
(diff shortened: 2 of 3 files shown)


utf:
Thanks!

Would you mind leaving the setup.cfg empty and putting the codespell stuff back in the precommit file?

Basically, I want to get rid of setup.cfg but at the moment it is still required by pip for editable installations (or alternatively you can use a basic setup.py file). See the warning at the top of this page: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

janosh:
> Would you mind leaving the setup.cfg empty and putting the codespell stuff back in the precommit file?

Of course.

utf:
Thanks very much!

PR #151, opened 2022-07-02
Fix mock_run_vasp to accept args and test docs

mjwen:
This PR introduces two fixes related to the test system. 

- Fix `mock_run_vasp` to accept `*args` and `**kwargs` to match the signature of `run_vasp`. Otherwise, if a job maker sets some of these arguments (for example, the coming MD maker will set `handler`), the test will fail.  

- Fix doc on writing tests: changing the unit cell of the Si structure to avoid failing [check_poscar](https://github.com/materialsproject/atomate2/blob/a31b86b1e2f4d1a6665be7764c5cc67673904c91/tests/vasp/conftest.py#L200). In `check_poscar`, the `user` poscar will use a lattice of 
  ```
  3.348898 0.000000 1.933487
  1.116299 3.157372 1.933487
  0.000000 0.000000 3.866975
  ```
  while the `ref` poscar will use whatever lattice provided in the test file. The current doc on test suggests to use 
```
0.000000 2.730000 2.730000
2.730000 0.000000 2.730000
2.730000 2.730000 0.000000
```
 
  A test will fail if the two do not match. Currently, I just update the doc to make the the same. But I believe there should be better ways to do it. For example, what if a new maker do not want to use Si as the test system? Basically, it seems the `user` poscar is hard-coded? I know it is copied from somewhere to a tmpdir (cwd), but I don't know where the copy happens.  

None

## TODO (if any)
None

## Checklist

title.

- [x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  your local machine. Start with running [black](
  https://black.readthedocs.io/en/stable/index.html) on your new code. This will
  [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/), followed by [flake8](
  http://flake8.pycqa.org/en/latest/).
- [NA] Docstrings have been added in the[Numpy docstring format](
  https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [NA] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.
- [NA] Tests have been added for any new functionality or bug fixes.
- [x] All linting and tests pass.

### DIFF
--- a/docs/dev/vasp_tests.md
+++ b/docs/dev/vasp_tests.md
@@ -69,9 +69,10 @@ from monty.serialization import dumpfn
 
 # silicon structure
 si_structure = Structure(
-    lattice=[[3.348898, 0.0, 1.933487],
-             [1.116299, 3.157372, 1.933487],
-             [0.0, 0.0, 3.866975],
+    lattice=[
+        [3.348898, 0.0, 1.933487],
+        [1.116299, 3.157372, 1.933487],
+        [0.0, 0.0, 3.866975],
     ],
     species=["Si", "Si"],
     coords=[[0.25, 0.25, 0.25], [0, 0, 0]],
--- a/docs/dev/vasp_tests.md
+++ b/docs/dev/vasp_tests.md
@@ -74,7 +74,7 @@ si_structure = Structure(
              [0.0, 0.0, 3.866975],
     ],
     species=["Si", "Si"],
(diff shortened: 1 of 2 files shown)


utf:
Thanks!

PR #153, opened 2022-07-05
Regenerate calc_types enums

mjwen:
`run_types.yaml` was updated at some point but it seems `_generate.py` did not get run to generate `enums.py`. This PR regenerates it.


utf:
Great, thanks!

PR #157, opened 2022-07-19
CP2K Support

nwinner:
## Summary

This PR is for a WIP integration of CP2K support for atomate2. Defects will be included in a different PR.

* New CP2K features that will be in the main pymatgen by the time this is changed from WIP to production ready.

## TODO (if any)

- [x] Sets
- [x] Basic Jobs
- [x] Basic Flows including hybrid flows
- [x] Basic schemas/calc types 
- [x] Basic tests.


src/atomate2/cp2k/jobs/core.py
```diff
@@ -0,0 +1,341 @@
...
+    name: str = "md"
+    input_set_generator: Cp2kInputGenerator = field(default_factory=MDSetGenerator)
+    task_document_kwargs: dict = field(
+        default_factory=lambda: {
+            "store_trajectory": True
+        }
+    )
+
+# TODO This should go in common
+def _get_transformations(
```
utf, line 308: Please move this to common and integrate it with the vasp version of the function.

src/atomate2/cp2k/sets/base.py
```diff
@@ -0,0 +1,572 @@
...
+        self,
+        cp2k_input: Cp2kInput,
+        optional_files: dict | None = None,
+    ):
+        """
+        Initialize the set
+
+        Parameters
+        ----------
+        cp2k_input: Cp2kInput object for representing the main cp2k input file
```
utf, line 50: Can you format this properly?

src/atomate2/cp2k/sets/base.py
```diff
@@ -0,0 +1,572 @@
...
+        optional_files: If CP2K contains external file links, then they can be stored here.
```
utf, line 51: Can you format the docstring properly and add an example of how the links should work?

src/atomate2/cp2k/sets/base.py
```diff
@@ -0,0 +1,572 @@
...
+class Cp2kInputGenerator(InputGenerator):
+    """
+    A class to generate Cp2k input sets.
+
+    Parameters
+    ----------
+    user_input_settings:
+
+    user_kpoints_settings:
+
```
utf, line 148: Needs completing

src/atomate2/cp2k/sets/base.py
```diff
@@ -0,0 +1,572 @@
...
+    auto_kspacing: bool = True
+    use_structure_charge: bool = False
+    sort_structure: bool = True
+    symprec: float = SETTINGS.SYMPREC
+    force_gamma: bool = False
+    config_dict: dict = field(default_factory=lambda: _BASE_CP2K_SET)
+
+    def __post_init__(self):
+        """Post init formatting of arguments."""
+        pass
```
utf, line 184: Can this be removed?

src/atomate2/cp2k/sets/base.py
```diff
@@ -0,0 +1,572 @@
...
+
+        Note, if both ``structure`` and ``prev_dir`` are set, then the structure
+        specified will be preferred over the final structure from the last CP2K run.
+
+        Parameters
+        ----------
+        structure
+            A structure.
+        prev_dir
+            A previous directory to generate the input set from.
```
utf, line 203: `optional_files` is not described.

src/atomate2/cp2k/sets/base.py
```diff
@@ -0,0 +1,572 @@
...
+
+def recursive_update(d: Dict, u: Dict):
+    """
+    Update a dictionary recursively.
+
+    Args:
+        d: Input dictionary
+        u: Update dictionary
+    
+    Example:
```
utf, line 561: Can you use numpy style docstrings not google.

src/atomate2/cp2k/sets/core.py
```diff
@@ -0,0 +1,309 @@
...
+]
+
+
+@dataclass
+class StaticSetGenerator(Cp2kInputGenerator):
+    """
+    Class to generate CP2K static input sets.
+
+    Parameters
+    ----------
```
utf, line 40: Can you fix the docstrings for all classes in this file?

src/atomate2/cp2k/sets/core.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+@dataclass
+class HybridSetGenerator(Cp2kInputGenerator):
+
+    hybrid_functional: str = "PBE0"
+    screen_on_initial_p: bool = False
+    screen_p_forces: bool = False
+    eps_schwarz: float = 1e-7
+    eps_schwarz_forces: float = 1e-5
```
utf, line 105: Is there a benefit to defining these as class variables? It could be cleaner to put the values in the update function directly.

This should avoid confusion about how you change the hybrid functional. For example, as it stands the hybrid functional of some input set generators can be switched by changing the `hybrid_functional` dataclass field whereas for others you have to set `user_input_settings={"hybrid_functional": "HSE06"}`.

If you remove these class variables (which are a 1:1 mapping of the input update settings anyway) then the only standardised way of changing the functional will be through `user_input_settings`.

src/atomate2/cp2k/sets/core.py
```diff
@@ -0,0 +1,309 @@
...
+                "eps_schwarz_forces": self.eps_schwarz_forces,
+            },
+        }
+        if hasattr(structure, "lattice"):
+            updates['activate_hybrid']['cutoff_radius'] = get_truncated_coulomb_cutoff(structure)
+
+        return updates
+
+@dataclass
+@multiple_input_updators()
```
utf, line 123: I see why you've done this and if you feel strongly about it then I'm happy to keep it. However, I think this will lead to confusion with users and it will be much clearer and not that much repetition to just write out the update values explicitly. You could define a single dictionary with the HSE parameters at the top of the file if want to avoid boilerplate (but personally I don't think even this is necessary).

For the VASP workflows, there has traditionally been a lot of confusion about why certain values are getting set. I worry that this implementation (which is technically very neat) would also cause confusion.

src/atomate2/cp2k/sets/core.py
```diff
@@ -0,0 +1,309 @@
...
+        The time step (in femtosecond) for the simulation. The VASP `POTIM` parameter.
+    **kwargs
+        Other keyword arguments that will be passed to :obj:`VaspInputGenerator`.
+    """
+
+    ensemble: str = "NVT"
+    temperature: float = 300
+    nsteps: int = 1000
+    time_step: int = 2
+    thermostat: str = "NOSE"
```
utf, line 252: Again, I'm not sure that these class fields are really needed.

Typically I'd introduce class variables if I want to control multiple settings simultaneously. I.e., have a preset, so you could have one field, say "md_type", with options "NVT", "NPT", etc and by switching the type it will automatically update the thermostat/timestep/temperature. But in this case the settings are just 1:1 mappings to the input_settings.

utf (commented):
Hi @nwinner, this is a fantastic pull request, apologies that it's taken me some time to look at it in detail. Thank you for all your work on it, it's very impressive.

I really have very few comments (mainly on incomplete docstrings). I have a suggestion to remove the `multiple_input_updators` functionality since I think this could cause some confusion (more information is in the comments).

Once these minor changes are made I think this is ready to merge.

src/atomate2/cp2k/sets/core.py
nwinner: I think you're right that it should be changed. It's slightly more obtuse for users, because you can see that in order to set hybrid settings, there's this extra layer of having to know that you need to enter a new cp2k module using `{activate_hybrid: {...}}` instead of specifying `hybrid_functional` directly. I'm sure people can look into the source for an example otherwise, and since the hybrid job objects have class variables for setting it, I think you're ultimately right.


Will be resolved in next commit.

src/atomate2/cp2k/sets/core.py
nwinner: While I like how clean this is (and spent a good few hours trying to figure out how to do it). I'll defer to readability instead of elegance.  

Will fix in next commit.

src/atomate2/cp2k/flows/core.py
```diff
@@ -0,0 +1,286 @@
...
+@dataclass
+class HybridStaticFlowMaker(HybridFlowMaker):
+    """
+    Maker to perform a PBE restart to hybrid static flow
+
+    Parameters
+    ----------
+    name : str
+        Name of the flows produced by this maker.
+    pbe_maker : .BaseCp2kMaker
```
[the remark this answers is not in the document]
nwinner: While initializing with other methods is possible. PBE is definitely the most common, so I guess its alright to change.

src/atomate2/cp2k/flows/core.py
```diff
@@ -0,0 +1,286 @@
...
+    name : str
+        Name of the flows produced by this maker.
+    relax_maker1 : .BaseCp2kMaker
+        Maker to use to generate the first relaxation.
+    relax_maker2 : .BaseCp2kMaker
+        Maker to use to generate the second relaxation.
+    """
+
+    name: str = "double relax"
+    relax_maker1: BaseCp2kMaker = field(default_factory=RelaxMaker)
```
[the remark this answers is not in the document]
nwinner: Fixed where I could find it.

src/atomate2/cp2k/flows/core.py
```diff
@@ -0,0 +1,286 @@
...
+    hybrid_functional: str = "PBE0"
+    initialize_with_pbe: bool = field(default=True)
+    initialize_maker: BaseCp2kMaker = field(default_factory=StaticMaker)
+    hybrid_maker: BaseCp2kMaker = field(default_factory=HybridStaticMaker) 
+
+    def __post_init__(self):
+        self.hybrid_maker.hybrid_functional = self.hybrid_functional
+        if self.initialize_with_pbe:
+            self.hybrid_maker.input_set_generator.screen_on_initial_p = True
+            self.hybrid_maker.input_set_generator.screen_p_forces = True
```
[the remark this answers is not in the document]
nwinner: Uses powerup now

src/atomate2/cp2k/jobs/base.py
```diff
@@ -0,0 +1,215 @@
...
+    VolumetricData,
+    Trajectory,
+]
+
+
+def cp2k_job(method: Callable):
+    """
+    Decorate the ``make`` method of CP2K job makers.
+
+    This is a thin wrapper around :obj:`~jobflow.core.job.Job` that configures common
```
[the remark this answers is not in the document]
nwinner: Fixed

src/atomate2/cp2k/jobs/core.py
nwinner: Moved to new `atomate2.common.utils` and updated vasp.

nwinner:
@utf there are a few unresolved conversations above, but they can probably be closed once you give your opinion so its in the record.

Some checks are failing due to the spell-checker in pre-commit, which is looking into the test files for spell checking. It probably shouldn't do that. It also doesn't like the abbreviation "ot", because it thinks its a spelling mistake.

utf (commented):
Looking great. Thank you! Some final comments.

src/atomate2/cp2k/jobs/core.py
```diff
@@ -0,0 +1,363 @@
...
+        A generator used to make the input set.
+    """
+
+    name: str = "hybrid static"
+    hybrid_functional: str = "PBE0"
+    input_set_generator: Cp2kInputGenerator = field(default_factory=HybridStaticSetGenerator)
+
+    def __post_init__(self):
+        """Update the input settings with hybrid_functional attribute"""
+        self.input_set_generator.user_input_settings.update({"activate_hybrid": {"hybrid_functional": self.hybrid_functional}})
```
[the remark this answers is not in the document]
nwinner: So, let's say I go with option 1 and just trust the input set generator. Does that mean for the flows, like the one shown below, you would like to also remove `hybrid_functional` as an attribute? If we remove that, then the maker would have to be set by the user, and then the user would also have to set the input set generator inside of it. Just seems like a couple layers for something that is incredibly common, changing the hybrid functional. Or, do you think that the job should rely on the input set generator, but the flow can use an attribute to override things?
```
@dataclass
class HybridFlowMaker(Maker):
    hybrid_functional: str = "PBE0"
    initialize_with_pbe: bool = field(default=True)
    pbe_maker: Maker = field(default_factory=StaticMaker)
    hybrid_maker: Maker = field(default_factory=HybridStaticMaker)
```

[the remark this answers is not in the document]
utf: I think my main issue with this is overriding `user_input_settings` because it means the user can never override the job when I think we should strictly defer to the user if they are overriding.

Another alternative would be to override the input set config dict instead. That would achieve the same goal and still allow the user to override.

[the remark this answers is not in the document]
nwinner: Okay how about we delete the post-init for the jobs, but for flows, leave in the following:

```
def __post_init__(self):
    updates = {"activate_hybrid": {'hybrid_functional': self.hybrid_functional}}
    if self.initialize_with_pbe:
        updates['activate_hybrid'].update({"screen_on_initial_p": True, "screen_p_forces": True})
    self.hybrid_maker = update_user_input_settings(self.hybrid_maker, updates)
```

This does still do an override though.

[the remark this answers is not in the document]
nwinner: I made a further modification, which might help with the problem. `update_user_input_settings` is not modified to deal with nested dicts. This lets the powerup be called multiple times. The flow uses the powerup to set the necessary parameters, but users can still override with that powerup.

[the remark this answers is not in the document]
utf: Hey Nick, somehow I missed your comment.

From what I can see this current implementation will still suffer from the same issue I mentioned above, namely that this will be broken:

```
input_set = HybridStaticSetGenerator(
    user_input_settings={"activate_hybrid": {"hybrid_functional": "HSE"}}
)
maker = HybridStaticMaker(input_set_generator=input_set)
```

I think the best plan is to just update the config dict instead. That should fix all behaviour.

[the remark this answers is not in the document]
nwinner: If we update the config_dict, then `HybridStaticSetGenerator.get_input_updates()` will override it. So Job sets to HSE06, but then the set will override to the default PBE0...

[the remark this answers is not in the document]
nwinner: Alright. Now this should meet your specifications:

attributes have been dropped from the jobs, so to specify a job with HSE06 instead of default PBE0, you use the user_input_settings like you said

```python
input_set = HybridStaticSetGenerator(
    user_input_settings={"activate_hybrid": {"hybrid_functional": "HSE06"}}
)
maker = HybridStaticMaker(input_set_generator=input_set)
```

The flow now works with

```python
def __post_init__(self):
        """Initializing with PBE allows CP2K to screen exchange integrals using
        the PBE density matrix, which creates huge speed-ups. Rarely causes
        problems so it is done as a default here.
        """

        updates = {"activate_hybrid": {"hybrid_functional": self.hybrid_functional}}
        if self.initialize_with_pbe:
            updates["activate_hybrid"].update(
                {"screen_on_initial_p": True, "screen_p_forces": True}
            )
        self.hybrid_maker.input_set_generator.user_input_settings = recursive_update(
            updates, self.hybrid_maker.input_set_generator.user_input_settings
        )
```

Where config dict is left alone, so that get_input_updates doesn't override it, and instead we just call an update on the user input settings. As written `recursive_update` treats the settings from `__post_init__` as the base dict and treats the `user_input_settings` as the update dictionary. So whatever user specifies will be applied, but if nothing is supplied then it uses these flow settings.

utf (approved):
Thanks @nwinner. That looks like a good solution. I think we're ok to merge now.

utf:
@janosh I see codespell is flagging some issues in non-python or md files. Is it possible to only include md and py files in the linting? I tried to figure out how to do it but couldn't work it out.

janosh:
You could use `pre-commit`'s `types_or`:
```yml
- repo: https://github.com/codespell-project/codespell
  rev: v2.2.2
  hooks:
    - id: codespell
      types_or: [python, rst, markdown, yaml, toml, ...]
```

utf:
That's really helpful! Thanks.

utf:
@nwinner I've pushed the codespell fix to main so if you merge that in it should fix those errors at least.

utf:
@nwinner you might need to set up the package data in `pyproject.toml` to fix the other failing tests:
See this section:

```toml
[tool.setuptools.package-data]
atomate2 = ["py.typed"]
"atomate2.vasp.sets" = ["*.yaml"]
"atomate2.vasp.schemas.calc_types" = ["*.yaml"]
```

nwinner:
Thanks @utf. I've merged the new master. Not sure why some linting is still failing though. I have the latest mypy on my laptop running pre-commit, and yet the github wf says mypy is failing. I'll investigate.

nwinner:
So I found some issues. 

The pymatgen version in the config file for atomate2 is before the cp2k 2.0 updates, which is in 2023.1.9 release and beyond. That's causing some test errors. 

Bigger issue is Custodian's latest release is May 25, 2022, and the compatibility with cp2k 2.0 was released in December 2022. So there's no release to reference in the config file. Can the requirements link to a github version, or do I have to have shyue release a new version of custodian?

utf:
Ok, we can update the pymatgen version and we should ask to get another custodian release. I’d rather not use the GitHub custodian.

nwinner:
I've emailed Shyue about when/if he can do a new custodian release. I'll wait to hear back from him.

nwinner:
@utf We got a new Custodian release. Allowed me to fix a few things with tests. So this is ready to merge unless you have any more comments.

utf (approved):
This looks fantastic! Very excited to have this in atomate2. Thanks for all the work with the tests, they look perfect.

PR #161, opened 2022-08-24
[WIP] First pass at porting QChem into atomate2, waiting on concurrent PR approvals

rdguha1995:
## Summary

This is a draft attempt at porting Qchem into Atomate2 :

* I have closely tried to mimic the same layout and structure as VASP. 
* We have 4 folders of interest - **sets, schemas, jobs and flows**
* **_Sets_**

> - `base.py` in sets creates a `QChemInputGenerator` which is basically wrapping around the `QChemDictSet` from `pymatgen`
> - `core.py` in sets creates the input sets for different QChem jobs after inheriting the base generator.

* **_Jobs_**

> - `base.py` in jobs creates a `BaseQChemMaker` for running any type of Qchem-Atomate2 job. Note that the `base.py` in jobs also inherits `files.py` and `run.py`
> -  `core.py` inherits the `BaseQChemMaker` and creates jobs for the different Qchem-Atomate2 calculations
> 

* **_Flows (WIP)_**

> - This still requires work. From what I have understood, flows are used when you want to combine two non-io firetasks. Drawing on that parallel, the two current Qchem-Atomate fireworks which fits the bill is FFOpts and FFTS. I have created a skeletal flow for that but I am not sure whether we even need to have flows as a whole in Qchem-Atomate2. Would defer to @rkingsbury .

* **_Schemas (TODO)_**

> - Still need to figure out how we tackle the Qchem `TaskDocument`. Should we just inherit it from emmet? Again @rkingsbury would be the best person to answer that question.

I think this is a good start, but there are surely some missing pieces we need to figure out.


src/atomate2/qchem/flows/core.py
```diff
@@ -0,0 +1,169 @@
...
+    name : str
+        Name of the flows produced by this maker.
+    freq_maker : .BaseQChemMaker
+        Maker to generate the frequency calculation.
+    opt_maker : .BaseQChemMaker
+        Maker to generate the optimization calculation.
+    """
+
+    name: str = "frequency flattening structure optimization"
+    max_iterations=10
```
utf, line 46: For dataclasses you need to specify the type. E.g., this should be
```suggestion
    max_iterations: int = 10
```

Otherwise the dataclass won't work properly.

Please double check all your dataclasses.

src/atomate2/qchem/flows/core.py
```diff
@@ -0,0 +1,169 @@
...
+    freq_before_opt=False
+    perturb_geometry=False
+    mode=None
+    scale=1.0
+    max_errors=20
+
+    opt_maker: BaseQChemMaker = field(default_factory=OptMaker)
+    freq_maker: BaseQChemMaker = field(default_factory=FreqMaker)
+
+    def make(self, molecule: Molecule, prev_qchem_dir: str or Path or None = None, mode=mode, scale=scale):
```
utf, line 58: You can't specify types like that in type hinting. You need to use the `|` character. You also can't use `mode` or `scale` since they are dataclass variables. Instead you should remove these parameters and use `self.mode` in the function.
```suggestion
    def make(self, molecule: Molecule, prev_qchem_dir: str | Path | None = None):
```

src/atomate2/qchem/jobs/base.py
```diff
@@ -0,0 +1,158 @@
...
+#     CompleteDos,
+#     Locpot,
+#     Chgcar,
+#     Wavecar,
+#     Trajectory,
+#     "force_constants",
+#     "normalmode_eigenvecs",
+# ]
+
+_DATA_OBJECTS = [QCInput]
```
utf, line 37: DATA_OBJECTS controls what objects will get put into the jobflow "data" store which is used for large objects. I don't think QCInput should be in here as:
1. QCInputs won't be getting stored after a calculations.
2. They aren't large!

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,157 @@
...
+            new_geom_opt=new_geom_opt,
+            overwrite_inputs=overwrite_inputs,
+            nbo_params=nbo_params,
+        )
+
+        basis_set = self._basis_set(basis_set_updates)
+        scf_algorithm = self._get_scf_algorithm(scf_algorithm_updates)
+
+
+        return QChemDictSet(
```
utf, line 85: Can we not do this. It will make it very challenging to diagnose why certain things are getting written in the input sets. You will have to first check `QChemInputGenerator`, then check `QChemDictSet` and then finally check `QCInput`.

We should only have the generator `QChemInputGenerator` and the input set `QCInput`. The long term goal is to use only the modern input sets and remove the old style input sets. This won't be possible if the new input sets rely on the old ones.

So please can you refactor `QChemDictSet` into `QChemInputGenerator`. It does not look like it will take that long as the QChem inputs are a lot simpler than the VASP input sets.

utf (commented):
Hi @rdguha1995,

This looks great! Thanks very much.

My main comment is on the input sets. I think it is a very bad idea wrapping around the old style ones in pymatgen. I've written more detail in the review comments. It will definitely be worthwhile to refactor the old dict set into atomate2. I'm happy to help if you run into trouble but I think it should be quick.

On a separate note, you should install the pre-commit so that your commits are automatically formatted in line with the atomate2 style. To do that you can run the following in the root atomate2 folder:

```python
pip install pre-commit
pre-commit install
```

Then, any time you make changes the formatting tools will be applied. If changes are made, you will have to add them to the commit for the changes to be included. To run pre-commit on all files you can use:

```
pre-commit run --all
```

Happy to chat on zoom if you have any questions.

utf:
> Still need to figure out how we tackle the Qchem TaskDocument. Should we just inherit it from emmet?

I've started a discussion about this on emmet: https://github.com/materialsproject/emmet/issues/517

For this PR, I think it makes sense to use the existing task documents from emmet, but you will need to add new functions to initialise the task documents from the raw QChem files (e.g., see what I did in the VASP module in atomate2). When you add these functions in emmet, please can you mention me in the PR so I can review there too.

rdguha1995:
@utf , thanks for starting this thread on emmet. I will try to make a separate PR in emmet and assign you as a reviewer. In the meantime, would it be helpful if I am also tagged in https://github.com/materialsproject/emmet/issues/517  ?

src/atomate2/qchem/files.py
```diff
@@ -0,0 +1,153 @@
...
+def write_qchem_input_set(
+    molecule: Molecule,
+    input_set_generator: QChemInputGenerator,
+    directory: Union[str, Path] = ".",
+    from_prev: bool = False,
+    **kwargs,
+):
+    """
+    Write QChem input set.
+
```
rkingsbury, line 132: I don't understand why this method is needed. Shouldn't this always be handled by `InputSet.write_input`? If there is a need to instantiate a calculation from a previous directory, shouldn't that also be handled within the corresponding generator?

src/atomate2/qchem/jobs/base.py
rkingsbury: Agreed. This field would be for large data files, for example if we write out a CUBE file of the electron density or something like that.

src/atomate2/qchem/sets/core.py
```diff
@@ -0,0 +1,132 @@
...
+            prev_scf: str = None,
+            new_geom_opt: dict = None,
+            nbo_params: dict = None,
+            overwrite_inputs: dict = None, 
+        ) -> dict:
+            """
+            Get updates to the basis set for a single point calculation.
+            """
+            return {"basis_set": "def2-tzvppd", "scf_algorithm": "diis", "job_type": "sp"}
+
```
rkingsbury, line 43: Each of the `QChemInputGenerator` is implemented with a `get_basis_set_updates` method that returns a pre-populated dict. Wouldnt it be simpler to just do this directly, e.g.

```
class SinglePointSetGenerator(QChemInputGenerator):
    """Class to generate QChem Single Point input sets."""
    return QChemInputGenerator("basis_set"="def2-tzvppd",
                                                     "scf_algorithm"="diis",
                                                     "job_type": "sp")
'''

File: src/atomate2/qchem/sets/base.py [hunk 7]
```diff
@@ -0,0 +1,157 @@
...
+            prev_scf=prev_scf,
+            new_geom_opt=new_geom_opt,
+            overwrite_inputs=overwrite_inputs,
+            nbo_params=nbo_params,
+        )
+
+        basis_set = self._basis_set(basis_set_updates)
+        scf_algorithm = self._get_scf_algorithm(scf_algorithm_updates)
+
+
```
rkingsbury, line 84: The way I envisioned implementing an `InputGenerator` is that any of the core calculation settings, such as basis_set and scf_algorithm, are set as class attributes on instantiation, e.g.

```QChemInputGenerator(basis_set = 'def2-svpd')``` etc. 

Basically, any argument that needs to be passed to `InputSet` _other than` the strutcure. Then_ the `get_input_set` passes the class attrs, plus the `Molecule`, to the input set init method.

I could be missing something, but for me, having all these extra methods makes the code harder to understand, when really their only purpose is to set some kwargs in the InputSet constructor.

File: src/atomate2/qchem/sets/base.py (hunk 4 above)
Comment (rkingsbury) [reply src/atomate2/qchem/sets/base.py:85]: @utf I completely understand the sentiment and agree that refactoring `QChemDictSet` into a modern `InputSet` is the long term goal. When I looked into doing that before though, it did not seem straightforward. So I had suggested to Rishabh that as long as the legacy `QChemDictSet` behaves the same way as a modern `InputSet` (e.g., has a `write_input` method with the correct call signature), a sensible incremental solution would be to use an `InputGenerator` to create a legacy `InputSet`.

In general, I feel that the QChem `InputSet` do a lot less overrriding of input kwargs than the VASP ones so I don't think there's as high a risk of this becoming a confusing mess. @rdguha1995 it would be good to see if you can skip over the `DictSet` layer and directly create `QCInput` here, which would reduce the risk of a kwarg getting overridden somewhere. But I'm not sure how well that will work because I think the existing `QCInput` is somewhere in between the functionality of a modern `InputSet` and `InputFile`.

Aside from the risk of confusing overrides, I don't believe taking this approach adds any extra dependencies or complexifies the code (but please correct me if I'm wrong). If that's the case, I'd propose that we use this paradigm to get to beta quality atomate2 support for Q-Chem and tackle the `InputSet` refactor a bit down the line, just to minimize the number of moving parts.

File: src/atomate2/qchem/sets/core.py (hunk 6 above)
Comment (rkingsbury) [reply src/atomate2/qchem/sets/core.py:43]: In the simplest form, I had not even envisioned having any inheritance in the `InputSetGenerator`. E.g. the only function of the Generator would be to define specific sets of kwargs that get passed to `InputSet`

```
class OptSetGenerator(InputGenerator):

     def __init__(basis_set, scf_algorithm, dft_rung, etc.)
           ....

     def get_input_set(Molecule):

           return `QChemInputSet(molecule, basis_set, scf_algorithm, dft_rung).

```

There may be some technical advantages to have the intermediate class `QChemInputGenerator`, especially with regard to inheriting previous calc directories. But if we can possibly do without it, I'd advocate for the approach above. Keep things simple!

@utf I know VASP makes use of the dual layer model and I think that's what @rdguha1995 is emulating here. Do you have any additional thoughts on when/why that's advantageous?

**Review by rkingsbury — COMMENTED** [review]
Thanks for your work on this @rdguha1995! This will be a great service to use Molecular DFT folks. See some high-level review comments below. I can review in greater detail as this gets more refined.

File: src/atomate2/qchem/files.py (hunk 5 above)
Comment (utf) [reply src/atomate2/qchem/files.py:132]: The InputSet.write_input defines the way to write the input. But at some point you have to actually call it. And this function is the place where the function is called.

**Comment by utf** [comment]
Hi @rkingsbury, I think we should chat properly about the input sets as I'm very reluctant to merge something that is designed to be a temporary fix - in my experience it will never get fixed and we'll end up with a code base that is a collection of confusing and mixed standards.

The new approach to input sets in atomate2 actually makes things a lot more transparent from a user point of view if implemented correctly. The current input sets tackle two issues that got very messy with the old input set style:

1. You typically have a set of base settings for a calculation. However, these are not fixed in stone and one could conceivably change them and it wouldn't affect the type of calculation being performed. Think energy cut-off or basis set used. There is also a a second set of settings that define the calculation being performed and these must be set to achieve the desired result. E.g., think number of ionic steps for static vs relaxation calculations. This distinction becomes very important when starting from previous calculations. In particular, you want to ignore the "base settings" and only apply the settings that define this type of calculation.
2. Often the "update" settings (i.e., the ones that define the calculation type) depend on the actual molecule or previous output. E.g., they are dynamic and can't be defined by just a static dictionary. Therefore, this setup is a very flexible way to allow customisable input set updates. In particular, you only have to define a single function that can be used for both fresh and from previous calculations!

This also has the benefit that the updates that define an input set are defined in a single function (not a mixture of class variables, init methods, from_prev, and other functions). If you scroll through the VASP input sets or even the QChem input sets in this PR, it is very easy to discern the settings that define the calculation. If you look through the current input sets in pymatgen then it is much less obvious (and indeed there are sometimes inconsistencies between the different parts of the input set).

That said, I don't think the current implementation of the base input set and override functions in this PR is necessarily right. It looks like we can to simplify things a lot. Especially if we refactor the old dict set into the new generator. 

At our next meeting, we can discuss this in more detail to work out what the requirements of the input sets are, how they will be used, and the simplest way to implement them. I'm confident that this will result in a better user experience from the outset. And I also don't think it will take that much engineering effort.

**Review by rkingsbury — APPROVED** [review]

**Comment by rdguha1995** [comment]
Hey @utf , and @rkingsbury . This has been a long time coming and I am sorry for not being more prompt on this. I made a bunch of changes to this PR in October to refactor the refactoring `QCInput` and `QChemDictSet` into `QChemInputSet` and `QChemInputGenerator`.

I think this PR is a step closer to the infrastructure @utf envisions. There definitely might be more work to do but it would be great if a review can be performed on this PR. @utf I think that might be under your purview.

Thanks team for all the help. I have a feeling that if we give this one solid push, we would be much closer to incorporating `QChem` into `atomate2`

File: src/atomate2/qchem/jobs/base.py [hunk 8]
```diff
@@ -0,0 +1,158 @@
...
+
+from jobflow import Maker, Response, job
+from monty.serialization import dumpfn
+from monty.shutil import gzip_dir
+from pymatgen.core.structure import Molecule
+from pymatgen.io.qchem.inputs import QCInput
+
+from atomate2.qchem.files import copy_qchem_outputs, write_qchem_input_set
+from atomate2.qchem.run import run_qchem, should_stop_children
+from atomate2.qchem.schemas.task import TaskDocument
```
utf, line 17: Is this already in emmet? if so you can now use that one provided any atomate2 changes are moved across.

src/atomate2/qchem/schemas/task.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # TODO - type of `tags` field seems to differ among task databases
+    # sometimes List, sometimes Dict
+    # left as Any here to ensure tags don't cause validation to fail.
+    tags: Any = Field(None, description="Metadata tags")
+
+    warnings: Dict[str, bool] = Field(
+        None, description="Any warnings related to this task document"
+    )
+
```
utf, line 130: You will need to add a `from_directory` function to this task document/the one in emmet.

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,511 @@
...
+        #     molecule,
+        #     prev_basis=prev_basis,
+        #     prev_scf=prev_scf,
+        #     new_geom_opt=new_geom_opt,
+        #     overwrite_inputs=overwrite_inputs,
+        #     nbo_params=nbo_params,
+        # )
+
+        # basis_set = self._basis_set(basis_set_updates)
+        # scf_algorithm = self._get_scf_algorithm(scf_algorithm_updates)
```
utf, line 475: Reminder that from previous and `get_input_set_updates` are not implemented.

utf (commented):
Sorry for the delay with this @rdguha1995 @rkingsbury.

This is looking great and exactly along the lines of atomate2!

I added some minor comments. The main remaining things are:
1. Finishing the inputset implementation.
2. Finishing the schema implementation.
3. Adding some tests and wider documentation.


Happy to help discuss these points further (especially 3) if that would be helpful.

src/atomate2/qchem/sets/base.py
rdguha1995: Thanks a lot @utf for reviewing this! I am happy that we are moving closer :-) . I will be out for a conference next week. I hope to get back to this and implement all of what's left sometime in April!

I will ping you for feedback and assistance.

rkingsbury (commented):
Thanks @rdguha1995 ! I'm excited to see this come to fruition. See some comments throughout.

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,495 @@
...
+                    for k, v in temp_plots.items():
+                        plots_dict[k] = v
+                if sub == "nbo":
+                    if nbo_dict is None:
+                        raise RuntimeError(
+                            "Can't overwrite nbo params when NBO is not being run! Exiting..."
+                        )
+                    temp_nbo = lower_and_check_unique(sub_dict)
+                    for k, v in temp_nbo.items():
+                        nbo_dict[k] = v
```
orionarcher, line 402: could be replaced by something like:

```python
prop_dicts = {
    "rem": rem_dict,
    ...
    "plots": temp_plots,
}
if self.overwrite_inputs:
    for sub, sub_dict in self.overwrite_inputs.items():
        prop_dict = prop_dicts.get(sub, {})
        for k, v in lower_and_check_unique(sub_dict):
            prop_dict[k] = v
```

orionarcher (changes requested):
This is looking great @rdguha1995, very excited to have this in atomate2! 

My main comment is that we should move `QCInput` to the `QCInput.inputs['qcinput`] rather than `QCInput.qc_set`. That will allow the IO operations of base `InputSet` to be used.

In my mind the file writing logic currently in `QCInput` really belongs in `QChemInputGenerator`, but given that we don't want to make big breaking changes to `pymatgen`, I think it's fine as is.

Lastly, we should standardize whether we want `QC` or `QChem` as the prefix for objects. I'd advocate for `QC` to match `pymatgen`.

Happy to take another look or contribute code as needed to push this forward.

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,495 @@
...
+        Get a QChem Input Set as a dictionary for a molecule.
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        QchemInputSet
+            A QChem input set
+        """
```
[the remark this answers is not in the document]
rdguha1995: Slightly confused by this. The docstring change is correct, but looking at the `QCInput` implementations:-

`class QCInput(InputFile):
    """
    An object representing a QChem input file. QCInput attributes represent different sections of a QChem input file.
    To add a new section one needs to modify __init__, __str__, from_sting and add static methods
    to read and write the new section i.e. section_template and read_section. By design, there is very little (or no)
    checking that input parameters conform to the appropriate QChem format, this responsible lands on the user or a
    separate error handling software.
    """

    def __init__(
        self,
        molecule: Molecule | list[Molecule] | Literal["read"],
        rem: dict,
        opt: dict[str, list] | None = None,
        pcm: dict | None = None,
        solvent: dict | None = None,
        smx: dict | None = None,
        scan: dict[str, list] | None = None,
        van_der_waals: dict[str, float] | None = None,
        vdw_mode: str = "atomic",
        plots: dict | None = None,
        nbo: dict | None = None,
        geom_opt: dict | None = None,
        cdft: list[list[dict]] | None = None,
        almo_coupling: list[list[tuple[int, int]]] | None = None,
        svp: dict | None = None,
        pcm_nonels: dict | None = None,
    ):`
    
    We do need the other parameters as well right? for getting an input set?

src/atomate2/qchem/sets/core.py
```diff
@@ -0,0 +1,81 @@
...
+]
+
+
+@dataclass
+class SinglePointSetGenerator(QChemInputGenerator):
+    """Class to generate QChem Single Point input sets."""
+
+    def get_input_set_updates(self) -> dict:
+        """Get updates to the input dict for a single point calculation."""
+        return {"basis_set": "def2-tzvppd", "scf_algorithm": "diis", "job_type": "sp"}
```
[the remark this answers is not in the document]
rdguha1995: `get_input_set_updates` and all the accompanying `update` functions were implemented by me mirroring the `vasp` efforts. I guess `get_incar_updates` and other such functions are critical as subsequent jobs are executed in VASP. 

I don't think that is the case with `qchem`. For now, I am overloading the parent methods and commenting out `get_input_set_updates`.

src/atomate2/qchem/files.py
```diff
@@ -0,0 +1,153 @@
...
+    # from_prev : bool
+    #     Whether to initialize the input set from a previous calculation.
+    **kwargs
+        Keyword arguments that will be passed to :obj:`.QChemInputSet.write_input`.
+    """
+    # prev_dir = "." if from_prev else None
+    qis = input_set_generator.get_input_set(molecule)
+
+    logger.info("Writing QChem input set.")
+    qis.write_input(directory, **kwargs)
```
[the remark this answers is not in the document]
rdguha1995: Lol agreed. I didn't realize while writing it that without `from_prev`, this whole function is effectively one line.

src/atomate2/qchem/jobs/base.py
```diff
@@ -0,0 +1,160 @@
...
+        if prev_qchem_dir is not None:
+            copy_qchem_outputs(prev_qchem_dir, **self.copy_qchem_kwargs)
+
+        if "from_prev" not in self.write_input_set_kwargs:
+            self.write_input_set_kwargs["from_prev"] = from_prev
+
+        # write qchem input files
+        write_qchem_input_set(
+            molecule, self.input_set_generator, **self.write_input_set_kwargs
+        )
```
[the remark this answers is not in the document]
rdguha1995: Done!

src/atomate2/qchem/flows/core.py
```diff
@@ -0,0 +1,169 @@
...
+__all__ = [
+    "FrequencyFlatteningOptimizeMaker",
+    "FrequencyFlatteningTransitionStateMaker",
+]
+
+
+@dataclass
+class FrequencyFlatteningOptimizeMaker(Maker):
+    # need to incorporate iterations and perturb geometry
+    """
```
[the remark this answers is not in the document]
rdguha1995: In my opinion, `Flows` as a whole needs considerable work and should be sorted out after a detailed discussion with `qchem` users after people have started feeling comfortable with `atomate2` and `qchem`. I am not even sure besides `FFOpt`, whether there is a broader scope for any other flows in `qchem`. For now, I have added warnings, but as a whole `Flows` needs to be revisited for `qchem`

src/atomate2/qchem/schemas/calc_types/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    import atomate2.qchem.schemas.calc_types.generate
+
+from atomate2.qchem.schemas.calc_types.enums import CalcType, LevelOfTheory, TaskType
+from atomate2.qchem.schemas.calc_types.utils import (
+    calc_type,
+    level_of_theory,
+    lot_solvent_string,
+    solvent,
+    task_type,
+)
```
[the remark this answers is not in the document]
rdguha1995: Yes. In my opinion, the whole `schema` paradigm should not be necessary anymore, now that we have everything implemented in `emmet`. Moving out the entire `modules` for now

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,495 @@
...
+
+# _BASE_QCHEM_SET = loadfn(resource_filename("atomate2.qchem.sets", "BaseQchemSet.yaml"))
+
+__all__ = ["QChemInputSet", "QChemInputGenerator"]
+
+
+class QChemInputSet(InputSet):
+    """
+    A class to represent a QChem input file as a set.
+
```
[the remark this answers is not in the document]
rdguha1995: Don't mind doing this, but the return type is a `QChemInputSet` ?

src/atomate2/qchem/jobs/base.py
```diff
@@ -0,0 +1,160 @@
...
+    Returns
+    -------
+    callable
+        A decorated version of the make function that will generate QChem jobs.
+    """
+    return job(method, data=QCInput, output_schema=TaskDoc)
+
+
+@dataclass
+class BaseQChemMaker(Maker):
```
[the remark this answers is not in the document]
rdguha1995: @orionarcher, happy to change this. Is it just for consistency with `pymatgen`?

src/atomate2/qchem/sets/base.py
rdguha1995: After discussions with @rkingsbury , this whole paradigm has been moved out for now

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,495 @@
...
+                optional_inputs[name] = obj.from_file(directory / name)
+
+        return QChemInputSet(**inputs, optional_files=optional_inputs)
+
+        # Todo
+        # Implement is_valid property
+
+
+@dataclass
+class QChemInputGenerator(InputGenerator):
```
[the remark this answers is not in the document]
rdguha1995: Changed both `QChemInputSet` and `QChemInputGenerator` to `QCInputSet` and `QCInputGenerator` respectively. Just to be consistent

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,495 @@
...
+
+        inputs = {"Input_Dict": self.qcinput}
+        inputs.update(self.optional_files)
+
+        for k, v in inputs.items():
+            if v is not None and (overwrite or not (directory / k).exists()):
+                with zopen(directory / k, "wt") as f:
+                    f.write(v.__str__())
+            elif not overwrite and (directory / k).exists():
+                raise FileExistsError(f"{directory / k} already exists.")
```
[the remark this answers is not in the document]
rdguha1995: I can get behind this idea but I am slightly hesitant at this point. Looking at the current implementation of `write_input` in `pymatgen` it looks like if the contents are anything else than a `InputFile`, the contents will be written as a `str`. Currently we are storing the inputs as a `dict` here.

@orionarcher , maybe we can have a conversation about this, but i am leaving this untouched for now.

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,495 @@
...
+        pcm_dict: dict | None = None,
+        solv_dict: dict | None = None,
+        smx_dict: dict | None = None,
+        scan_dict: dict | None = None,
+        vdw_dict: dict | None = None,
+        plots_dict: dict | None = None,
+        nbo_dict: dict | None = None,
+        geom_opt_dict: dict | None = None,
+    ):
+        self.molecule = molecule
```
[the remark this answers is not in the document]
rdguha1995: Agreed, it was an oversight on my part. Corrected now.

src/atomate2/qchem/sets/base.py
rdguha1995: Keeping this unchanged for now,

@orionarcher , I feel that this doesn't change the current implementation too much. Agreed that it's long-winded, but the new `prop_dicts` will also be clunky.

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,511 @@
...
+    """
+    A class to generate QChem input set.
+
+    Parameters
+    ----------
+    job_type (str)
+        QChem job type to run. Valid options are "opt" for optimization,
+        "sp" for single point, "freq" for frequency calculation, or "force" for
+        force evaluation.
+
```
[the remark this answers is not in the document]
rdguha1995: Done!

src/atomate2/qchem/jobs/base.py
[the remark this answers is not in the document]
orionarcher: Yes precisely. I think best to keep consistent with `QC`

rdguha1995:
Hey Team - @rkingsbury  and @orionarcher!

Thanks a lot for all the comments! I incorporated most of them except inheriting the `write_input` from the parent `InputSet`. I understand that it would be useful in the long run and we can probably incorporate it in the future, but at this point, I have kept it there to be consistent with the `VASP` architecture. 

The one (but big) thing left is devising and implementing `tests`. For now, I will not play around with the `flows`. They are still a WIP, but I will try to implement `tests` for the `sets` , `jobs` and the other modules. 

Let me know if you have any specific feedback

rdguha1995:
Happy New Year @utf ! 


Just wanted to bring you up to speed with this. I finished integrating tests and other remaining components for this PR over the break. It is pretty much ready to merge now. I think some tests relating to the ML mace workflows are failing which is preventing the tests to complete. Let me know if I can do anything else from my end. Thanks for all the help. I will keep thinking about more functionalities to add to this in the future (i.e. possibilities of some flows.). But I think this can be a good starting point for qchem users on atomate2.


Let me know what you think

Also tagging @Zhuoying on this!

src/atomate2/settings.py
```diff
@@ -185,7 +185,34 @@ class Atomate2Settings(BaseSettings):
         None, description="Additional settings applied to AMSET settings file."
     )
 
-    model_config = SettingsConfigDict(env_prefix="atomate2_")
```
utf, line 188: Should this have been removed?

src/atomate2/settings.py
```diff
@@ -185,7 +185,34 @@ class Atomate2Settings(BaseSettings):
...
+        " throw an error",
+    )
+
+    QCHEM_STORE_ADDITIONAL_JSON: bool = Field(
+        default=True,
+        description="Ingest any additional JSON data present into database when "
+        "parsing QChem directories useful for storing duplicate of FW.json",
+    )
+
+    class Config:
```
utf, line 214: I think this was the old pydantic style config, and can be removed.

tests/conftest.py
```diff
@@ -44,7 +46,8 @@ def clean_dir(debug_mode):
     os.chdir(new_path)
     yield
     if debug_mode:
-        print(f"Tests ran in {new_path}")  # noqa: T201
```
utf, line 47: Can you revert this change?

utf (commented):
Hi @rdguha1995, thanks for these changes. It seems like the testing and linting are failing unrelated to the mace tests.

rkingsbury:
Thanks for your painstaking work on this @rdguha1995 ! Your timing is perfect - one of my students is just about in a position to "graduate" from manual Q-Chem calculations to atomate/atomate2 and I'll encourage him to test this out.

I see that there is a code stub for Frequency Flattening, but if I understand correctly, that `Flow` is meant to be implemented in future work, correct?

rdguha1995:
Thanks @utf for all the pointers. 

I resolved all the standing issues with my current code. The only thing left now is increasing my test coverage. I think adding a few tests for `drones` and `sets` will do the trick! Please let me know if you have any suggestions.

@rkingsbury thanks! It was a whirlwind, but a learnt a lot here. Others trying this out would be ideal. I almost have a tunnel vision on this code now, so fresh perspectives would be great. I am sure there are things we can improve upon. In terms of the flows, yes I have tabled that code for now, primarily because of the dynamic nature of the flows in `qchem`. For instance in `FFOpts`, we don't know beforehand how many iterations of Freqs and Opts we would need before starting the calculation. I think there are functionalities in jobflow which accounts for dynamic flows but I have not looked into them yet. We can have a discussion about that soon. I would eventually want to port over `FFOpt` to `atomate2` as well

rkingsbury:
> @rkingsbury thanks! It was a whirlwind, but a learnt a lot here. Others trying this out would be ideal. I almost have a tunnel vision on this code now, so fresh perspectives would be great. I am sure there are things we can improve upon. In terms of the flows, yes I have tabled that code for now, primarily because of the dynamic nature of the flows in `qchem`. For instance in `FFOpts`, we don't know beforehand how many iterations of Freqs and Opts we would need before starting the calculation. I think there are functionalities in jobflow which accounts for dynamic flows but I have not looked into them yet. We can have a discussion about that soon. I would eventually want to port over `FFOpt` to `atomate2` as well

This is a HUGE service to the MP ecosystem 🙏🏻 . I know it's been a slog. Have you been in touch with Sam about your work on this? I'm sure he'll be interested in testing as soon as it's ready, too.

Re: flows - all makes sense - I don't know either, but I think dynamic flows should be relatively straightforward to implement due to the prevalence of generators in `atomate2` code - we would basically just have to yield new Opt and Freq jobs over and over again based on a stop condition. Anyway, we can discuss more about that later on (in another PR).

Zhuoying:
Since @utf has reviewed its latest version, which has passed most of the tests, I would like to merge it as it is.
@rdguha1995 Thanks for the great PR! Very amazing job!
Please feel free to submit a new PR to increase the test coverage anytime when it is ready.

utf:
Sorry @Zhuoying, I reverted the merge.

A lot has changed since I last looked at this PR in detail. My last comments were just high level things.

 @rdguha1995 Please can you submit a new PR and I will have a detailed look over in the next few days.

Zhuoying:
@utf Sure, no problem!
@rdguha1995 We can use a new PR for more discussions!

PR #165, opened 2022-09-01
Future type annotations

janosh:
Future annotations are backwards compatible to py3.7 which is EOL Jun 2023.


utf:
Hi @janosh, thanks for this PR but it seems like it has broken the tests?

janosh:
@utf Sorry, had forgotten about this. Looks like the failure arose from the distinction between annotations [PEP 526](https://peps.python.org/pep-0526) and type hints [PEP 484](https://peps.python.org/pep-0484).


Class variables in `defect.py` 

https://github.com/materialsproject/atomate2/blob/2f4f5fab73ab4beb76cae53c1a62900845ab99be/src/atomate2/vasp/schemas/defect.py#L27-L38

are the former, i.e. annotations, not type hints. I wasn't however that they don't support future annotations below 3.10.

janosh:
Ah, maybe it's actually an issue with `pydantic`. https://github.com/pydantic/pydantic/issues/2760

Would explain why I've never seen such errors before. 😄

utf:
Still seems like there are issues? Should I close this for now?

janosh:
@utf Hang on, I think this is easy to fix.

The exact cause of the problem is well described here: https://github.com/pydantic/pydantic/issues/3300#issuecomment-1034007897. The advice there is to update to 3.10 which is not an option here so instead I'll just revert to old type hints within children of `pydantic.BaseModel`.

utf:
Thanks! Would you also mind removing the new type hints in [src/atomate2/amset/schemas.py](https://github.com/materialsproject/atomate2/pull/165#diff-ca6e44c8b7faa49c340febcadc4a819f54cd77f62af49c6214ed866b2815bfbd) and then I can merge

utf:
Great. Thank you!

Issue #166, opened 2022-09-09
BUG: accuracy of output structure


JaGeo:
**Describe the bug**
This is potentially related to a problem that occurs in the phonon workflow for one structure: BORN charges that do not reflect the correct symmetry. This then results in a wrong non-analytical term correction around Gamma.

I start with this POSCAR:
```
Mg3 Sb2
1.0
   2.3003714889113018   -3.9843602950772405    0.0000000000000000
   2.3003714889113018    3.9843602950772405    0.0000000000000000
   0.0000000000000000    0.0000000000000000    7.2813299999999996
Mg Sb
3 2
direct
   0.0000000000000000    0.0000000000000000    0.0000000000000000 Mg2+
   0.3333333333333333    0.6666666666666666    0.3683250000000000 Mg2+
   0.6666666666666667    0.3333333333333334    0.6316750000000000 Mg2+
   0.3333333333333333    0.6666666666666666    0.7747490000000000 Sb3-
   0.6666666666666667    0.3333333333333334    0.2252510000000000 Sb3-
````

Then, I run the following workfow:

```Python
from atomate2.vasp.jobs.core import DielectricMaker, TightRelaxMaker
from pymatgen.core.structure import Structure
from jobflow import run_locally
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
from atomate2.vasp.powerups import update_user_incar_settings
from jobflow import Flow
flow=[]
struct=Structure.from_file("POSCAR")
job=TightRelaxMaker().make(structure=struct)
job2=DielectricMaker().make(structure=job.output.structure)
flow=Flow([job,job2],job2.output)
flow = update_user_incar_settings(flow, {"NPAR": 8}, class_filter=TightRelaxMaker)
run_locally(flow, create_folders=True)
```

The structure after the tight relaxation looks like this:
```
Mg3 Sb2                                 
   1.0000000000000000     
     2.2774694095435986   -3.9446927300134034    0.0000000000000000
     2.2774694095435986    3.9446927300134034   -0.0000000000000000
    -0.0000000000000000   -0.0000000000000000    7.1854093835379320
   Mg   Sb
     3     2
Direct
 -0.0000000000000000 -0.0000000000000000 -0.0000000000000000
  0.3333333333333357  0.6666666666666643  0.3675316159328170
  0.6666666666666643  0.3333333333333357  0.6324683840671832
  0.3333333333333357  0.6666666666666643  0.7761920492952012
[...]

Issue #156, opened 2022-07-14
Standardized approach for job checkpoint/restart/continuation


jmmshn:
Hi @mkhorton @davidwaroquiers @mjwen. Agreed that we need a proper discussion about restarting. This is something @jmmshn has opinions about too.

I have some ideas for how we could do it but each with their own tradeoffs. I'd be interested to hear if there were any good ideas from your workshop @mkhorton.

I will create a separate issue to discuss this further.

_Originally posted by @utf in https://github.com/materialsproject/atomate2/issues/134#issuecomment-1184215723_

jmmshn:
I also posted this on the forum but I don't think that shows up here:
https://matsci.org/t/atomate2-restart-for-long-running-hse-jobs/42345

jmmshn:
@mkhorton so I think there are two different problems to solve:
1. copying contcar to poscar when restarting relaxation runs.
2. checkpointing of the actualy vasp calculation

The majority of compute time that I have been wasting for the last couple of years has been on long running relaxation jobs where restarting a failed relaxation always starts from the the initial structure so I think all relaxations jobs that did not finish can be considered in a "checkpointed" state even though they don't engage in with any formal checkpointing system other than the CONTCAR file. 
Solving 1. would fix the problem will lenthy relaxations but would not really help MD runs (unless they can be stiched together?)

But that would basically solve all of my problems now.

mkhorton:
I hear you, if we just want to concern ourselves with contcars/relaxations, it's a much easier problem to solve (questions of wasted compute due to badly-progressing optimizations aside), and dynamic addition of additional jobs seems the way to go. But I think the question remains of how to formalize this, what pattern to adopt, etc?

jmmshn:
Yeah, I think we should have a quick chat about this as a group next week?

mjwen:
I feel a big part is how to formalize it such that multiple jobflow jobs can be regarded as one, and later builders do not need to worry about how to stitch them together.

jmmshn:
> I feel a big part is how to formalize it such that multiple jobflow jobs can be regarded as one, and later builders do not need to worry about how to stitch them together. 

This also gets complicated by the fact that (I assume) most "restarts" will be initialized from `lpad rerun_wfs` so maybe we just need a restart flag like `ISTART` in VASP that dictates how the calculation should behave given different available data in the directory.  This can be something added to all the VASP makers.

EX. 
[...]

PR #167, opened 2022-09-12
Fix precision error

JaGeo:
## Summary
This is an WIP attempt to fix the precision error mentioned in https://github.com/materialsproject/atomate2/issues/166 (@utf )

It currently has tons of test failures as one can imagine. I think we also need to recreate a huge part of the test data. I am not sure I have even all programs required to do so. I would be very happy about help in this regard. All VASP flows fail and further tests as I added the contcar to some methods/functions.


src/atomate2/vasp/schemas/calculation.py
```diff
@@ -493,7 +497,8 @@ def from_vasp_outputs(
         outcar_dict = outcar.as_dict()
         outcar_dict.pop("run_stats")
 
-        structure = vasprun.final_structure
+        # structure = vasprun.final_structure
```
utf, line 500: Can you change this line to:
```suggestion
        # use structure from CONTCAR as it is written to greater precision than in the vasprun
```

utf (commented):
Thanks! 

A couple of minor comments now. Will take another look after the tests run.

src/atomate2/vasp/schemas/task.py
```diff
@@ -649,8 +650,11 @@ def _get_task_files(files, suffix=""):
         for file in files:
             if file.match(f"*vasprun.xml{suffix}*"):
                 vasp_files["vasprun_file"] = file
+                vasp_files["vasprun_file"] = file
```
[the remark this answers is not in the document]
JaGeo: No, you are correct, of course. Will remove it.

JaGeo:
I have added the suggestions. As this is quite a drastic change, I am happy about any kind of suggestion. I might also be overlooking something with regard to CONTCAR fetching ...

utf:
So hopefully we won't have to re-generate the test data. It looks at first glance that the fixes should be relatively simple.

Most of the errors look like an issue with sorting the contcars?

```
else sorted(contcars)[-1]
IndexError: list index out of range
```
Does your code definitely work with gzipped contacts for example?

In the schema tests (`atomate2/tests/vasp/schemas`) there are many:
```
        files = test_object.task_files[task_name]
>       calc_doc, _ = Calculation.from_vasp_files(dir_name, task_name, **files)
E       TypeError: from_vasp_files() missing 1 required positional argument: 'contcar_file'
```

These can be easily addressed by adding the path to the contcar file to the test dicts, e.g., [here](https://github.com/materialsproject/atomate2/blob/4dbeccb98f12968ba294cb34af345f03c7a497a1/tests/vasp/schemas/conftest.py#L49) and any other `task_files` dicts in that file.

Similarly, there are errors such as:
```
        test_object = get_test_object(object_name)
        folder = vasp_test_dir / test_object.folder / "outputs"
        vasprun_file = folder / test_object.task_files[task_name]["vasprun_file"]
        outcar_file = folder / test_object.task_files[task_name]["outcar_file"]
        vasprun = Vasprun(vasprun_file)
        outcar = Outcar(outcar_file)
>       test_doc = CalculationOutput.from_vasp_outputs(vasprun, outcar)
E       TypeError: from_vasp_outputs() missing 1 required positional argument: 'contcar'
```
These can be fixed by adding code such as:
```
        contcar_file = folder / test_object.task_files[task_name]["contcar_file"]
        contcar = Poscar.from_file(contcar_file)
        test_doc = CalculationOutput.from_vasp_outputs(vasprun, outcar, contcar)
```

JaGeo:
I don't think it has anything to do with the gzipping. I immitated the OUTCAR and vasprun.xml reading exactly. 
Files are copied before via vasp.files and it wasn't copying the CONTCAR. 
Now, the existing CONTCAR however leads to different errors. I will further look into it and hopefully have it resolved by the end of the day.

JaGeo:
Found the issue. The CONTCAR is already a renamed POSCAR here. This is then used to read the structure ... I think I fixed some more issues ...

JaGeo:
Once all tests are passing, I will do a double relaxation to check that POSCAR and CONTCAR copying is not somehow messed up now. Would be happy if you could have a look as well as I am worried to have overlooked something.

JaGeo:
Should we add an additional tag to check if CONTCAR has been copied to POSCAR before reading? I am a bit unsure about the best approach here as I feel this could lead to a lot of problems as copy_vasp_outputs has the option to not copy contcar to poscar.

utf:
Thinking about this a bit more, the structure the each calculation uses never actually comes from the POSCAR/CONTCAR in the current directory. Instead it comes from the structure that gets passed in to the make function. E.g., it comes from the structure that is stored in the output of the previous job.

The copying CONTCAR from previous directory stuff is hangover from atomate1 where the structure was passed by copying files. One option is remove the copying, although it could potentially be useful when we implement checkpointing functionality (see #156).

Maybe for now we can leave it in there. Either way, that bit of the code will not impact which structure gets used.

JaGeo:
Okay, great. 
I will be rerunning my phonon workflow for Mg3Sb2 now and check if all BORN charge symmetry problems have now vanished. I hope they have ...

utf:
Thanks for being so thorough! Hope it fixes this issue too.

JaGeo:
Born charges are computed correctly but there is still another issue in the phonon workflow for this structure... Will continue checking. ..

utf:
Does this fix the Born problem? Am I ok to merge?

JaGeo:
I think it fixes the general issue that the precision of the CONTCAR files gets lost within a flow which could influence the results. Another issue is now actually showing up in the phonon workflow that wasn't there before. I am currently fixing it.... Sorry for this. These problems did not show up in the very symmetric structures ...

tests/vasp/conftest.py
```diff
@@ -256,6 +257,7 @@ def copy_vasp_outputs(ref_path: Union[str, Path]):
     import shutil
 
     output_path = ref_path / "outputs"
+    print(output_path.iterdir())
```
utf, line 260: Please remove

utf (commented):
Its great that you're being so thorough with testing the workflow. I'm very grateful that you've caught these bugs now rather than further down the line. 

If you don't mind, I will merge this PR now (once that stray debugging line is removed), as to avoid complicating the PR.

tests/vasp/conftest.py
JaGeo: Will do after a meeting :)

utf:
Thank you :)

PR #169, opened 2022-09-14
Flow settings not correctly updated in the documentation

JaGeo:
## Summary

By accident, I recently also had this error in my code. I think it originated from this part of the documentation. It's correct in the first example but not here. `update_user_kpoints_settings` will return a new flow and it would get lost in this way.


utf:
Sorry about that!

JaGeo:
Thanks. And no worries.

PR #170, opened 2022-09-15
Fix born error in phonon workflow

JaGeo:
## Summary
This is a work in progress pull-request to fix the born charge error in the phonon workflow. This current version is producing the following result (*.json) but I need to add additional tests and fix documentation and so on. I wanted to raise this pull-request so that people are aware there is still an issue in the phonon workflow.

The agreement with the phonon data on the MP website is very good. I checked at Gamma as well to confirm the NAC as well. The differences are smaller than 1 cm-1.

@utf, I will let you know once it is ready for review.

[Mg3Sb2.zip](https://github.com/materialsproject/atomate2/files/9574642/Mg3Sb2.zip)


JaGeo:
I think I have fixed the issue. I did not realize that phonopy only converts the BORN charges to the symetrically inequivalent ones for file size reasons. This lead to issues for more complex structures. I have now replaced this reduction with a function for symmetrization only. I have tested it for Mg3Sb2 (see files above) and the result is in very good agreement with the MP website. 

To make sure that such more complex structures are also covered by the tests, I have extended them.

**Next stage of testing:** people in my group will work with the workflow and will try to reproduce some more of their results (especially also information on total energies, free energies) to check that I did not make a logical error somewhere. As the workflow and the code is so complex, I still fear this is likely.

utf (commented):
Seems like a simple fix in the end (but I can imagine it was a lot more debugging!).

Thanks for catching this.

JaGeo:
Great thanks! I think I would have been faster without the three days of a conference 😅 but I also checked the order of the forces as a first step and had to produce new test data.

PR #175, opened 2022-10-04
Small correction to contributors

JaGeo:
## Summary

Tiny correction to contributors page (orcid and affiliation). @utf, thank for updating the contributors.


utf:
Ah thanks for catching those.

PR #176, opened 2022-10-05
Job to retrieve `Structure` from MP API at run-time

mkhorton:
## Summary

@utf would you be ok with this being added, or have any thoughts/concerns?

Motivation is setting up a lot of Flows in advance (e.g., potentially to be run over several months via FireWorks), and wanting to ensure that the latest `Structure` object is always retrieved from MP at the time the Flow is run, instead of ahead of time and baking it in, such that Flows do not have to be re-generated when a new MP database is released. Database version and `task_id` are also stored for provenance. This is intended to help with production for downstream properties.

* Optional dependency `mp-api` added

## TODO (if any)

- [ ] Tests -- where's an appropriate place for tests for this? I saw `symmetrize_structure` is also not tested currently or I'd have put one alongside that.

## Checklist

title.

- [x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  your local machine. Start with running [black](
  https://black.readthedocs.io/en/stable/index.html) on your new code. This will
  [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/), followed by [flake8](
  http://flake8.pycqa.org/en/latest/).
- [x] Docstrings have been added in the[Numpy docstring format](
  https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [x] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.
- [ ] Tests have been added for any new functionality or bug fixes.
- [ ] All linting and tests pass.


utf:
This looks great. I have two quick comments:

1. Can you add mp-api and a version number to the "strict" requirements section so that this will be installed with the tests.
2. You can add the test to test/common/jobs.py. We should a a test for structure symmetrisation later.

pyproject.toml
```diff
@@ -61,6 +61,7 @@ strict = [
     "phonopy==2.16.2",
     "seekpath==2.0.1",
     "numpy",
+    "mp-api>=0.27.5"
```
utf, line 64: Can this one be pinned to a specific version (e.g., `==`) as it makes the testing environment reproducible.

mkhorton: Sure thing -- fair warning, this package is still being updated quite significantly until its "official release". For the functionality we're relying on here I expect it to be stable, but we'll probably want to keep this updated to the latest version fairly regularly.

mkhorton:
Ok, PR updated. Added tests for the other `common.jobs` since it felt rude not to!

For the API test, do you want this mocked? Without mocking, it may introduce some brittleness to the tests, but with mocking we might miss when/if this breaks, so maintainer's choice :)

pyproject.toml
utf: Thanks for the heads up. Hopefully dependabot should keep this in check.

utf:
Thanks Matt, really helpful! I'm happy with the live tests to be honest. Once tests pass I'll merge this.

utf:
Great, thank you!

PR #177, opened 2022-10-06
correct kpoint density in phonon workflow

JaGeo:
## Summary

I have found another small bug in the workflow. The kpoint density has to be computed relative to the primitive cell detected or used in phonopy.


utf:
Good catch!

JaGeo:
Yeah, I was partially re-using the code for another project and saw the inconsistency! As I said, we are still in the testing phase as I had made soo many changes to the original implementation...

I hope the free energies did not change too much due to this change. Might have to adapt the tests in a next step in case they do change in the relevant digits ....

JaGeo:
Looks like the failures are coming from another implmentation. Locally, the phonon workflow tests run.

mkhorton:
Hm, it's definitely just because of the missing API key -- I added the `MP_API_KEY` as a repository secret, but perhaps that's not accessible via a PR for security reasons?

mkhorton:
Ok, so use of these secrets in actions is limited to people with official `collaborator` status on the repo. We can add @JaGeo of course, but I'm still trying to find best practices for other contributors.

mkhorton:
I think the cleanest and best fix will simply be to skip the relevant test if `MP_API_KEY` environment variable isn't set.

```
@pytest.mark.skipif(not os.environ.get("MP_API_KEY"))
```

(Not sure if the env variable will be completely absent, or just empty, but this should work.)

I can submit PR if preferred, or just add it to this PR?

utf:
Thanks for looking into it. Happy to accept a PR with that behaviour.

utf:
Thanks @JaGeo

PR #180, opened 2022-10-13
Patch: expanduser path

nwinner:
Most interpreters will not automatically recognize the tilde. Use pathlib to expand the tilde into the absolute path for the default config file.


utf:
Great thanks.

Issue #171, opened 2022-09-20
Clarify calcs_reversed in atomate2


Zhuoying:
Previously in the original atomate, `calcs_reversed` key is explicitly named for putting the final step as the first entry (https://github.com/hackingmaterials/atomate/blob/4577c94c1850fcebd8624d49ba1ab27f8e3e6d43/atomate/vasp/drones.py#L300)

In atomate2, the calculations are not in a reversed order, but still use this `calcs_reversed` key name, which may cause some confusions.
https://github.com/materialsproject/atomate2/blob/fe657b421a7d3b1a861da87f6d5994f3950c593f/src/atomate2/vasp/schemas/task.py#L382

To resolve this issue, we could:
(1) doing the calculation list in a reversed order, obey the principle of making the final step as first entry
 OR 
(2) keep the calcs in normal sequence (put initial step as the first entry), but rename it with a different key.

Please let me know which way you pro and I could submit a PR later.

utf:
Hi @Zhuoying, thanks for bringing this to my attention. This should definitely be fixed. The best option is option 1 you suggested. If you could submit a PR that would be fantastic.

utf:
Fixed in #182

PR #182, opened 2022-10-14
Correct calcs_reversed to real reverse order

Zhuoying:
## Summary

Follow-up PR from Issue <https://github.com/materialsproject/atomate2/issues/171>:

* Correct `calcs_reversed` in reverse order and make it unified throughout in 
* Unify `calcs_reversed` throughout and drop the positive order docs of `calc_docs`
* Fix tests


utf:
This looks great! Thanks very much @Zhuoying.

PR #187, opened 2022-10-30
[WIP] Added waveder support

jmmshn:
## Added support to store Wavder objects in pymatgen

The `Wavder` object is helpful for a host of optical analysis functions so I'm trying to figure out how to store it properly.
I'm of the opinion that by default people will not want to use it for general optical workflows but will like to have it for special cases.

@utf do you think just adding a basic parser like this is OK, also do you have any opinion on how test files should be added?
These files are quite large so I was thinking about spoofing something in the current optics test folders.

### DIFF
--- a/src/atomate2/vasp/schemas/calculation.py
+++ b/src/atomate2/vasp/schemas/calculation.py
@@ -25,6 +25,7 @@ from pymatgen.io.vasp import (
     PotcarSingle,
     Vasprun,
     VolumetricData,
+    Waveder,
 )
 
 from atomate2 import SETTINGS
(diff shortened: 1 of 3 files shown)


utf:
Hi Jimmy, sorry I didn't get back to you about this PR. I think spoofing something very small in the optics directory would be great.

src/atomate2/vasp/schemas/calculation.py
```diff
@@ -828,6 +840,26 @@ def _get_volumetric_data(
...
+        The directory containing the WAVEDER file.
+
+    Returns
+    -------
+    Optional[Waveder]
+        The WAVEDER data.
+    """
+    try:
+        return Waveder.from_binary(dir_name / "WAVEDER")
+    except Exception:
```
utf, line 859: Could we not catch all exceptions?

utf:
Hi @jmmshn, what's the status of this PR?

esoteric-ephemera:
Hey @jmmshn I see this is stale and would require some changes in emmet to work now. If you still want to implement this, happy to guide those changes across packages, otherwise I'll close this out

PR #197, opened 2022-11-03
WIP: Lobster workflow

JaGeo:
## Summary

This is basically mostly a reminder that @naik-aakash have started to work on the Lobster integration. It will be based on the atomate implementation but I think we can go a step further than this and also include pre-convergence steps for speed-up and also include our lobsterpy tools to summarize the bonding analysis data better. 

## TODO

There are still many, many things to do. I plan to work on this in free minutes to get this finished until the end of this year. I will notify you @utf when I think it would be ready for a first review. This will add the feature proposed in issue https://github.com/materialsproject/atomate2/issues/160

## Checklist

Before a pull request can be merged, the following items must be checked:

- [ ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  The easiest way to handle this is to run the following in the **correct sequence** on
  your local machine. Start with running [black](
  https://black.readthedocs.io/en/stable/index.html) on your new code. This will
  automatically reformat your code to PEP8 conventions and removes most issues. Then run
  [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/), followed by [flake8](
  http://flake8.pycqa.org/en/latest/).
- [ ] Docstrings have been added in the[Numpy docstring format](
  https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [ ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.
- [ ] Tests have been added for any new functionality or bug fixes.
- [ ] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more
efficient if you already fix most errors prior to submitting the PR. It is highly
recommended that you use the pre-commit hook provided in the repository. Simply run
`pre-commit install` and a check will be run prior to allowing commits.

PR #200, opened 2022-11-09
Lobster workflow 

JaGeo:
## Summary

This is now a clean version of the Lobster workflow draft that @naik-aakash and I will continue working on. I had to close https://github.com/materialsproject/atomate2/pull/197


src/atomate2/lobster/schemas.py
```diff
@@ -149,6 +151,18 @@ class CondensedBondingAnalysis(BaseModel):
         description="Stores the COHP plot data based on relevant bond labels "
         "for site as keys",
     )
+    whichbonds: str = Field(
```
JaGeo, line 154: _ ... 😉

naik-aakash: Ops. Sorry will fix this. I by mistake ended up merging it

JaGeo:
**Todo:**

- [ ] Adapt all default values (especially k-points)
- [ ] Refinement of schemas
- [ ] Fix documentation and linting

JaGeo:
@utf We now have a version of workflow, schema and so on that could be reviewed as soon as you have time. It performs all necessary computations and produces a summary of the computations in the database.

The current workflow will run a bulk optimization, a pre-convergence step of the WAVECAR (with symmetry and default number of bands) will follow, then the typical run required for Lobster (without symmetry, correct number of bands for subsequent lobster runs) will be performed, and then a test of all potential basis sets in Lobster projections will be done. Afterwards, all WAVECARs will be deleted to save disk space.  

I/We (@naik-aakash) would potentially like to add some automatic output plots and additional features while including your review requests.

src/atomate2/lobster/jobs.py
```diff
@@ -0,0 +1,90 @@
...
+    resubmit: bool = False
+    task_document_kwargs: dict = field(default_factory=dict)
+
+    @job(output_schema=LobsterTaskDocument, data=[CompleteCohp, LobsterCompleteDos])
+    def make(
+        self,
+        wavefunction_dir: str | Path = None,
+        basis_dict: dict | None = None,
+        user_lobsterin_settings: dict | None = None,
+        additional_outputs: list[str] | None = None,
```
utf, line 49: I'm just wondering about the motivation for putting `user_lobsterin_settings` and `additional_outputs` here rather than as fields on the Maker class? Normally, I'd put arguments in the `make` function if they will change for each job (e.g., depending on the structure or basis set). But it seems like these two parameters are reasonably fixed. 

If you define them on this Maker class, this would allow you to remove them from the Flow maker class and avoid duplication of parameters. 

Just a thought, happy to keep it if you prefer it this way round.

src/atomate2/lobster/schemas.py
```diff
@@ -0,0 +1,585 @@
...
+    Doscar,
+    Icohplist,
+    Lobsterin,
+    Lobsterout,
+    MadelungEnergies,
+)
+
+from atomate2 import __version__
+
+# from atomate2.common.schemas.structure import StructureMetadata
```
utf, line 28: Remove commented out imports.

src/atomate2/lobster/schemas.py
```diff
@@ -0,0 +1,585 @@
...
+__all__ = ["LobsterTaskDocument"]
+
+logger = logging.getLogger(__name__)
+
+
+class LobsteroutModel(BaseModel):
+    """Collection to store computational settings from the LOBSTER computation."""
+
+    restart_from_projection: bool = Field(
+        "Bool indicating if the run has been restarted from a projection"
```
utf, line 52: Are you not missing the default value (`None`) and `description=` in this model? IIRC, the None is needed so that you can make the model without all the fields present. This is useful if you query the outputs database for only a few of the keys in the model, as otherwise there will be an error when you try and instantiate the model without all the keys present.

src/atomate2/lobster/schemas.py
```diff
@@ -0,0 +1,585 @@
...
+            filename=os.path.join(dir_name, "ICOOPLIST.lobster.gz"),
+            are_coops=True,
+            are_cobis=False,
+        )
+
+        icohp_dict = icohplist.icohpcollection.as_dict()
+        icobi_dict = icobilist.icohpcollection.as_dict()
+        icoop_dict = icooplist.icohpcollection.as_dict()
+
+        def get_strng_bonds(bondlist, are_cobis, are_coops, relevant_bonds: dict):
```
utf, line 388: The `from_directory` is already quite long. Can you put this as a private method outside of `from_directory` to make this a bit easier to read?

src/atomate2/lobster/schemas.py
```diff
@@ -0,0 +1,585 @@
...
+            coop_data=coop_obj,
+            cobi_data=cobi_obj,
+            dos=dos,
+            lso_dos=lso_dos,
+            charges=charges,
+            madelung_energies=madelung_energies,
+        )  # doc
+
+
+def _get_structure() -> Structure:
```
utf, line 573: Is this needed or just copy and paste from amset schemas file?

src/atomate2/vasp/flows/lobster.py
```diff
@@ -0,0 +1,194 @@
...
+        A maker to perform a relaxation on the bulk.
+        Set to ``None`` to skip the
+        bulk relaxation
+    additional_static_run_maker: .BaseVaspMaker or None
+        A maker to perform a preconvergence run
+        before the wavefunction computation without symmetry
+    vasp_lobster_maker : .BaseVaspMaker
+        A maker to perform the computation of the wavefunction before the static run.
+        Cannot be skipped.
+    code: str
```
utf, line 50: Can remove this I think, doesn't seem to be implemented.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,301 @@
...
+        basis = [key + " " + value for key, value in dict_for_basis.items()]
+        lobsterin = Lobsterin(settingsdict={"basisfunctions": basis})
+        nbands = lobsterin._get_nbands(structure=structure)
+        nband_list.append(nbands)
+
+    return {"nbands": max(nband_list), "basis_dict": list_basis_dict}
+
+
+@job
+def update_user_incar_settings_maker(
```
utf, line 142: Ah! I see you ran into this issue as well. Which is that can't put output references in the class fields for Makers. Kind of annoying and a limitation of jobflow. I might try and fix this in the future.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,301 @@
...
+
+@job
+def get_basis_infos(
+    structure: Structure,
+    vaspmaker: BaseVaspMaker,
+    address_max_basis: str = None,
+    address_min_basis: str = None,
+):
+    """
+
```
utf, line 91: Can you use numpy style docstrings here and add an initial line saying what the function does.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,301 @@
...
+@job
+def update_user_incar_settings_maker(
+    vaspmaker: VaspLobsterMaker,
+    nbands: dict,
+    structure: Structure,
+    prev_vasp_dir: Path | str,
+):
+    """
+    Update the INCAR settings of a maker
+    Args:
```
utf, line 150: Can you use numpy style docstrings please.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,301 @@
...
+    additional_outputs,
+    optimization_run_job_dir,
+    optimization_run_uuid,
+    static_run_job_dir,
+    static_run_uuid,
+    additional_static_run_job_dir,
+    additional_static_run_uuid,
+):
+    """
+
```
utf, line 179: Also numpy style docstrings here also.

src/atomate2/vasp/flows/lobster.py
```diff
@@ -0,0 +1,194 @@
...
+    calculation_type:
+        "standard" for a normal Lobster run
+    delete_all_wavecars:
+        if true, all WAVECARs will be deleated after the run
+    user_lobsterin_settings:
+        Dict including additional information
+    user_supplied_basis: dict or None
+        dict to supply a basis for the lobsterrun
+    additional_outputs:
+        list of filenames to include in output (str)
```
utf, line 63: See previous comment about these fields. My alternative suggestion would be to replace these with a `pure_lobster_maker` field instead which would hold the fields instead of the flow maker. Again, might not suitable in this case but please consider it.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,301 @@
...
+
+@job
+def delete_lobster_wavecar(
+    dirs: list, dir_vasp: Path | str = None, dir_preconverge: Path | str = None
+):
+    """
+    Deletes all WAVECARs
+
+    Args:
```
utf, line 230: Numpy style please.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,301 @@
...
+@job(output_schema=LobsterTaskDocument)
+def generate_database_entry(
+    **kwargs,
+):
+    """
+    Analyze the LOBSTER runs and summarize the results.
+
+    Parameters
+    ----------
+
```
utf, line 292: Can you tidy up the blank lines?

utf (commented):
Hi @JaGeo, thanks very much for this! It looks great. Basically all of my comments are minor docstring fixes.

Thanks for adding this feature!

JaGeo:
Thank you very much for the comments, @utf ! Wewill probably try to finish it within the next two weeks.

src/atomate2/lobster/files.py
```diff
@@ -0,0 +1,101 @@
...
+        file_client=file_client,
+    )
+
+    gunzip_files(
+        include_files=files,
+        allow_missing=True,
+        file_client=file_client,
+    )
+
+    # rename_files({"transport.json": "transport.prev.json"}, allow_missing=True)
```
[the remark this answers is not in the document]
JaGeo: Done

src/atomate2/lobster/jobs.py
JaGeo: Thanks for the suggestion. You are probably correct that this would reflect the overall logic of atomate2 much better. I have to check if I can implement it without breaking some functionality of the workflow again.

JaGeo: Implemented.

src/atomate2/lobster/schemas.py
JaGeo: Refactored.

src/atomate2/vasp/jobs/lobster.py
JaGeo: Yes, of course. I am so used to google that I forgot that atomate2 is using numpy.

src/atomate2/lobster/schemas.py
JaGeo: Yes, it was a copy and paste error. Sorry!

JaGeo:
I will add default values for the schma and then I will try to include some automatic output plots/jsons as well (similar to phonon workflow).

src/atomate2/lobster/schemas.py
utf: I think the default value should actually be None for everything here. That allows pydantic to reconstruct the model even with fields missing. E.g. if you only query part of the model from the database

JaGeo: Okay, good! I will add this! Thank you.

JaGeo:
I will test the workflow once more and check all output files tonight and let you know if everything is fine from my side. I also added automatic output plots for cohps now.

src/atomate2/vasp/jobs/lobster.py
utf: Sorry to be a pain, but there is a slight issue with this and some other docstrings.  The format should be like:

```python
    """
    Parameters
    ----------
    name: type
        Description.
    """
```

JaGeo: Thanks. It's very helpful that you check so carefully!

JaGeo:
Doc strings should hopefully be fixed now. The automatic plots are also working.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -0,0 +1,332 @@
...
+        )
+        outputs["add_static_dir_name"].append(dir_vasp_add_stat)
+        outputs["static_dir_name"].append(dir_vasp_stat)
+
+    flow = Flow(jobs, output=outputs)
+    return Response(replace=flow)
+
+
+@job(output_schema=LobsterTaskDocument)
+def generate_database_entry(
```
utf, line 317: Is this needed? I can't see it used anywhere?

utf (commented):
Just one final question and then I think its ready to merge. It looks great!

src/atomate2/vasp/flows/lobster.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    additional_static_run_maker: BaseVaspMaker | None = field(
+        default_factory=lambda: StaticMaker(
+            input_set_generator=StaticSetGenerator(
+                user_incar_settings={"LWAVE": True},
+                user_kpoints_settings={"grid_density": 6000},
+            ),
+        )
+    )
+    lobstermaker: BaseVaspMaker | None = PureLobsterMaker()
```
utf, line 76: Just noticed this, you need to use the default_factory and please can you add this to the docstring.
```suggestion
    lobstermaker: BaseVaspMaker | None = field(
        default_factory=lambda: PureLobsterMaker()
    )
```

utf:
I had one last thought regarding the names of the Makers which could be a bit confusing. What do you think about the following renaming:

- `PureLobsterMaker` -> `LobsterMaker` (runs lobster)
- `LobsterMaker` -> `VaspLobsterMaker` (runs the VASP+Lobster workflow)
- `VaspLobsterMaker` -> `NoSymStaticMaker`

These are just suggestions, feel free to ignore

JaGeo:
Thanks! Will do! 
We will also now use these workflows in production runs to make sure I haven't overlooked anything.

src/atomate2/vasp/jobs/lobster.py
JaGeo: Yeah, you are correct. I think this is from an earlier version of the workflow and I forgot to remove it. Sorry!

JaGeo:
I believe I have addressed all comments from your side.

JaGeo:
In case you think it is ready to merge, it would be great if you could then release another atomate2 version.

utf:
Thanks @JaGeo and @naik-aakash. I'll merge this now and push a new version.

utf:
Inevitably I noticed something after I merged and released!

Currently it is not possible to configure the run_lobster command, e.g. by changing the run type. To fix this you should add a `run_lobster_kwargs` field to the `LobsterMaker` class and pass them through here:
https://github.com/materialsproject/atomate2/blob/4668fc35555d67355d559864714b78a0698cd11a/src/atomate2/lobster/jobs.py#L89

JaGeo:
I see! Thanks and sorry! I will work on a fix tomorrow! It's just so many kwargs and things to consider when developing such a workflow.

utf:
No worries! I'll submit a PR in a minute. It would be great if you can have a look over when you get the chance.

Issue #211, opened 2022-11-28
BUG: VaspObject name case sensitive


jmmshn:
https://github.com/materialsproject/atomate2/blob/02c7705609253f83c993e698f66cab0586cfff14/src/atomate2/vasp/schemas/calculation.py#L811

Something needs to change here about the case sensitivity:
Currently, the `name` attribute returns "LOCPOT" but the comments say we should use lower cases eg. `{"store_volumetric_data": ["locpot"]}`

PR #212, opened 2022-11-29
Bugfix for `store_volumetric_data`

jmmshn:
## Fixed bug when trying to store volumetric data 

The volumetric data was not actually stored in the additional store when you use the code as instructed.
This should fix it.

A test will be available in the defect workflow.
Although @utf  if you know where to stick an easy test case for this in the existing core tests please feel free to add something.

This should fix issue #211

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
     "Topic :: Other/Nonlisted Topic",
     "Topic :: Scientific/Engineering",
 ]
-requires-python = '>="3.8"'
+requires-python = '>=3.8'
 dependencies =[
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
     "Topic :: Other/Nonlisted Topic",
     "Topic :: Scientific/Engineering",
 ]
-requires-python = '>="3.8"'
+requires-python = '>=3.8'
```
jmmshn, line 26: This fixes some weird upstream bug for parsing pyproject.

utf: I just made this change in main!

PR #214, opened 2022-12-02
Defect refactor only

jmmshn:
## Refactor defect code to common

Since we are trying to combine efforts for different codes I'm gonna strip out the code-agnostic stuff to make integration a bit easier so we don't have to maintain more than 1 version.

The `schemas` `jobs` and `flows` modules have been placed in `common.analysis.defect`
The 1-D finite different derivative stuff such as the WSWQ file is VASP specific so that will live under the `vasp` subdirectory.

I will not copy over any Formation energy code for now just a refactor of the CCD code.
Once the new location is locked in we can close this PR and add the formation energy code.


utf:
Thanks @jmmshn, looks good!

PR #215, opened 2022-12-02
VASP Defect Code

jmmshn:
# VASP Defect Code

Once the general defect code is merged from #214 we can look at the VASP-specific implementation of the defect workflow will just instantiate the general one with some VASP-specific defaults.

I removed the final aggregation step since I believe that's best done with an additional database for the competing phases. 
So the current workflow is only responsible for spawning the defect calculations.


src/atomate2/vasp/schemas/calculation.py
```diff
@@ -672,7 +672,7 @@ def from_vasp_files(
 
         vasprun_kwargs = vasprun_kwargs if vasprun_kwargs else {}
         volumetric_files = [] if volumetric_files is None else volumetric_files
-        vasprun = Vasprun(vasprun_file, **vasprun_kwargs)
+        vasprun = Vasprun(vasprun_file, parse_potcar_file=True, **vasprun_kwargs)
```
jmmshn, line 675: So what I said about `Structure.charge` vs `Structure._charge` was wrong.
This was the culprit of my HSE caculations running with the wrong charge states.
I think this setting should always be one for vasprun parsing since we are running in a directory that has POTCARs

jmmshn:
For Posterity, something like the function below is needed for now to perform the postprocessing without periodic regridding
since it is possible for grid size to change if you relax the bulk structure with `ISIF=3`.  If you run with `ISIF=2` only it might get slightly different numbers.  vdw group tends to use `ISIF=3` so I'm setting that as the default behavior.

```python
def submit_defect(defect, relax_maker, bulk_dir=None, run=False):
    """Submit a defect calculation"""
    ss = MPStaticSet(defect.get_supercell_structure())
    ng, ngf = ss.calculate_ng()
    ng_settings = dict(zip(["NGX", "NGY", "NGZ", "NGXF", "NGYF", "NGZF"], ng + ngf))
    relax_maker = update_user_incar_settings(relax_maker, ng_settings)
    maker = FormationEnergyMaker(
        relax_maker=relax_maker,
    )
    flow = maker.make(defect=defect, bulk_supercell_dir=bulk_dir)
    return flow
```

src/atomate2/vasp/sets/defect.py
```diff
@@ -46,13 +181,87 @@ def get_incar_updates(
...
+            "IBRION": 1,
+            "LAECHG": False,
             "ISIF": 2,
+            "EDIFF": 1e-5,
             "EDIFFG": -0.05,
             "LREAL": False,
             "NSW": 99,
-            "LCHARG": False,
             "ENCUT": 500,
+            "NELMIN": 6,
```
utf, line 193: Need to set GGA = PE in this and other sets.

src/atomate2/vasp/sets/defect.py
```diff
@@ -46,13 +181,87 @@ def get_incar_updates(
             A dictionary of updates to apply.
         """
         return {
-            "IBRION": 2,
+            "ALGO": "All",
```
utf, line 184: Can we use ALGO = Normal for defect structures since they are large?

src/atomate2/vasp/sets/defect.py
jmmshn: added

jmmshn (commented):
updated based on comments

src/atomate2/common/analysis/defects/jobs.py
```diff
@@ -166,3 +182,183 @@ def get_ccd_documents(
...
+    relax_output: TaskDocument = relax_job.output
+    summary_d = dict(
+        uc_structure=uc_structure,
+        sc_entry=relax_output.entry,
+        sc_struct=relax_output.structure,
+        sc_mat=sc_mat.tolist(),
+        dir_name=relax_output.dir_name,
+        uuid=relax_job.uuid,
+    )
+    return Response(output=summary_d, replace=[relax_job])
```
utf, line 276: Does this definitely work? I thought if a  job returns a replace, the final output will be the replace flow rather than the output in the original response.

So I would have thought you'd need to do:

```suggestion
    flow = Flow([relax_job], output=summary_d)
    return Response(replace= flow)
```

E.g., see https://github.com/materialsproject/atomate2/blob/aa0bb0737eead988d1dc31fa46d0da0f4e61bc8c/src/atomate2/vasp/jobs/elastic.py#L226-L227

src/atomate2/common/analysis/defects/jobs.py
```diff
@@ -166,3 +182,183 @@ def get_ccd_documents(
...
+        charged_output: TaskDocument = charged_relax.output
+        all_chg_outputs[qq] = {
+            "structure": charged_output.structure,
+            "entry": charged_output.entry,
+            "dir_name": charged_output.dir_name,
+            "uuid": charged_relax.uuid,
+        }
+        # TODO: check that the charge state was set correctly
+
+    return Response(output=all_chg_outputs, replace=defect_q_jobs)
```
utf, line 361: Same with this here. E.g., see https://github.com/materialsproject/atomate2/blob/aa0bb0737eead988d1dc31fa46d0da0f4e61bc8c/src/atomate2/vasp/jobs/elastic.py#L226-L227

utf (commented):
Thanks Jimmy. Just noticed one thing

src/atomate2/common/analysis/defects/jobs.py
jmmshn: You're right the outputs are all references.
It's just that they are all meant to be comsumed by other jobs so this error did not affect anything.

src/atomate2/common/analysis/defects/jobs.py
jmmshn: You're right the outputs are all references. It's just that they are all meant to be comsumed by other jobs so this error did not affect anything.  I merged your change.

src/atomate2/vasp/schemas/calculation.py
jmmshn: ^ This was wrong too it ended up being the CONTCAR copy during the Calculation parsing step.

This is fixed as part of this PR.

https://github.com/materialsproject/atomate2/blob/b3f6f3ef1a3eec261bc3b32bbd1a1263a2d67c22/src/atomate2/vasp/schemas/calculation.py#L500-L504

src/atomate2/common/schemas/defects.py
```diff
@@ -1,16 +1,112 @@
...
+            bulk_entry=fed.bulk_entry,
+            defect_entries=fed.defect_entries,
+            vbm=fed.vbm,
+            band_gap=fed.band_gap,
+            pd_entries=fed.pd_entries,
+            inc_inf_values=fed.inc_inf_values,
+            **kwargs,
+        )
+
+    def as_FormationEnergyDiagram(
```
utf, line 88: To follow python style can you change this to:

```suggestion
    def as_formation_energy_diagram(
```

src/atomate2/common/schemas/defects.py
```diff
@@ -1,16 +1,112 @@
...
+    )
+
+    dielectric: Union[float, List[List[float]]] = Field(
+        None,
+        description="The dielectric constant or tensor, can be used to compute "
+        "finite-size corrections.",
+    )
+
+    @classmethod
+    def from_FormationEnergyDiagram(
```
utf, line 67: To follow python style can you change this to:

```suggestion
    def from_formation_energy_diagram(
```

utf (commented):
Sorry, noticed one more thing. And then will merge.

utf:
Fantastic, thanks Jimmy!

PR #232, opened 2023-01-16
Documentation of Phonon workflow

QuantumChemist:
## Summary

I updated the relevant files for the documentation of the phonon workflow.
Corrected the header in src/atomate2/vasp/flows/phonons.py
Added a header in src/atomate2/vasp/jobs/phonons.py and src/atomate2/vasp/schemas/phonons.py
Added a section concerning phonons in docs/user/codes/vasp.md
Added an entry in docs/user/codes/vasp-workflows.csv


utf:
Fantastic. Thank you very much.

PR #246, opened 2023-02-15
Auto_lreal

jmmshn:
## Automatically determine LREAL to keep vasp from complaining

This should not change the behavior of any workflows since they LREAL values are being set in almost every InputerGenerator but this allows us to remove that in future.


utf:
Hey Jimmy, thanks for this. Is there any downside to just setting LREAL = Auto?

The [VASP manual says](https://www.vasp.at/wiki/index.php/LREAL):

> We recommend using the real-space projection scheme for systems containing more than about 30 atoms. We also strongly recommend using only LREAL=Auto.

jmmshn:
I think the recommendation means "strongly recommend using only LREAL=Auto instead of LREAL=True"
`Auto` is doing the integration in real space.
I think the general rule is that reciprocal space integration is better but scales poorly but you should never use `LREAL=True` in modern VASP.
The threshold value of 16 here is just to replicate the internal logic of VASP to suppress the error message.

On another documentation page:

https://www.smcm.iqfr.csic.es/docs/vasp/node132.html

They use 20 instead of 30 as the threshold, so I feel it's better to go off the source code.

utf:
Thanks Jimmy. Looks good.

PR #250, opened 2023-02-21
Ruff

### LINKED ISSUE #249: Migrate linting to Ruff
Just a suggestion but one that has worked out very well for pymatgen over in https://github.com/materialsproject/pymatgen/pull/2847: we could migrate linting of atomate2 to [`ruff`](https://beta.ruff.rs).

Ruff combines the functionality of most other Python linters into 1 tool written in Rust which makes it ~100x faster than linters written in Python. It brought the `pymatgen` linting CI script's run time from 9 min down to 3 min (almost entirely spent installing deps and running `mypy` now, ruff itself only takes 1 sec).

Happy to take this on if interested.

janosh:
Closes #249.

Just to give an idea, this is what adding `ruff` linter would look like.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,11 @@ default_language_version:
   python: python3
 exclude: '^src/atomate2/vasp/schemas/calc_types/'
 repos:
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: v0.0.250
+  hooks:
+  - id: ruff
+    args: [--fix]
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.4.0
   hooks:
@@ -10,10 +15,6 @@ repos:
     args: [--remove]
   - id: end-of-file-fixer
   - id: trailing-whitespace
-- repo: https://github.com/myint/autoflake
-  rev: v2.0.0
-  hooks:
-  - id: autoflake
 - repo: https://github.com/psf/black
   rev: 22.12.0
   hooks:
@@ -24,10 +25,6 @@ repos:
   - id: blacken-docs
     additional_dependencies: [black]
     exclude: README.md
-- repo: https://github.com/pycqa/isort
-  rev: 5.12.0
-  hooks:
-  - id: isort
 - repo: https://github.com/pycqa/flake8
   rev: 6.0.0
   hooks:
@@ -63,8 +60,3 @@ repos:
   - id: codespell
     stages: [commit, commit-msg]
     args: [--ignore-words-list, 'titel,statics,ba,nd,te']
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.3.1
-  hooks:
-    - id: pyupgrade
-      args: [--py38-plus]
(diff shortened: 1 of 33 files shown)


utf:
This looks great, but could you fix the failing linting

janosh:
Definitely. Just wanted to hear your feedback first.

src/atomate2/vasp/flows/defect.py
```diff
@@ -43,6 +43,18 @@
...
+    Parameters
+    ----------
+    name: str
+        The name of the flow created by this maker.
+    relax_maker: BaseVaspMaker
+        A maker to perform the relaxation of the defect.
+    static_maker: BaseVaspMaker
+        A maker to perform the static calculation of the defect.
+    """
+
```
janosh, line 57: @utf This public class had no doc string. I added one but not sure it's correct. Could you check?

tests/conftest.py
```diff
@@ -89,7 +89,7 @@ def memory_jobstore():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def log_to_stdout():
+def log_to_stdout_auto_use():
```
janosh, line 92: This function was name-clashing with another fixture of the same name in this file. Since this one has `autouse=True`, I renamed this one since the name shouldn't matter?

janosh:
> Awesome thank you. Ruff looks excellent! I'll have to switch to it on all my projects

That's what I did. 😄  And certainly didn't regret it. The speed is incredible as is the number of linting rules it provides!

src/atomate2/vasp/flows/defect.py
utf: Looks great. Thanks

tests/conftest.py
utf: Yeah, that looks good.

.pre-commit-config.yaml
```diff
@@ -24,10 +25,6 @@ repos:
   - id: blacken-docs
     additional_dependencies: [black]
     exclude: README.md
-- repo: https://github.com/pycqa/isort
-  rev: 5.12.0
-  hooks:
-  - id: isort
 - repo: https://github.com/pycqa/flake8
```
janosh, line 28: We could prob replace `flake8` here with `ruff` too. I know `ruff` implements these plugins since I was using those as well.

```
    - pyproject-flake8==6.0.0
    - flake8-bugbear==22.12.6
    - flake8-typing-imports==1.14.0
    - flake8-docstrings==1.6.0
```

But haven't checked these two yet:

```
    - flake8-rst-docstrings==0.3.0
    - flake8-rst==0.8.0
```

janosh:
What should be `ruff`'s max line length? Bearing in mind that a number different from `black`'s 88 can cause linter fights since `ruff` also formats code.

utf:
So I think black has the behaviour that 88 is the ideal length but if it can't make the line any shorter then it's okay to go beyond that.

I think there was an option in flake-bugbear to support this behaviour: https://github.com/PyCQA/flake8-bugbear/issues/54

I'm not sure if that is also available in ruff?

janosh:
IIUC https://github.com/charliermarsh/ruff/issues/389#issuecomment-1304209976 suggests just disabling E501. Charlie decided to not support B950 for now.

janosh:
I went ahead and fixed the existing E501s. All but one case were easily refactored to avoid long lines and the last one I  added `noqa: E501`.

utf:
Thanks!

PR #259, opened 2023-03-07
[BUGFIX] Typo when restarting from from previous bulk SC calculation

jmmshn:
The current defect code errors out when you restart from a `bulk_dir`.
This fixes that problem.


utf:
Thanks!

PR #260, opened 2023-03-12
Tidy lobster workflow

utf:
I refactored some parts of the lobster workflow to make it cleaner.

1. Renamed `additional_static...` to `preconverge_static...` so that it is consistent throughout all parts of the code.
2. Renamed `bulk_relax` to just `relax` because bulk is only really needed when you're doing defects/strains etc.
3. Rewrote the delete wavecar functionality as it was wrapping delete_files in a job unnecessarily. There was no need to do the replace at this bit of the code as you were already in a job.
4. Enable changing the lobster run settings.
5. Changed the `additional_fields` option to be consistent to with how additional fields is used in the other schemas. This is mainly a way to store additional information in the schema. I added a new option `store_lso_dos` for specific storing of the DOS as it seemed that this was the only additional file supported.
7. Moved some code around in the schema file. Doing this I realised the code would break if the following condition failed:
`icohplist_path.exists() and cohpcar_path.exists() and charge_path.exists()` as struct and describe wouldn't be defined.
8. Some other minor docstring and code optimisations.

@JaGeo please could you have a look over and let me know if this is all ok?


src/atomate2/vasp/flows/lobster.py
```diff
@@ -141,49 +136,38 @@ def make(
...
             lobster_maker=self.lobster_maker,
             basis_dict=basis_infos.output["basis_dict"],
-            wavefunction_dir=vaspjob.output.dir_name,
-            optimization_run_job_dir=optimization_run_job_dir,
-            optimization_run_uuid=optimization_run_uuid,
-            static_run_job_dir=static_run_job_dir,
-            static_run_uuid=static_run_uuid,
-            additional_static_run_job_dir=additional_static_run_job_dir,
-            additional_static_run_uuid=additional_static_run_uuid,
+            wavefunction_dir=lobster_static.output.dir_name,
```
JaGeo, line 154: This looks all fine. I just realized that wavefunction_dir should be the same as static dir. We could probably simply remove it.

utf: Done!

utf:
Ok, I made that change. I also renamed `delete_all_wavecars` to `delete_wavecars` for brevity.

PR #266, opened 2023-03-13
Fix Lobster Schema

JaGeo:
## Summary

@utf , I am very sorry but I have overlooked an issue with the lobster schema. lobsterpy_data wasn't set correctly. I fixed it and also extended the tests to avoid such issues in the future.


utf:
Ah that looks like my fault! Sorry about that.

JaGeo:
I should have checked more carefully yesterday and I also should have written better tests.

PR #267, opened 2023-03-13
Add more documentation for Lobster

JaGeo:
## Summary

I have added a few more lines of documentation for Lobster.

An additional question that is related: Should we also work on tutorials for certain, more complex flows?


utf:
Thanks @JaGeo. I was also thinking we could have a separate tutorial for each advanced workflow. I should make one for amset, and it would be great if @jmmshn could think about a defects one at some point too. That will definitely increase usability and hopefully get more people using atomate2.

Issue #159, opened 2022-08-05
BUG: Optimization of NaCl


JaGeo:
**Describe the bug**
`ISMEAR=2` is set for non-metals.

I try to optimize the structure of NaCl with the following script (atomate development version):

```
from atomate2.vasp.jobs.core import RelaxMaker
from jobflow import run_locally
from pymatgen.core import Structure

from atomate2.vasp.powerups import update_user_incar_settings

structure = Structure.from_file("POSCAR.NaCl.vasp")
relax_job = RelaxMaker().make(structure)
relax_job = update_user_incar_settings(relax_job, {"NPAR": 4}, class_filter=RelaxMaker)

run_locally(relax_job, create_folders=True)
```
POSCAR.NaCl.vasp
```
Na4 Cl4
1.0
5.691694 0.000000 0.000000
0.000000 5.691694 0.000000
0.000000 0.000000 5.691694
Na Cl
4 4
direct
0.000000 0.000000 0.000000 Na+
0.000000 0.500000 0.500000 Na+
0.500000 0.000000 0.500000 Na+
0.500000 0.500000 0.000000 Na+
0.500000 0.000000 0.000000 Cl-
0.500000 0.500000 0.500000 Cl-
0.000000 0.000000 0.500000 Cl-
0.000000 0.500000 0.000000 Cl-
```

I end up with this INCAR:
```
ALGO = Fast
EDIFF = 1e-05
EDIFFG = -0.02
ENAUG = 1360
ENCUT = 680
GGA = Ps
IBRION = 2
ISIF = 3
ISMEAR = 2
ISPIN = 2
KSPACING = 0.22
LAECHG = True
LASPH = True
LCHARG = False
LELF = False
LMIXTAU = True
LORBIT = 11
LREAL = Auto
LVTOT = True
LWAVE = False
MAGMOM = 8*0.6
NELM = 200
NPAR = 4
NSW = 99
PREC = Accurate
SIGMA = 0.2
```

I think it is due the combination of this line here
https://github.com/materialsproject/atomate2/blob/73b9b3e71be8addc237c9d6cfce3dd1c09aa2ca8/src/atomate2/vasp/sets/core.py#L43

And this one here:
https://github.com/materialsproject/atomate2/blob/73b9b3e71be8addc237c9d6cfce3dd1c09aa2ca8/src/atomate2/vasp/sets/base.py#L976

[...]

PR #270, opened 2023-03-17
VASP inputset updates

utf:
The VASP input sets have been reconfigured based on user feedback. The ``auto_kspacing`` option has been removed and KSPACING is no longer used in the atomate2 input sets by default. We have returned to using ``reciprocal_density`` as in atomate1. These changes mean the k-point mesh is no longer dependent on the precise band gap of the system. Instead, there are now two k-points settings, one for insulators and one for metals. This should remove issues when changing the functional from PBEsol -> HSE, in which the band gap increases but the k-point mesh would be expected to stay the same.

Two new options have been added to the ``BaseVaspInputSetGenerator``:

- ``auto_metal_kpoints``: If true and the system is metallic, try and use `` reciprocal_density_metal`` instead of ``reciprocal_density`` for metallic systems.
- ``auto_ismear``: If true, the values for ISMEAR and SIGMA will be set automatically depending on the bandgap of the system. If the bandgap is not known (e.g., there is no previous VASP directory) then ISMEAR=0 and SIGMA=0.2; if the bandgap is zero (a metallic system) then ISMEAR=2 and SIGMA=0.2; if the system is an insulator, then ISMEAR=-5 (tetrahedron smearing).

Fixes: #159 

@JaGeo and @jmmshn please could you take a look and let me know if you think this is ok?


src/atomate2/vasp/sets/base.py
```diff
@@ -231,9 +231,17 @@ class VaspInputGenerator(InputGenerator):
         Functional to use. Default is to use the functional in the config dictionary.
         Valid values: "PBE", "PBE_52", "PBE_54", "LDA", "LDA_52", "LDA_54", "PW91",
         "LDA_US", "PW91_US".
-    auto_kspacing
-        Whether to set the kspacing value based on the band gap. Note, this only works
+    auto_metal_kpoints
+        If true and the system is metallic, try and use ``reciprocal_density_metal``
+        instead of ``reciprocal_density`` for metallic systems. Note, this only works
```
JaGeo, line 236: Would it be beneficial to have an easy way to turn on the tighter k-spacing (e.g., when you already suspect you have a metal)? Or should the user then simply set the values themselves?

utf: Personally, I think then it's ok if the user just sets in themselves. If you already know it's a metal then I don't think there is much difference between setting the reciprocal density manually vs having another flag to set.

JaGeo: The rest looks fine to me. I haven't have done a test run for phonons with it yet. I could do it later if you like.

utf: Thanks @JaGeo. I just tried running the phonon workflow and it seems to run fine.

JaGeo: Great and thanks!

jmmshn: I'm also onboard with the user-supplied metal kpoints thing since this seems more like a question about validating data after the fact so it should be taken care of in emmet.

jmmshn:
@utf what do you think about setting `auto_lreal` to `True` by default?  It stops a really aggressive warning from VASP.

JaGeo:
I always felt that this LREAL warning is from a time with much less computing time and I personally nearly always used LREAL=False. LREAL=False is much more accurate.

jmmshn:
@JaGeo Good point `LREAL=.False.` would probably be better.  Either way, we have `LREAL=.Auto.` which is still a real-space setting.  I'd be onboard with setting `LREAL=.False.` for everything.

arosen93:
I am strongly on the side of setting `LREAL=.FALSE.` for everything unless it is some intermediate calculation (e.g. the first relax in a double relax). For systems with large numbers of atoms per unit cell, I have seen concerning errors in energies with LREAL = Auto. The speedup isn't huge.

utf:
Ok, I've switched to `LREAL=False` by default.

mkhorton:
Agreed on LREAL, also caused issues on elastic workflows.

Added comments to #352 on this change; given how far reaching the change is, I think it perhaps warrants some extra discussion.

matthewkuner:
@mkhorton out of curiosity, what issues did LREAL cause with elastic workflows for you?

mkhorton:
Very bad ones! @mjwen can comment further. The MP default of `LREAL: Auto` is probably a bad one for the typical size of cells we use, and (especially) switching between LREAL settings in the context of an elastic calculation can give very bad results due to the sensitivity to stress tensor and small changes in energy etc.

mjwen:
Right, just as @mkhorton said, the fitting of elasticity tensors needs very accurate stresses, and the `LREAL=Auto` real space ones are just not good enough. In some cases, particularly the crystals with low material symmetry, this causes a lot of trouble.  

And I'm totally in favor of setting it to False as the default atomate2 value.

JaGeo:
Same for phonon runs. Forces and optimizations have to be extremely accurate. In my PhD, I never only ever used `LREAL=False`

matthewkuner:
@mjwen is that true even for larger cells? I.e. 64 atoms?

mjwen:
That I'm not sure. Haven't looked at the effects of the number of atoms.

jmmshn:
@utf So weird problem that might require a revisit to `auto_ismear` it looks like `ismear=-5` interacts really badly with `NEDOS` basically if you set NEDOS to a usable number then the calculations can never finish the post-processing.

So one way around this is to use the `nscf uniform` workflow but for w/e reason, the DOS produced from a user-generated list of KPOINTs has the wrong units when compared with VASP's automatically generated Kpoints.

https://matsci.org/t/dos-units/48565

Not sure how to address any of these others especially since I'm not even sure if this `ismear=-5` + `NEDOS` issue is even a persistent problem across different clusters.

utf:
Out of interest, what is a useable number? I've never had any issues with ISMEAR=-5 and NEDOS = 2000.

jmmshn:
That's exactly what I'm using, if that runs fine for you then it might be a problem with our binary.

src/atomate2/vasp/sets/base.py
```diff
@@ -968,59 +973,6 @@ def _remove_unused_incar_params(incar, skip=None):
...
-
-        # Ensure adequate number of KPOINTS are present for the tetrahedron method
-        # (ISMEAR=-5). If KSPACING is in the INCAR file the number of kpoints is not
-        # known before calling VASP, but a warning is raised when the KSPACING value is
-        # > 0.5 (2 reciprocal Angstrom). An error handler in Custodian is available to
-        # correct overly large KSPACING values (small number of kpoints) if necessary.
-        if np.product(kpoints.kpts) < 4 and incar.get("ISMEAR", 0) == -5:
-            incar["ISMEAR"] = 0
-
-    elif "KSPACING" in user_incar_settings:
```
matthewkuner, line 1001: @jmmshn Do you think these lines could be the cause of the test failures in #412 ?

src/atomate2/vasp/sets/base.py
```diff
@@ -638,21 +637,30 @@ def _get_incar(
...
+                incar.update({"SIGMA": 0.2, "ISMEAR": 0})
+            elif bandgap == 0:
+                incar.update({"SIGMA": 0.2, "ISMEAR": 2})  # metal
+            else:
+                incar.update({"ISMEAR": -5, "SIGMA": 0.05})  # insulator
+
+        if self.auto_lreal:
+            incar.update({"LREAL": _get_recommended_lreal(structure)})
+
+        if kpoints is not None:
```
matthewkuner, line 657: @jmmshn All of the new test failures in #412  are cases where ISMEAR is changed from -5 to 0. Might this be the culprit?

jmmshn: @matthewkuner Looks like things got fixed by @utf.   Either way the LREAL behavior is not really checke by too many things so I doubt it was the cause.

PR #272, opened 2023-03-18
Extension of Lobster schema and additional tests

JaGeo:
## Summary

This extends the Lobster schema in such a way, that always all bonds and all cation-anion bonds will be included in the automatic analysis of the COHPs. Furthermore, re-plotting should now be much easier than before as I save Cohps.
I have also included more tests for the schema. In this way, I also found another error in the implementation. There was an error in the (refactored) detection of the StrongestBonds. I should have written better tests in the first place.


JaGeo:
I remove some files from the test data that were not needed. I hope this is fine to merge now. I will vanish into the weekend now.

utf:
Thanks @JaGeo!

JaGeo:
Thanks!

We will use it for data generation in the next weeks and again check it in detail. Unfortunately, it takes a bit of time to make sure the workflows, computational settings, and schemas are of very good quality and work for every use-case.

PR #274, opened 2023-03-20
Linting

utf:
Apply some stricter linting rules.


utf:
Thanks for the context!

PR #276, opened 2023-03-20
Phonon tweaks

### LINKED ISSUE #271: Extremely tight phonon k-point mesh
Hi @JaGeo.

I just tried running the phonon workflow and noticed that the k-point density used is extremely high. It's set as `grid_density=7000` which equals a reciprocal density of around `600` for silicon. Just for reference, the default reciprocal density in Atomate1 is 64 for insulators and 200 for metals. So this works out almost 10x the default k-point density.

For silicon, this means a supercell with lengths 22 Å is being run with a 2x2x2 k-point mesh!

I'd be very surprised if we need to go this high to get converged results. My feeling is that we should could use `reciprocal_density=100` and still get converged results. From my previous experience, I usually run displaced supercells with the same k-point density as the relaxation and this normally works well. My experience is that it is usually the force and energy convergence that are essential to get good phonons.

What do you think about:
1. Switching from `grid_density` to `reciprocal_density`, this means the number of k-points is independent of the number of atoms and only depends on the cell size.
2. Dropping the density to `reciprocal_density=100`

Obviously, we should do some tests to make sure this gives reasonable results still.

utf:
Optimise phonon workflow.

Main changes:
- Displacement k-point density dropped down.
- Use `auto_ispin`.
- Use `prev_vasp_dir` when possible to enable `auto_ispin`.

@JaGeo I'm currently running some tests to make sure the results are still reasonable.

Fixes #271.


JaGeo:
Great. We could also add a line in the documentation that k-point settings need to be carefully tested before using the workflow in high-throughput production as there is not yet a large-scale test.

JaGeo:
Is it fine to merge from your side, @utf ? I will work on the phonon workflow next week and it would be great if this gets merged before to not deal woth merge conflicts.

Issue #275, opened 2023-03-20
Improvement of preconvergence step lobster workflow


JaGeo:
@utf , there is still one other issue with the Lobster workflow. I have implemented this pre-convergence step of the WAVECAR. I have, however, the feeling that the speed-up is extremely small. I am not sure why this is the case. VASP still needs many electronic steps after starting from such a pre-converged WAVECAR. I might need to do some more tests to make this really efficient or test some larger structures.

utf:
I have a couple of thoughts having looked more closely at the lobster workflow:
1. I wonder if the need for many more steps is because NBANDS is only set after the preconvergence job? If you could do the pre-convergence with the same higher NBANDS that might make things faster since it would have converged all necessary bands?
2. Can you use a non-scf uniform input set for the `LobsterStaticMaker`? Obviously this won't work for hybrid calculations but that should allow you to go to very dense k-point meshes without much slowdown. If we did this, we should make the preconvergence step compulsory. In fact, an even better approach would be to replace both the `preconverge_static_maker` and `lobster_static_maker` with a single maker `lobster_uniform_maker` that is actually a flow maker. E.g., some variation of `UniformBandStructureMaker` but with symmetry disabled for the second job.
3. Can we switch to using `reciprocal_density` instead of `grid_density` for the `LobsterStaticMaker`?

JaGeo:
1. Yes, this is also my assumption. I have to test it in detail for some cases once our cluster is back up.
2. I haven't performed Lobster computations in this way. This would also be something we need to check.
3. Sure. I think it is much better

JaGeo:
Strategy 2 seems to work now:

51 s with smaller number of reciprocal points (reciprocal_density: 100) and without symmetry (50s), uniform band structure computation with wavecar writing (reciprocal_density: 400, 150 s)

400 s with only one static computation (400 s, reciproal_density: 400)

I will now try to confirm that the Lobster outputs also still look fine based on this.

JaGeo:
![image](https://user-images.githubusercontent.com/22094846/226878604-2fa1fae5-3b24-4a95-b505-90db376676c7.png)
It looks very similar. Just VASP sets the band gap differently in VASP 6 and in this new maker ISPIN is set to 1 in this case.

I also compared ICOHPLIST.lobster and again the agreement is very good.

utf:
Nice! That looks promising. 

[...]

PR #277, opened 2023-03-21
Improve Lobster workflow preconverge step, kpoints, docs

JaGeo:
## Summary

I have now implemented some of the suggestions from https://github.com/materialsproject/atomate2/issues/275 (same number of bands in preconverge static step and static step). 

I also changed the number of kpoints. Again, I think some more tests are necessary to confirm that this reciprocal density of kpoints is emough.

I added a tutorial on how to access the lobsterpy plots from the database.


JaGeo:
@utf, maybe, we could also check if auto_ispin is useful in this workflow as well.

JaGeo:
Still not happy with the preconverge step. Will try to work on it further.

utf:
Ok! I have a feeling using the NonSCF uniform maker will do the trick!

JaGeo:
Yes, you are probably right. I will check it out tomorrow!

JaGeo:
- [ ] Test update (LobsterStaticMaker and LobsterUniformMaker)
- [ ] check output schema for uuids and directories
- [ ] check deletion of WAVECARs

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -61,14 +61,14 @@ class LobsterStaticMaker(BaseVaspMaker):
     name: str = "static_run"
     input_set_generator: VaspInputGenerator = field(
         default_factory=lambda: StaticSetGenerator(
-            user_kpoints_settings={"grid_density": 6000},
+            user_kpoints_settings={"reciprocal_density": 400},
             user_incar_settings={
                 "IBRION": 2,
```
utf, line 66: And IBRION, ISIF, NSW from here.

src/atomate2/vasp/jobs/lobster.py
```diff
@@ -151,6 +159,8 @@ def update_user_incar_settings_maker(
         Structure object.
     prev_vasp_dir : Path or str
         Path or string to vasp files.
+    append_name: str or None
```
utf, line 162: I guess you don't need append_name here since that was for the preconvergence stuff

JaGeo: Yes, you are right. I need to clean up.

src/atomate2/vasp/jobs/lobster.py
JaGeo: Short question: should I enable auto_ispin and things like this here as well?

utf: Yeah, I think you can enable it. The default is for the lobster workflow to do a relaxation anyway, so I think it could be useful.

src/atomate2/vasp/flows/lobster.py
```diff
@@ -154,18 +140,97 @@ def make(
...
+            )
+        )
+    )
+    bs_maker: BaseVaspMaker = field(
+        default_factory=lambda: NonSCFMaker(
+            input_set_generator=NonSCFSetGenerator(
+                user_kpoints_settings={"reciprocal_density": 400},
+                user_incar_settings={
+                    "LWAVE": True,
+                    "NSW": 0,
```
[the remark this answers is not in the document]
JaGeo: Will ICHARG be set to 11 even if I give different incar settings?

src/atomate2/vasp/flows/lobster.py
```diff
@@ -154,18 +139,91 @@ def make(
...
+                user_kpoints_settings={"reciprocal_density": 400},
+                user_incar_settings={
+                    "LWAVE": True,
+                    "ISYM": 0,
+                },
+            )
+        )
+    )
+
+    def make(self, structure: Structure, prev_vasp_dir: str | Path | None = None):
```
JaGeo, line 205: Do you have a suggestion with regard to code reuse? I think I simply copied this part and just adapted the part above.

utf: You could just subclass the `UniformBandStructureMaker` and remove the make function entirely?

utf: Jimmy has done it this way in the phonon workflow and means we don't get too many makers. The line between adding a new maker vs reusing the old ones is quite ambitious though, so happy either way.

JaGeo: I like the constant in this case. I was thinking about subclassing but this would have added another layer of complexity as well.

JaGeo: (btw, I also like that we still get the VASP dos in this case. It also allows to compare Lobster and VASP dos if necessary. thus, I will keep the higher NEDOS setting)

JaGeo:
@utf to me it looks like I currently can only access the output from the uniform_job and nothing else in the UniformBandStructureMaker. Shouldn't we add the dir_name and uuid from the static_job before as well?

utf:
I think its ok that you only store the dir_name and uuid of the non-scf step. If you really need it you could use the uuid of the non-scf to get the inputs of that job and get the static dir.

If you really want to include them we could add a `all_outputs` property to the `Flow` class in jobflow but this could be quite brittle if your flow maker has an job which isn't a vasp job.

utf:
Great, thank you for these changes!

PR #280, opened 2023-03-27
[WIP] Defects CP2K

nwinner:
WIP for my defects module. 

The jobs/flows could be deleted if @jmmshn 's module moves into common. The builder for assimilating after the fact could be adjusted and added to common as well. To my knowledge there is no equivalent in VASP module. Its not totally code agnostic at the moment though.


jmmshn:
Hi @nwinner, the flow is already in common.  The parts that should be language agnostic have been left as abstract methods.  

I also added a tutorial on dealing with DFT-code agnostic workflows here:
https://github.com/materialsproject/jobflow/blob/main/docs/tutorials/4-generalized-makers.ipynb

Maybe you can just do a PR for driving the CP2K defect calculations only and we can figure out the schema later after a bit of discussion.  In short, I really think the workflow should only take you as far as the DefectEntry creation and Freysoldt correction.   The final chempot stuff should be left to builders.

esoteric-ephemera:
Hey @nwinner I see this is stale, but it would be nice to get defects stuff for CP2K into atomate2. Let me know if you're still interested in this or might need help getting it implemented, otherwise I'll close this out

PR #289, opened 2023-03-28
fix _get_strong_bonds function

naik-aakash:
## Strongest bonds for ICOBI and ICOBI were incorrect

* Fixes introduced in the associated `_get_strong_bonds` function in schema.


JaGeo:
Thanks @naik-aakash for fixing the issue.

PR #292, opened 2023-03-31
[Bug Fix] For stringing defect calculations together

jmmshn:
remove <hostname> while passing `dir_name` from bulk calculations


utf:
Thanks for this. Just fyi, I added a convenience function to do this: e.g., https://github.com/materialsproject/atomate2/blob/3b057159ef80a0305b2d10b9aaa959923cfdd494/src/atomate2/vasp/files.py#L56

PR #297, opened 2023-04-03
BUGFIX `auto_lreal`

jmmshn:
- Current implementation returns None in some cases.  This can result in `LREAL=None` in the INCAR
- Also added a new keyword for `copy_vasp_outputs` to allow `force` kwarg to be activated in `copy_files`.


utf:
Thanks!

Issue #300, opened 2023-04-08
Molecule schema should probably be merged within emmet


arosen93:
Currently, the `MoleculeMetadata` class is in Atomate2 [here](https://github.com/materialsproject/atomate2/blob/main/src/atomate2/common/schemas/molecule.py). Longer term, it probably makes sense to either port it to emmet where it can co-exist with [StructureMetadata](https://materialsproject.github.io/emmet/core/#emmet.core.structure.StructureMetadata) or, potentially better yet, adopt/merge with whatever @espottesmith is planning to do (or has already done) in terms of making a `MoleculeMetadata` for MP. For now, what we have works though.

espottesmith:
There is a MoleculeMetadata class now in emmet main. Take a look and see if it fits your needs.

arosen93:
Ah thanks! I must have missed it when looking before.

PR #301, opened 2023-04-08
Switch to emmet's `MoleculeMetadata`

arosen93:
Addresses #300, wherein I realized we had a `MoleculeMetadata` schema in both Atomate2 and emmet, resulting in near-duplicate blocks of code. I merged the two together in emmet [here](https://github.com/materialsproject/emmet/pull/691) and updated Atomate2 to use emmet's `MoleculeMetadata` instead just like you did for `StructureData`. This should consolidate things within Atomate2 nicely. It's also coincidentally in line with Alex's recent efforts to move some more general schema things to emmet.

Edit: We get about 250 lines out of Atomate2 and into emmet (where they belong), which I consider a win.


arosen93:
@utf: This should be ready to merge. One quick comment/question though to you and/or @munrojm. In [`test_cclib.py`](https://github.com/arosen93/atomate2/blob/molecule/tests/common/schemas/test_cclib.py#L95), I had to change `jsanitize(doc, strict=True, enum_values=True)` to `jsanitize(doc, enum_values=True)` because the `strict=True` flag was causing an issue with jsanitizing a `DateTime` object, specifically `datetime.datetime(2023, 4, 8, 23, 40, 26, 928134)` that represents the `build_date` of the emmet `BaseModel` [here](https://github.com/materialsproject/emmet/blob/920edf87cfdfc5b9601e7b6679389a4584f7736d/emmet-core/emmet/core/base.py#L30). For what I'm talking about specifically, see [this failed test](https://github.com/materialsproject/atomate2/actions/runs/4647643801/jobs/8224750401#step:6:373). I assume that FireWorks is going to convert the `DateTime` to a string so this should be okay, but if not let me know. I'm not really sure why this is coming up now, unless `build_date` was recently added to emmet.

arosen93:
@utf, could you give this a look when you get the chance? Thanks!

utf:
This looks great. Thanks Andrew

arosen93:
Taken care of now, boss. Feel free to merge as your discretion!

utf:
Thanks!

Issue #302, opened 2023-04-09
Suggestion: Move cclib `TaskDocument` to emmet


arosen93:
Motivated by the move of the VASP `TaskDocument` to emmet, I think the same should be done with cclib.

PR #303, opened 2023-04-09
This PR seeks to move the `cclib`-based `TaskDocument` to emmet

arosen93:
Based on https://github.com/materialsproject/atomate2/issues/302.


arosen93:
I'm closing this issue. After talking with @phuck, it seems like this should not make its way into emmet until/if it becomes part of MP.

PR #320, opened 2023-05-02
added simple chg check

jmmshn:
## Basic charge states check

Make sure that the charge state matches the parsed output charge.
Mistakes can happen due to upstream pymatgen problems that were only fixed recently.


utf:
Great, thanks!

PR #322, opened 2023-05-03
Add forcefield schemas/makers to atomate2

matthewkuner:
## Summary

* Add the ability to use forcefields (e.g. CHGNet) to relax structures in Atomate2.

* chgnet, which depends on pytorch (relatively large)

## TODO 

* Merge with main

## Comments
Based on what Shyue Ping told me [here](https://github.com/materialsvirtuallab/matgl/issues/44#issuecomment-1533952719), I think we should wait until M3GNet is fully moved to the [matgl repository](https://github.com/materialsvirtuallab/matgl), because that will the location for future development of M3GNet. This will also make it so that users will only need pytorch for both CHGNet and M3GNet (as opposed to needing TensorFlow too, as TensorFlow powers the older M3GNet implementation).


matthewkuner:
I tested this as a standalone job and as a two-part job with the CHGNet prerelax feeding into a VASP job. Both worked.

matthewkuner:
People may want to use CHGNet makers as an input in the VASP flows that currently require inputted variables to be of type BaseVaspMaker. Would it be reasonable to change this to allow CHGNetRelaxMakers as well? (or perhaps some base FFRelaxMaker class that I could make)

matthewkuner:
I forgot to add the new dependency for chgnet--unsure where in the `pyproject.toml` that should be done / how the version restrictions should work. @utf any help?

src/atomate2/forcefields/jobs.py
```diff
@@ -0,0 +1,81 @@
...
+        Keyword arguments that will get passed to :obj:`StructOptimizer()`.
+    keep_info : list
+        Which information from the relaxation trajectory to save using
+        the :obj:`.FFStructureRelaxDocument.from_chgnet_result`.
+
+    """
+
+    name: str = "CHGNet relax"
+    relax_cell: bool = False
+    relax_kwargs: dict = field(default_factory=dict)
```
utf, line 39: What do you think about removing the relax_cell argument and just putting it in the default relax_kwargs. E.g.
```suggestion
    relax_kwargs: dict = field(default_factory={"relax_kwargs": False})
```

src/atomate2/forcefields/jobs.py
```diff
@@ -0,0 +1,81 @@
...
+    optimizer_kwargs: dict = field(default_factory=dict)
+    keep_info: list = field(
```
utf, line 41: Remove this and put the default value on the task document. Add a new field `task_document_kwargs` that will get passed to the task document function. If people want to override this field then they can do something like:

```python
maker = CHGNetRelaxMaker(task_document_kwargs={"keep_info": ["energies"]})
```

This way is better since if we add more kwargs to the task document later on then we don't need to add new attributes to the Maker class.

Note, when you add the default to the task document function defaults, specify them as a tuple rather than a list.

src/atomate2/forcefields/jobs.py
```diff
@@ -0,0 +1,81 @@
...
+            "energies",
+            "forces",
+            "stresses",
+            "magmoms",
+            "atom_positions",
+            "cells",
+        ]
+    )
+    # NOTE: the 'atoms' field will always be removed later because it is not
+    # serializable (as of May 2023)
```
utf, line 52: Remove, will discuss this later.

src/atomate2/forcefields/jobs.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @job(output_schema=FFStructureRelaxDocument)
+    def make(self, structure: Structure):
+        """
+        Perform a relaxation of a structure using CHGNet.
+
+        Parameters
+        ----------
+        structure: ~pymatgen.core.structure.Structure
```
utf, line 61:
```suggestion
        structure : .Structure
```

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
+"""Job to prerelax a structure using an MD Potential."""
+
+import logging
+from typing import List
+
+from pydantic import BaseModel, Field
+
+# from jobflow import Maker, job
```
utf, line 8: Remove

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class InputDoc(BaseModel):
+    """The inputs used to run this job."""
+
+    structure: Structure = Field(None, description="The inputted structure.")
+    relax_cell: bool = Field(
+        None,
+        description="Whether cell lattice was allowed to change during relaxation.",
+    )
```
utf, line 19: Can remove.

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+    trajectory: dict = Field(
+        None, description="Step-by-step trajectory of the structural relaxation."
+    )
+
+    steps: int = Field(
+        None, description="total number of steps needed to relax the structure."
+    )
+
+
+class FFStructureRelaxDocument(BaseModel):
```
utf, line 62: Can you make this subclass StructureMetaData like the VASP task document does. This will add a load of useful metadata (like formula, volume, composition, etc) to the document automatically.

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+
+    input: InputDoc = Field(
+        None, description="The inputted information used to run this job."
+    )
+
+    output: OutputDoc = Field(
+        None, description="The outputted information from this relaxation job."
+    )
+
+    MD_potential: str = Field(
```
utf, line 77: Lower case attributes!
```suggestion
    md_potential: str = Field(
```

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+        None,
+        description="name of the interatomic potential used for relaxation.",
+    )
+
+    MD_potential_version: str = Field(
```
utf, line 82:
```suggestion
    md_potential_version: str = Field(
```

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @classmethod
+    def from_chgnet_result(
+        cls,
+        input_structure: Structure,
+        relax_cell: bool,
+        relax_kwargs: dict,
+        optimizer_kwargs: dict,
+        result: dict,
+        keep_info: list,
```
utf, line 95: Re-order and provide a default value. Also, do you actually need the input structure, won't this just be the first structure in the trajectory?

Lastly, based on my comment above about changing `trajectory` -> `ionic_steps`, the `keep_info` could be modified to remove the plurals in some cases (because we now have a list of dicts). Also `keep_info` could be renamed to `ionic_step_data`.

```suggestion
        result: dict,
        relax_kwargs: dict = None,
        optimizer_kwargs: dict = None,
        keep_info: tuple = ("energy", "forces", "stress", "magmoms", "structure"),
```

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # NOTE: units for stresses were converted to kbar
+    # (* -10 from standard output) to comply with MP convention
+    stress: List[float] = Field(
+        None, description="The stress on the cell in units of kbar (in Voigt notation)."
+    )
+
+    trajectory: dict = Field(
+        None, description="Step-by-step trajectory of the structural relaxation."
+    )
```
utf, line 55: Can we try and emulate the `ionic_steps` document in the VASP Task document. See https://github.com/materialsproject/emmet/blob/2289affb31d9b18aa82e49484ca772a2ce80e934/emmet-core/emmet/core/vasp/calculation.py#LL305C5-L305C114

and 

https://github.com/materialsproject/emmet/blob/2289affb31d9b18aa82e49484ca772a2ce80e934/emmet-core/emmet/core/vasp/calculation.py#L262


So basically, it is a list of dicts rather than multiple dicts of lists. Rather than the funky `e_fr_energy` etc in the IonicStep linked, you can just have `energy`. You can also remove `electronic_steps` and add `magmoms`.

This will involve reorganising the output of CHGNet somewhat but will bring the outputs into line with VASP, CP2K, etc.

Note that we should also store the structure at each step. You can convert an ase atoms object using:
```python
from pymatgen.io.ase import AseAtomsAdaptor
structure = AseAtomsAdaptor.get_structure(atoms)
```

src/atomate2/forcefields/schemas.py
```diff
@@ -0,0 +1,167 @@
...
+        trajectory_to_save = {key: trajectory[key] for key in keep_info}
+        if "atoms" in keep_info:
+            warning_msg = """
+                WARNING: `Atoms` objects can't be serialized (as of May 2023) and
+                thus are automatically removed.
+                """
+            logging.warning(warning_msg)
+            trajectory_to_save.pop(
+                "atoms"
+            )  # can't serialize `AseAtoms` objects, so remove them from trajectory.
```
utf, line 143: See previous comments about converting atoms objects to structures.

utf (commented):
Hi @matthewkuner, thanks very much for this! It is in a great state for an initial PR.

I've added a bunch of comments. They are mainly style changes, but I had some more thoughts about the task document (I've been thinking about this more since we spoke on slack).

I think it would be nice to add:
1. A `CHGNetStaticMaker` that will just calculate forces but not do a relaxation. After you've made the changes I suggested this should be very quick to implement by subclassing `CHGNetRelaxMaker`. Just something like:
```python
@dataclass
class CHGNetStaticMaker(CHGNetRelaxMaker):
    """
   Maker to calculate forces and stresses using the CHGNet force field.

    Parameters
    ----------
    name: str
        The job name.
    relax_kwargs : dict
        Keyword arguments that will get passed to :obj:`StructOptimizer.relax`.
    optimizer_kwargs : dict
        Keyword arguments that will get passed to :obj:`StructOptimizer()`.
    task_document_kwargs : dict
        Keyword arguments that will get passed to :obj:`ForceFieldTaskDocument()`.
    """

    name: str = "CHGNet static"
    optimizer_kwargs: dict = field(default_factory={"relax_cell": False, "steps": 1})
    task_document_kwargs: dict = field(default_factory=dict)
```
2. Can you add tests. CHGNet should be fast enough to run in the testing environment. Just use silicon and limit steps to 10

As for where to add the dependency. You should add it to the `strict` set of dependencies with a specified version number (the latest one). You can also add a new set of dependencies called `chgnet` which isn't pinned to a specific version. E.g., see how this is done for the `lobster` set: https://github.com/materialsproject/atomate2/blob/43216201d2324c7f197db92653f3f0e61146a991/pyproject.toml#L44

Thanks again, and please get in touch if you have any questions.

src/atomate2/forcefields/schemas.py
matthewkuner: RE: keeping the input structure--I was just including it to be in-line with the VASP TaskDoc (in case users wanted an easy-to-reach initial structure). It is indeed a duplicate, but at the same time so is having an `output.structure` and `output.output.structure` field in the VASP TaskDoc. Let me know your thoughts--I don't feel too strongly one way or the other.

utf: Sorry for the confusion, the structure should definitely be in the task document. But you don't need to pass it separately to this function. It can be found in the result already. So just extract it from there rather than requiring it to be passed manually.

src/atomate2/forcefields/jobs.py
matthewkuner: I'm slightly on the fence about this, solely because I think it is the most important parameter that a user can set (hence I was drawing special attention to it by having it as a separate variable). Let me know your thoughts.

src/atomate2/forcefields/flows.py
```diff
@@ -0,0 +1,55 @@
...
+from jobflow import Flow, Maker
+from pymatgen.core.structure import Structure
+
+from atomate2.forcefields.jobs import CHGNetRelaxMaker
+from atomate2.vasp.jobs.base import BaseVaspMaker
+from atomate2.vasp.jobs.core import RelaxMaker
+
+
+@dataclass
+class CHGNetToVaspMaker(Maker):
```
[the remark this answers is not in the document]
matthewkuner: I'm not convinced that a random user would identify that this is a CHGNet --> VASP flow based on the new name alone. Hmm..

src/atomate2/forcefields/schemas.py
matthewkuner: Ah, okay that makes sense

src/atomate2/forcefields/schemas.py
matthewkuner: so you'd want one of the output fields to be `ionic_steps`, which is a list. That list will contain as many dicts as are specifies, which would include things like `energy`, `structure`, etc.? Sure I can do that.

Separately--considering that the linked VASP Task Document doesn't even let users specify which parts of the trajectory to keep, I don't think we should either. I'll just replace the tuple/list `keep_info` with a bool `keep_trajectory`. Let me know if you have any objections.

matthewkuner: Actually, it appears that users don't even have the option to _not_ store the trajectory for MD jobs. Seems pointless to exclude it for FF relaxations in that case. I'll just remove the tag entirely.

matthewkuner: RE: converting the `Atoms` objects to structures. It appears that CHGNet only outputs a single `Atoms` object, rather than one for each step... very odd. Any advice?

src/atomate2/forcefields/flows.py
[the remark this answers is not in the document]
utf: Maybe: `CHGNetRelaxVaspMaker` This is inline with other combined flows which don't include "To" when linking two calculations together. E.g. `RelaxBandStructureMaker`

src/atomate2/forcefields/jobs.py
utf: Having thought about this more, I think we should definitely remove it. For the VASP makers, we don't have flags which control the input set parameters directly on the maker itself. All settings for controlling the actual calculation should be in optimizer_kwargs. To make things easier, I'd suggest we explicitly reproduce some of the defaults so that it is clear what calculation is happening and in case the chgnet defaults change.

E.g.,

```python
relax_kwargs: dict = field(default_factory={"relax_kwargs": False, "steps": 500, "fmax": 0.1})
```

And then we can outline what these specific parameters mean in the docstring and point the user to the chgnet documentation for more details.

src/atomate2/forcefields/schemas.py
utf: I think having the keep_info tag is actually useful and just because its not there in VASP doesn't mean we can't add it here. So I would keep that in and forget about `keep_trajectory` for now. However, one we add in MD through force fields, we will want to add trajectory support (a latter PR).

You can generate a new structure from the positions and cells along the lines of:
```python
species = atoms.get_chemical_symbols()
for pos, cell in zip(atom_positions, cells):
    structure = Structure(cell, species, positions, coords_are_cartesian=False)
```

Double check whether the `atom_positions` are in cartesian coordinates and update `coords_are_cartesian` accordingly.

src/atomate2/forcefields/flows.py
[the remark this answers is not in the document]
matthewkuner: I am convinced

src/atomate2/forcefields/jobs.py
utf: Actually, I think I changed my mind about this. The CHGNet relaxation is much simpler than the VASP calculations, so perhaps its ok to have a couple of parameters specified in the Maker. But I think you should also add a steps attribute too.

src/atomate2/forcefields/schemas.py
matthewkuner: Ah, I didn't know about the `coords_are_cartesian` input, and missed it when looking through the documentation. That fixed it.

src/atomate2/forcefields/jobs.py
utf: Yep, exactly.

matthewkuner: Wait which one? I edited my original comment

mkhorton:
Adding a brief note that pymatgen has started offering an API that may make the code a bit simpler:

https://pymatgen.org/pymatgen.core.structure.html#pymatgen.core.structure.Structure.relax

I believe there’s an open issue to add chgnet as a relaxer too, I think @janosh is involved.

janosh:
> I believe there’s an open issue to add chgnet as a relaxer too, I think @janosh is involved.

Yes, I think CHGnet would make a great addition to `Structure.relax()` and should be straightforward to add. Last I checked with @BowenD-UCB he wanted to wait a bit to make sure the CHGNet API has fully solidified. I think that might be the case now?

src/atomate2/forcefields/schemas.py
matthewkuner: Currently, when I try to make this a subclass of StructureMetadata, I get the error 

> Traceback (most recent call last):
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/jobflow/core/job.py", line 587, in run
    output = jsanitize(response.output, strict=True, enum_values=True)
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 559, in jsanitize
    return jsanitize(
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 526, in jsanitize
    return {
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 527, in <dictcomp>
    str(k): jsanitize(
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 526, in jsanitize
    return {
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 527, in <dictcomp>
    str(k): jsanitize(
  File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 568, in jsanitize
    obj.as_dict(),
AttributeError: 'datetime.datetime' object has no attribute 'as_dict'

> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/jobflow/managers/local.py", line 98, in _run_job
>     response = job.run(store=store)
>   File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/jobflow/core/job.py", line 589, in run
>     raise RuntimeError(
> RuntimeError: Job output contained an object that is not MSONable and therefore could not be serialized.

Any ideas?

janosh: The serialization handler needs another check:
```py
if isinstance(obj, datetime. datetime):
   return str(obj)
```

Maybe check if atomate2 uses `str` or some other serialization function.

matthewkuner:
@mkhorton would implementing CHGNet into pymatgen's structure objects make this Flow redundant? Or are they distinct use-cases?

utf:
I don't think that function makes anything redundant, I think you we could just it inside your job rather than constructing the relaxer manually (once chgnet has been implemented)

janosh:
> Or are they distinct use-cases?

@matthewkuner Both are useful! `Structure.relax` for one-off analysis and this `maker` for when you want to integrate CHGNet into a HT wf. But `Structure.relax` could slightly simplify this `maker`'s implementation. No need to wait though. Can be a non-breaking change later.

utf:
@matthewkuner missed a "use" from my message. But effectively your job would simplify from

```python
     @job(output_schema=FFStructureRelaxDocument)
     def make(self, structure: Structure):
         """
         Perform a relaxation of a structure using CHGNet.

         Parameters
         ----------
         structure : .Structure
             A pymatgen structure.

         """
         from chgnet.model import StructOptimizer

         relaxer = StructOptimizer(**self.optimizer_kwargs)
         result = relaxer.relax(
             structure, relax_cell=self.relax_cell, **self.relax_kwargs
         )

         ff_structure_relax_doc = FFStructureRelaxDocument.from_chgnet_result(
             structure,
             self.relax_cell,
             self.relax_kwargs,
             self.optimizer_kwargs,
             result,
             self.keep_info,
         )

         return ff_structure_relax_doc
```

To 
```python
     @job(output_schema=FFStructureRelaxDocument)
     def make(self, structure: Structure):
         """
         Perform a relaxation of a structure using CHGNet.

         Parameters
         ----------
         structure : .Structure
             A pymatgen structure.

         """
         result = structure.relax(
             structure, relax_cell=self.relax_cell, **self.relax_kwargs
         )

         ff_structure_relax_doc = FFStructureRelaxDocument.from_chgnet_result(
             structure,
             self.relax_cell,
             self.relax_kwargs,
             self.optimizer_kwargs,
             result,
             self.keep_info,
         )

         return ff_structure_relax_doc
```

matthewkuner:
ahh, I understand now @utf. I'll keep an eye out for when it's implemented in pymatgen

matthewkuner:
@utf RE: adding chgnet dependencies (https://github.com/materialsproject/atomate2/pull/322#pullrequestreview-1413905414) , chgnet is not yet on PyPI. Does that matter?

JaGeo:
Just wanted to mention, we could think about connecting other workflows as well (.e.g, phonons). The workflow is written so general that depending on the output schema, such a workflow could be straightforward to implement as well.

matthewkuner:
@JaGeo I was already thinking about modifying the elastic constant workflow to allow a CHGNet --> VASP flow maker to be used as well. Seems like it could be implemented for a ton of workflows. Perhaps that would mean we should make a BaseForceFieldRelaxMaker class, though...

utf:
@JaGeo I also thought about that too. I had a look and it seems like the main issue is that jobs are connected using both `structure` and `prev_vasp_dir`. One workaround would be to add a *kwargs field to the make function of CHGNetMaker but this seems quite hacky.


Going forward, we will need to think of a more formal way of automatically connecting two jobs from different DFT codes. Basically providing an API.

utf:
@matthewkuner For now, we can just use the GitHub version (it should still be added to both strict and a new set of dependencies). E.g., see https://chrisholdgraf.com/blog/2022/install-github-from-pyproject/

JaGeo:
@matthewkuner yes, a more general force field schema might be interesting as well! The schema could probably be reused as well. Most of them rely on ase as well!

@utf yes, it would be awesome to think about this. Now, that more and more codes are coming, it would be great to combine their capabilities in an easier way!

src/atomate2/forcefields/schemas.py
```diff
@@ -131,27 +152,57 @@ def from_chgnet_result(
...
-            trajectory_to_save.pop(
-                "atoms"
-            )  # can't serialize `AseAtoms` objects, so remove them from trajectory.
-
-        steps = len(result["trajectory"])
+        n_steps = len(trajectory["energies"])
+
+        ionic_steps = []
+
+        if "keep_info" not in task_document_kwargs:
```
matthewkuner, line 159: @utf is this the proper way to use the `task_document_kwargs`? I feel it is not...

src/atomate2/forcefields/schemas.py
matthewkuner: why do you think the `relax_cell` parameter should be removed from the InputDoc?

janosh:
> @utf RE: adding chgnet dependencies ([#322 (review)](https://github.com/materialsproject/atomate2/pull/322#pullrequestreview-1413905414)) , chgnet is not yet on PyPI. Does that matter?

Fyi https://github.com/CederGroupHub/chgnet/pull/29

matthewkuner:
@janosh great timing!

src/atomate2/forcefields/schemas.py
utf: No this isn't quite right. There shouldn't be a task_document_kwargs argument to this function. Add `keep_info` as an argument. When you call `ForceFieldTaskDocument.from_chgcar_result` you can do:

```python
ForceFieldTaskDocument.from_chgcar_result(structure, ..., **self.task_document_kwargs)
```

The `**` will expand task_document_kwargs (which is a dictionary). So for example, if `task_document_kwargs={"keep_info": ("energy")}`, then the above code would be equivalent to 
```python
ForceFieldTaskDocument.from_chgcar_result(structure, ..., keep_info=('energy'))
```
But python will do all the unpacking for you.

matthewkuner: yep, that makes a lot more sense.

matthewkuner:
@utf in one of your previous comments, you mentioned subclassing `ForceFieldTaskDocument` from `StructureMetadata` instead of `BaseModel`. However, when I tried this, I get the following error:

> File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/monty/json.py", line 568, in jsanitize
>     obj.as_dict(),
> AttributeError: 'datetime.datetime' object has no attribute 'as_dict'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/jobflow/managers/local.py", line 98, in _run_job
>     response = job.run(store=store)
>   File "/Users/kuner/opt/anaconda3/envs/atomate2/lib/python3.9/site-packages/jobflow/core/job.py", line 589, in run
>     raise RuntimeError(
> RuntimeError: Job output contained an object that is not MSONable and therefore could not be serialized.

Any ideas?

utf:
@matthewkuner are you using the latest jobflow version? I should have fixed this in v0.1.11

matthewkuner:
@utf I was using v.0.1.9. I'll check it again with the latest version of jobflow soon.

matthewkuner:
@utf I think this is getting very close to complete. Let me know if you have any other things you'd like me to add/change.

arosen93:
This is very cool! Looking forward to using this in my own work! @utf, could we get a new version release after this is merged if you don't mind?

janosh:
I just read through some of the changes here as part of #362 and want to say this is a really excellent contribution! Thanks for the work you put into this @matthewkuner! 🙏

rkingsbury:
Looking great to me so far! Are you planning to replicate the unit tests we have in `pymatgen` / `atomate` for the legacy workflow?

rkingsbury:
> @rkingsbury I wrote my own few tests (as I didn't know pymatgen had a m3gnet method /tests until I had written my own tests), so I wasn't planning to. Is it important?

Ooops, sorry @matthewkuner , I posted that on the wrong PR! Please disregard. (Although I am excited to see this, too)

PR #344, opened 2023-05-19
More `ruff`

janosh:
Mostly unexciting changes. I left on TODO: there are currently 3 tests that use `pytest.raises()` to that only check the error type. These are flagged by `ruff` as too broad (recommendation is to also check the error message). [I ignored these lints](https://github.com/materialsproject/atomate2/compare/more-ruff?expand=1#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R169) but would be better to add `match="expected err msg"`.


utf:
Thanks!

PR #355, opened 2023-05-23
allow elastically unstable structures

matthewkuner:
## Summary

Introduce a `task_document_kwarg` to the `ElasticDocument` that allows the elastic workflow to still complete in the event that the structure is elastically unstable. This specifically affects the `DerivedProperties` field in the ElasticDocument, which will be set to none in the event that the structure is elastically unstable.

@utf I made it so that elastically unstable structures are, by default, **_allowed_** to complete. Let me know if you think this should be changed.

### DIFF
--- a/src/atomate2/common/schemas/elastic.py
+++ b/src/atomate2/common/schemas/elastic.py
@@ -148,6 +148,7 @@ class ElasticDocument(BaseModel):
         order: Optional[int] = None,
         equilibrium_stress: Optional[Matrix3D] = None,
         symprec: float = SETTINGS.SYMPREC,
+        allow_elastically_unstable_materials: bool = True
     ):
         """
         Create an elastic document from strains and stresses.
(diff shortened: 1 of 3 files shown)


matthewkuner:
@utf I made it so that elastically unstable structures are, by default, _**allowed**_ to complete. Let me know if you think this should be changed.

mkhorton:
Can you think of a way of adding logic here that might distinguish between truly dynamically unstable materials and materials for which the fit is just poor (e.g., has gone into a non-linear regime)? It seems important to distinguish these possibilities.

Separately, how is "elastic instability" determined -- negative eigenvalues of Voigt tensor, or some other method?

matthewkuner:
@mkhorton I was just looking to allow workflows to complete when the bulk/shear modulus are negative (which indicates elastic instability). Honestly, I'm not familiar enough with the fine details of elastic constant fitting methods to write logic for the fitting method not being applicable / identifying dynamic instability

mkhorton:
That makes sense, perfectly valid!

matthewkuner:
@mkhorton There is someone in Mark Asta's group who is more familiar with the math underlying elastic constant fitting/calculations. I will see if he has any ideas about automatically determining if the fitting method is not applicable. That probably belongs in another PR though

matthewkuner:
@utf let me know if this is good to go or if you'd like any changes made

utf:
Thanks @matthewkuner. Thanks for this.

PR #360, opened 2023-05-29
Fix `emmet-core` version in `setup.py`

arosen93:
Someone (likely me) accidentally set `emmet-core>=0.51.11` instead of `>=0.51.1`. Now fixed.


arosen93:
Never mind........

PR #362, opened 2023-06-02
[WIP] MP-compatible r2SCAN workflow (and a few general INCAR improvements)

### LINKED ISSUE #352: Port SCAN workflows to atomate2
MP currently stands at 33k r2SCAN calculations out of 154k materials. Eventually, we'd like to have 100% r2SCAN coverage.

The [current workflows are implemented in `atomate1`](https://github.com/hackingmaterials/atomate/blob/387938da7c578481c41aa4cfeb65a166778e6372/atomate/vasp/workflows/presets/scan.py#L9-L48C14). This issue is to track porting them to `atomate2`.

@utf Would be great to hear your thoughts (if any). @arosen93 and I were thinking of hacking something together to open a draft PR and then iterate with your feedback.

### LINKED ISSUE #363: BUG: `LMAXTAU` should be auto-set when there are f electrons and `LASPH` is True for meta-GGA calculations
As noted in https://github.com/materialsproject/pymatgen/issues/2325, `LMAXTAU` should be set to 8 if there are f electrons and `LASPH` is True in meta-GGA runs. This should be added to `vasp/sets/base.py` like the `LMAXMIX` swaps.

### LINKED ISSUE #364: BUG: `LMAXMIX` should be set based on the presence of s, p, d, f electrons not based on Z
Currently, `LMAXMIX` is set as follows:

https://github.com/materialsproject/atomate2/blob/c2d098a38a41c181be2c483f8adb83db0223b76d/src/atomate2/vasp/sets/base.py#L915-L927

However, this is based on Z value and not based on the presence or lack of s, p, d, or f electrons (see [VASP manual](https://www.vasp.at/wiki/index.php/LMAXMIX)). For instance, Sr has Z > 20 but does not have d electrons. We should be checking this based on the "block" of the electrons in the `Structure` object.

Andrew-S-Rosen:
Closes #352. Closes #363. Closes #364.

- #352
- #363
- #364

## R2SCAN Workflow

Note: This is in close collaboration with @janosh and @esoteric-ephemera.

This PR adds MP-compatible r2SCAN workflows to Atomate2 (#352). We are still adding tests for now, hence the WIP.

## VASP INCAR Improvements

I also baked in fixes for several VASP INCAR issues I brought up (#363, #364, #365).

## Review Request

@utf: Mind giving this a quick look to make sure we didn't do anything drastically wrong before we go and write all the tests? Thanks!

Also tagging @munrojm, @rkingsbury, and @mkhorton in case they're interested in following this.

## ToDo

- Finish adding unit tests
- Only copy the `CHGCAR` (not all the files) from the PBEsol job to the R2SCAN job in order to remain truthful to the original workflow.


munrojm:
This is great! Looking forward to having access to it.

janosh:
Linking https://github.com/materialsproject/pymatgen/issues/2325 as a TODO as suggested by @arosen93 so we don't forget about it.

arosen93:
Note to self while typing on mobile: this might be a swap to add in base.py of vasp/sets.

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,158 @@
...
+    input_set_generator: VaspInputGenerator = field(
+        default_factory=MPRelaxR2SCANGenerator
+    )
+
+    def make(self, structure: Structure, bandgap: float = 0.0) -> MPRelaxMaker:
+        """Set correct k-point density, smearing and sigma based on bandgap estimate."""
+        if bandgap < 1e-4:
+            kspacing = 0.22
+            ismear = 2
+            sigma = 0.2
```
rkingsbury, line 112: Should the bandgap tolerance be adjustable via kwarg?

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,158 @@
...
+    write_additional_data : dict
+        Additional data to write to the current directory. Given as a dict of
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    name: str = "MP PreRelax"
+    input_set_generator: VaspInputGenerator = field(
```
rkingsbury, line 64: Should the name be more descriptive, e.g. "MP PBESol PreRelax"?

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,65 @@
...
+    name : str
+        Name of the flows produced by this maker.
+    pre_relax_maker : .BaseVaspMaker
+        Maker to generate the first relaxation.
+    relax_maker : .BaseVaspMaker
+        Maker to generate the second relaxation.
+    """
+
+    name: str = "MP 2023 Relax"
+    pre_relax_maker: BaseVaspMaker = field(default_factory=MPPreRelaxMaker)
```
rkingsbury, line 42: Should the name be more descriptive e.g. "MP metaGGA Relax" or "MP r2SCAN Relax"?

I'm also not sure how I feel about the 2023 in the name. Since the papers came out in 2022, is there possibility for that to cause confusion?

src/atomate2/vasp/sets/BaseMPR2SCANRelaxSet.yaml
```diff
@@ -0,0 +1,160 @@
...
+    La3+: 0.6
+    Lu3+: 0.6
+    Mn: 5
+    Mn3+: 4
+    Mn4+: 3
+    Mo: 5
+    Nd3+: 3
+    Ni: 5
+    Pm3+: 4
+    Pr3+: 2
```
rkingsbury, line 53: I note that `MPSCANRelaxSet.yaml` in pymatgen does not include any MAGMOMs. I think it's good to have them here; just flagging as a potential source of discrepancies in test cases, etc.

rkingsbury (commented):
This is fantastic, thanks so much @arosen93  and @janosh! I have a postdoc starting in about a month who will likely want to make use of this, too. I left a few minor comments but overall it looks good to me (with the caveat that I am still not terribly familiar with `atomate2` VASP flows in general).

One enhancement to consider - in Figure 3 of the [mixing scheme paper](https://www.nature.com/articles/s41524-022-00881-w) (below), we describe different levels of calculation coverage, and note that sometimes (for large structures) it is useful / necessary to just do a static r2SCAN calculation. This gives you an energy that allows you to apply the mixing scheme even though the structure itself is probably too large to relax with r2SCAN.

image

If MP wants to adopt this strategy, I suggest breaking the 2nd relaxation into two `Job` - a static calculation followed by a relaxation. So the overall workflow would be:

PBESol relax --- CHGCAR----> r2SCAN static ----CHGCAR+WAVECAR----> r2SCAN relax

You could pass everything (`CHGCAR` and `WAVECAR` ) on to the final relaxation so it shouldn't cost much additional time, but this would give you the flexibility to stop after the static calc if you have large structures.

Just food for thought. The argument against this is that it will deviate slightly from what is described in the workflow paper, creating potential for confusion.

src/atomate2/vasp/jobs/mp.py
janosh: Good idea! I'll add `bandgap_tol: float = 1e-4`.

src/atomate2/vasp/jobs/mp.py
janosh: I'm on the fence about this. People might forget to customize the name if they change to, say, ML pre-relaxation in which case "MP PBESol PreRelax" would be misleading. I guess the job can try to be smart about it and set the name based on the `InputGenerator`. @utf What's `atomate2`'s general policy about this?

Speaking of which, since `atomate2` has 1st class ML pre-relax support now from @matthewkuner's excellent #322, is the type hint `VaspInputGenerator` on the pre-relax job sufficiently generic? Doesn't look like `class CHGNetRelaxMaker(Maker)` is compatible with `VaspInputGenerator` .

src/atomate2/vasp/sets/BaseMPR2SCANRelaxSet.yaml
janosh: Yes, this was a deliberate choice. They are identical with the MAGMOMs in [BaseMPGGASet.yaml](https://github.com/materialsproject/atomate2/blob/c98cf7c8744dbe001c1ffedf74430242db3015d9/src/atomate2/vasp/sets/BaseMPGGASet.yaml). Btw, we thought about extracting them into a separate `MAGMOMs.yaml` file and allowing for parent-child inheritance similar to `pymatgen`'s YAML sets but felt a bit clunky and over-engineered. Happy to discuss more though.

src/atomate2/vasp/jobs/mp.py
arosen93: I agree with @janosh on this. We don't want to tie the name to a given setting to allow for users to override things if they want to inherit the MP workflow but not use it as-is.

src/atomate2/vasp/flows/mp.py
arosen93: See comment below about not tying the name to a given setting, but I agree that 2023 is confusing. We can switch to 2022, although I don't know that the year is ideal. This was mostly inspired by the compatibility scheme naming. 

Any alternate suggestions given this info?

src/atomate2/vasp/sets/BaseMPR2SCANRelaxSet.yaml
arosen93: Note that we also added a default KSPACING, SIGMA, and ISMEAR to the input set that way if someone tried to use the MP settings on some random Atomate2 VASP workflow, they aren't running without any k-points. It also made writing the PBEsol step easier.

arosen93:
@rkingsbury that sounds like a great idea. I'll implement that, and we can have it be an optional kwarg that the user can decide to set or not set. Then users have it both ways.

Also, in case it's useful for you, the issue you previously ran into with WAVECARs causing a crash when going from PBEsol to R2SCAN on occasion shouldn't be a problem anymore. I added a patch to Custodian.

src/atomate2/vasp/jobs/mp.py
matthewkuner: @janosh do you think we should try to make CHGNetRelaxMaker compatible with the VaspInputGenerator? I wasn’t sure how to go about that when I was writing the code initially

arosen93: Regarding @janosh's point, I think the type hint itself is okay. After all, the user needs to use a pre-relax step that outputs a bandgap, and that's going to be impossible to get from `CHGNetRelaxMaker`.

Regarding your question about if it should be made compatible with `VaspInputGenerator`, I would lean towards no. It's probably best to keep it distinct since not all the fields in the schema can be populated from CHGNet.

janosh: I think `CHGNetRelaxMaker` is good the way it is and inherently different from `VaspInputGenerator`. My comment this morning was confused/not fully awake yet. 😄  `MPPreRelaxMaker` can stay the way it is. You would simply swap it out entirely in `MP2023RelaxMaker` if you want to use `CHGNetRelaxMaker`.

src/atomate2/vasp/flows/mp.py
arosen93: Actually, I like `MPMetaGGARelax` for the class name. I'll use that for now.

arosen93:
@rkingsbury: Thanks for your comment. I just added support for a static calculation before the relaxation per your request. I also added support for an optional static calculation _after_ the relaxation, in analogy to the current GGA workflow.

janosh:
@utf Once this is done, _please_ squash merge this. 😄

rkingsbury:
Looking great to me so far! Are you planning to replicate the unit tests we have in pymatgen / atomate for the legacy workflow?

arosen93:
Smart idea! I think we should.

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,214 @@
...
+            _get_kspacing_params(bandgap, bandgap_tol)
+        )
+
+        return super().make(structure=structure)
+
+
+@dataclass
+class MPStaticMaker(BaseVaspMaker):
+    """
+    Maker to create VASP static job using r2SCAN by default.
```
utf, line 136: Confused, this says r2SCAN by default by the default input set generator is `StaticSetGenerator` which is not r2SCAN?

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        self.input_set_generator.config_dict["INCAR"].update(
+            _get_kspacing_params(bandgap, bandgap_tol)
+        )
+
+        return super().make(structure=structure)
+
+
+@dataclass
+class MPStaticMaker(BaseVaspMaker):
```
utf, line 134: Should this not have r2SCAN in the name?

utf (commented):
Hi Prof @arosen93 and @janosh. This looks great. Excited to have it in atomate2.

A few questions:
1. It seems like we don't have the original MP PBE relaxation/band structure etc workflows in atomate2. Should we add these in? The could just be simple subclasses of the existing input sets just with the input set generator overriden.
2. Related to above, should we add versions of the inputset generators already configured to use the MP base. We could add a new file, `atomate2/vasp/sets/mp.py` with these input sets in. Again, these could be simple subclasses of the existing input sets with the config dict overriden.
3. Regardless of the above it would make sense to add r2SCAN to the names of the Makers to avoid any confusion.

### Moving kspacing logic

Perhaps a more contentious point. What do you think about moving the KSPACING logic to within `MPRelaxR2SCANGenerator.get_incar_updates` since then you don't need to pass the bandgap when linking up calculations (the bandgap is already available as an argument to this function). The only downside is that you won't be able to set the bandgap for the very first calculation, since it is always parsed from the previous calculation output. This shouldn't matter too much though since scan is mainly used after a pbesol pre-convergence.

### Changing kspacing defaults

An even more controversial point. I think if a band gap isn't available then the kspacing logic should default to an insulating value rather than a metallic one. Otherwise, your pre-converge step is always super expensive, as a metallic density k-point mesh is used. As this is only a pre-convergence and you do a double relaxation afterwards anyway, a lenient initial k-point mesh isn't a bad thing in my opinion.

janosh:
> Are you planning to replicate the unit tests we have in pymatgen / atomate for the legacy workflow?

@rkingsbury Yes, that's on my todo list for once the workflows have fully stabilized.

src/atomate2/vasp/jobs/mp.py
arosen93: You're right. Something must not have gotten saved correctly when we were doing the collaborative editing. I'll fix. Thanks!

src/atomate2/vasp/jobs/mp.py
arosen93: Good catch. I changed them all to `MPMetaStaticMaker`, `MPMetaRelaxMaker`, `MPMetaRelax`, etc.

arosen93:
> Hi Prof @arosen93 and @janosh. This looks great. Excited to have it in atomate2.

Thank you, Prof. Ganose! :D Lots of good ideas here.

> A few questions:
> 
> 1. It seems like we don't have the original MP PBE relaxation/band structure etc workflows in atomate2. Should we add these in? The could just be simple subclasses of the existing input sets just with the input set generator overriden.

That's right; we don't. One important caveat is that they would require the "old" PAW PBE potentials to be strictly compatible, but otherwise I think you are right that they should be added. Perhaps best saved for a follow-up PR though? Unless Janosh is willing.

> 2. Related to above, should we add versions of the inputset generators already configured to use the MP base. We could add a new file, `atomate2/vasp/sets/mp.py` with these input sets in. Again, these could be simple subclasses of the existing input sets with the config dict overriden.

Hmmm. What would be the benefit for this? From a user perspective, if they wanted to run any workflow with MP settings, couldn't they just supply a `config_dict` kwarg? I guess if we had a lot of MP workflows to add it could help make things more concise. Up to you. I'm indifferent, mostly.

> 3. Regardless of the above it would make sense to add r2SCAN to the names of the Makers to avoid any confusion.

Agreed. Thanks! I just added "Meta" to all the names, e.g. `MPMetaRelaxMaker`. My motivation for this naming schema was that `MPMetaGGARelaxMaker` started to get really long and unwieldy and `MPR2SCANRelaxMaker` was somewhat similar. Also, one could imagine passing another metaGGA functional like regular SCAN in, and then the name is a bit inconsistent. Happy to change further if you disagree, but they all have a clarifier in the name now.

> ### Moving kspacing logic
> Perhaps a more contentious point. What do you think about moving the KSPACING logic to within `MPRelaxR2SCANGenerator.get_incar_updates` since then you don't need to pass the bandgap when linking up calculations (the bandgap is already available as an argument to this function). The only downside is that you won't be able to set the bandgap for the very first calculation, since it is always parsed from the previous calculation output. This shouldn't matter too much though since scan is mainly used after a pbesol pre-convergence.

Hmmm, I do see the benefit. One counterpoint for us to consider though --- there has been some discussion amongst MP folks about one day using a "pre relax" that is based on an ML model (think M3GNet or CHGNet in style). Conceivably, it is possible that this pre-relax step might not produce a bandgap, and then the user would have to supply one (e.g. perhaps based on experiments or some other heuristic/ML model). That might be more difficult with moving the kspacing logic.

> ### Changing kspacing defaults
> An even more controversial point. I think if a band gap isn't available then the kspacing logic should default to an insulating value rather than a metallic one. Otherwise, your pre-converge step is always super expensive, as a metallic density k-point mesh is used. As this is only a pre-convergence and you do a double relaxation afterwards anyway, a lenient initial k-point mesh isn't a bad thing in my opinion.

For the pre-relax step, I agree with this decision.

Note for clarity: We've done a single relaxation afterwards and not a double for consistency with the original code. @rkingsbury: Do you recall if the R2SCAN relaxation step is done once or twice? Regardless, my personal opinion is that we should probably do a double regardless of what's in the original Atomate workflow.

rkingsbury:
> > ### Moving kspacing logic
> > Perhaps a more contentious point. What do you think about moving the KSPACING logic to within `MPRelaxR2SCANGenerator.get_incar_updates` since then you don't need to pass the bandgap when linking up calculations (the bandgap is already available as an argument to this function). The only downside is that you won't be able to set the bandgap for the very first calculation, since it is always parsed from the previous calculation output. This shouldn't matter too much though since scan is mainly used after a pbesol pre-convergence.
> 
> Hmmm, I do see the benefit. The only scenario where I could consider the R2SCAN step not having a prior band gap value is if an ML pre-relax step were done instead. Then it'd be more difficult to supply a band gap value (e.g. determined from experiments or another ML model). That said, for most meta-GGA functionals you usually need to start from a GGA wavefunction to get smooth SCF convergence, so I'm going to lean towards this being a non-issue and that we should adopt your plan.
> 
I don't see a problem with this (no strong opinion either way), so long as the overall workflow retains the bandgap-dependent KSPACING logic we describe in the paper.

> > ### Changing kspacing defaults
> > An even more controversial point. I think if a band gap isn't available then the kspacing logic should default to an insulating value rather than a metallic one. Otherwise, your pre-converge step is always super expensive, as a metallic density k-point mesh is used. As this is only a pre-convergence and you do a double relaxation afterwards anyway, a lenient initial k-point mesh isn't a bad thing in my opinion.
> 
> For the pre-relax step, I agree with this decision.
> 
I'm not in favor of this change, but I'm OK with it if y'all feel strongly. In my experience with the workflow, the PBEsol stop was always pretty fast despite metallic settings. @utf have you encountered specific problems with the PBEsol being real expensive? The thinking behind using metallic settings was to be conservative when the bandgap is unknown. That way you're sure to have a high-quality calculation even if the material turns out to be metallic.


> Note for clarity: We've done a single relaxation afterwards and not a double for consistency with the original code. @rkingsbury: Do you recall if the R2SCAN relaxation step is done once or twice? Regardless, my personal opinion is that we should probably do a double independent of what's in the original Atomate workflow. We want to make sure re-padding with ENCUT isn't an issue and that the bandgap doesn't change drastically over the course of the relaxation (less likely).

We evaluated single vs. double relaxations when developing the workflow, and found that the change in cell volume from the 1st to the 2nd r2SCAN relaxation was negligibly small (<0.4%), so based on this we opted to only do a single metaGGA relaxation. The GGA pre-relax is intended to serve the purpose of the first relaxation in a classic double-relax by getting the cell volume close enough.  Because we describe all this in the [paper](https://journals.aps.org/prmaterials/abstract/10.1103/PhysRevMaterials.6.013801) (see Appendix A, pasted below) - I'm not in favor of changing the workflow architecture unless we have data demonstrating negative consequences of skipping the 2nd relaxation.

image


In general, we went through a careful process guided by benchmarking data to develop the workflow, so although there is sensible logic behind these proposed changes, I personally would want to see data that shows why they are necessary before modifying the workflow. For user's sake I think it's important to stay as close to what's described in our papers as we can to minimize confusion.

arosen93:
Thanks for clarity, @rkingsbury. That all makes sense to me, and I'll defer to you on the above points. I am here to act as your scribe :)

mkhorton:
What what it’s worth, I agree that having a reference implementation (exactly as specified in paper) is very important, although I am actually more pro-modifications in general (e.g., I’m ok with changing the default behaviour if well-motivated, especially if it’s a clear improvement, provided the reference option remains).A well-motivated change in my view is one that doesn’t change the compatibility of the final energy but does either make the final output closer to the global minimum, or speeds up the calculation, both are good.On Jun 8, 2023, at 18:54, Andrew S. Rosen ***@***.***> wrote:﻿
Thanks for clarity, @rkingsbury. That all makes sense to me, and I'll defer to you on the above points. I am here to act as your scribe :)

janosh:
Just a minor point but wanted to say I'm in favor of the more explicit name `MPMetaGGARelaxMaker` over `MPMetaRelaxMaker`. Latter could be confused with a `BaseClass` from which to inherit to make different `MPRelaxMakers`. Might just be me though.

arosen93:
Sounds reasonable enough!

rkingsbury:
> Just a minor point but wanted to say I'm in favor of the more explicit name `MPMetaGGARelaxMaker` over `MPMetaRelaxMaker`. Latter could be confused with a `BaseClass` from which to inherit to make different `MPRelaxMakers`. Might just be me though.

Not just you! I agree. To me `Meta` is ambiguous - could be metaGGA but could be meta...something else. Personally I always lean towards longer and more explicit names to reduce chances for confusion.

janosh:
> Personally I always lean towards longer and more explicit names to reduce chances for confusion.

Same. Definitely sth I wish I'd started doing sooner. On my gravestone, I'll have them to inscribe "He learned the hard way the wisdom of loooong, explicit variable names".

matthewkuner:
Is it necessary to include "MP" at the start of any of the maker names? I feel like everyone knows it is Materials Project code anyway

arosen93:
Personally, I do think it's important to emphasize that this is the prescribed MP-compatible workflow and specs (as opposed to the Atomate2 defaults, for instance, which are not MP-compatible).

matthewkuner:
I'm not fully convinced. Isn't the default Vasp `RelaxMaker` MP-compatible as well, but just for the GGA functional? I also don't use xc functionals other than GGA so perhaps I don't know what I'm talking about

janosh:
Also in favor of keeping MP prefix. @matthewkuner unlike atomate1, atomate2 has wider scope than just MP. So there could one day be workflows from other research groups/DBs.

It's a nice form of branding too, adding a name recognition effect that might incentivize external contributions.

utf:
> Isn't the default Vasp RelaxMaker MP-compatible as well, but just for the GGA functional? 

@matthewkuner, no the default RelaxMaker is not MP compatible. It differs on a number of points, mainly:
- Use of PBEsol rather than PBE
- Use of new pseudopotentials

If you want to make the input sets MP compatible you need to update the base config. We have the atomate1 base config in atomate2 here: https://github.com/materialsproject/atomate2/blob/main/src/atomate2/vasp/sets/BaseMPGGASet.yaml

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,214 @@
...
+            based on the bandgap.
+
+        Returns
+        -------
+        MPRelaxMaker
+            The maker.
+        """
+        self.input_set_generator.config_dict["INCAR"].update(
+            _get_kspacing_params(bandgap, bandgap_tol)
+        )
```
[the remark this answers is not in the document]
arosen93: That seems logical, although I must admit I am not 100% confident where that move should be made. @janosh, whenever you have a moment, would you mind seeing if you can address Alex's comment here? Thank you! 🙏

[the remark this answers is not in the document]
janosh: Certainly! meeting a lot of people today and tomorrow so might have to wait 2 days.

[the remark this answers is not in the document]
arosen93: That's not an issue :) Please enjoy your time!

janosh:
Finally had time to pick this up again and started implementing `test_mp_relax()` with mock VASP runs in https://github.com/materialsproject/atomate2/pull/362/commits/ffcc2817c94aa2936e0983e57525e5df24bc9f3e. But getting weird error messages. @arosen93 / @utf are we setting the job name wrong?


<details><summary>Test logs</summary>


```py
---------------------------------- Captured log call ----------------------------------
INFO     jobflow.managers.local:local.py:152 Started executing jobs locally
INFO     jobflow.core.job:job.py:526 Starting job - BaseVaspMaker.make (5d4e3584-bb0b-4baa-aac5-1639c365061c)
INFO     atomate2.vasp.files:files.py:196 Writing VASP input set.
INFO     jobflow.managers.local:local.py:102 BaseVaspMaker.make failed with exception:
Traceback (most recent call last):
  File "/Users/janosh/.venv/py311/lib/python3.11/site-packages/jobflow/managers/local.py", line 98, in _run_job
    response = job.run(store=store)
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/janosh/.venv/py311/lib/python3.11/site-packages/jobflow/core/job.py", line 544, in run
    response = function(*self.function_args, **self.function_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/janosh/dev/atomate2/src/atomate2/vasp/jobs/base.py", line 159, in make
    run_vasp(**self.run_vasp_kwargs)
  File "/Users/janosh/dev/atomate2/tests/vasp/conftest.py", line 81, in mock_run_vasp
    ref_path = vasp_test_dir / _REF_PATHS[name]
                               ~~~~~~~~~~^^^^^^
KeyError: 'BaseVaspMaker.make'
```
</details>

arosen93:
Weird... I'm also not sure about this one...

esoteric-ephemera:
> ## Overview
> 
> Closes #352. Closes #363. Closes #364.
> 
>     * [Port SCAN workflows to atomate2 #352](https://github.com/materialsproject/atomate2/issues/352)
> 
>     * [BUG: `LMAXTAU` should be auto-set when there are f electrons and `LASPH` is True for meta-GGA calculations #363](https://github.com/materialsproject/atomate2/issues/363)
> 
>     * [BUG: `LMAXMIX` should be set based on the presence of s, p, d, f electrons not based on Z #364](https://github.com/materialsproject/atomate2/issues/364)
> 
> 
> ## R2SCAN Workflow
> 
> Note: This is in close collaboration with @janosh!
> 
> This PR adds MP-compatible r2SCAN workflows to Atomate2 (#352). We are still adding tests for now, hence the WIP.
> ## VASP INCAR Improvements
> 
> I also baked in fixes for several VASP INCAR issues I brought up (#363, #364, #365).
> ## Review Request
> 
> @utf: Mind giving this a quick look to make sure we didn't do anything drastically wrong before we go and write all the tests? Thanks!
> 
> Also tagging @munrojm, @rkingsbury, and @mkhorton in case they're interested in following this.
> ## ToDo
> 
>     * Finish adding unit tests
> 
>     * Only copy the `CHGCAR` (not all the files) from the PBEsol job to the R2SCAN job in order to remain truthful to the original workflow.

Suggest using WAVECAR from the PBEsol run + ISTART = 1 to initialize r2SCAN relaxation. r2SCAN needs the kinetic energy density (KED) in addition to the charge density to run. Both can be computed from WAVECAR, but the KED can't be computed from CHGCAR. Should help with efficiency / preconditioning run, maybe higher memory / disk usage

If WAVECAR isn't included but CHGCAR is, VASP will probably start with randomized wavefunctions, which slows things down

Similar situation with hybrid calculations - you need the WAVECAR to compute exact exchange for the initial step

arosen93:
I have been a strong proponent for copying the WAVECAR. @rkingsbury claimed to have issues with this approach, but I am pretty confident it is simply associated with changing between the gamma-point only version of VASP to the standard version of VASP. Custodian should now have a handler for that, and we can add logic here to account for it.

rkingsbury:
> I have been a strong proponent for copying the WAVECAR. @rkingsbury claimed to have issues with this approach, but I am pretty confident it is simply associated with changing between the gamma-point only version of VASP to the standard version of VASP. Custodian [now has a handler for that](https://github.com/materialsproject/custodian/pull/256), and we can add logic here to proactively account for it if we wish. But it would indeed be a departure, technically.
> 
> I actually see only minimal speedup when copying the CHGCAR in most cases, which isn't surprising.

Good idea; I have no objections to this. I tried it initially and had problems, but they might well have been due to the early state of the workflow development or to my own incompetence. If you can get it to work and it shows a speedup, go for it!

rkingsbury:
I will add a note that when this is done, someone please remember to update the public MP docs so that future workflows users can understand some of the subtle changes that y'all have made since the paper. They're all good enhancements!

We could just add this discussion as an additional reference on the metaGGA page, and perhaps add a short synopsis of the enhancements (WAVECAR, LMAXMIX, etc.)

arosen93:
Is there any reason why `ISYM = 0` is not included in the `MPSCANRelaxSet`? I pulled up an example MP calculation INCAR and didn't see it. This seems like a major oversight if it's not included... 😨

janosh:
> This seems like an oversight if it's not included since we always want to do relaxations without symmetry constraints... 

Maybe sth to discuss in Tuesday's update call? @esoteric-ephemera will be there to bring up a similar potential issue.

rkingsbury:
> Is there any reason why `ISYM = 0` is not included in the `MPSCANRelaxSet`? I pulled up an example MP calculation INCAR and didn't see it. This seems like an oversight if it's not included since we always want to do relaxations without symmetry constraints... 😨 Thoughts, @munrojm?

It looks like `ISYM=0` is not included in either `MPRelaxSet` or `MPScanRelaxSet` (at least not via the YAML files. Pardon my ignorance, but don't we want symmetry turned _on_ ? If I'm reading the [VASP wiki correctly](https://www.vasp.at/wiki/index.php/ISYM), `ISYM=0` disables symmetry and is only appropriate for MD calculations.

JaGeo:
I also never had an issue when optimizing crystal structures with symmetry (even the phonons are very good afterwards!). Though, it can prevent that the algorithms find the correct minimum. Especially for molecules, however, this is a huge issue.

JaGeo:
And, @rkingsbury , you are absolutely correct, it is also not part of the MPRelaxSet.

arosen93:
It's been my understanding that it is generally considered poor practice to do a geometry optimization with symmetry enabled (in any DFT program) because if your initial structure is in a high symmetry state, you will often be constrained to that same high symmetry state instead of letting the geometry optimizer potentially break the symmetry to find a lower energy state.

Perhaps this is one of those things that's more relevant for molecules than materials (where this suggestion definitely is crucial). After all, if you're pulling real crystal structures from the ICSD, most crystalline solids will probably have the same symmetry at the end of the geometry optimization either way.

But I do see it's not part of any MP calculations right now, so I'll defer to history on that one.

JaGeo:
@arosen93 No worries. I think I edited my comments three times 🙃. And, it's also a good sign that we come up with the same thoughts independently!

janosh:
[Slightly related recent `ISYM` discussion](https://berkeleytheory.slack.com/archives/C8ZSHT4TB/p1688423482102889).

JaGeo:
Reactivated my old account to read it. Don't think it is related. It has something to do with the CHGCAR reuse, I would suspect. We also had issues with MAGMOMs not being read correctly while using our Lobster workflow. "ICHARG" was key. Luckily, a very good Master student spotted this.

src/atomate2/vasp/sets/mp.py
arosen93, line 1: It was probably called `kspacing` intentionally because the INCAR flag is called KSPACING.

janosh: Sure thing. vars not using snake_case is one of my pet peeves. 😄

src/atomate2/vasp/sets/base.py
```diff
@@ -948,12 +945,13 @@ def _set_u_params(incar, incar_settings, structure):
     # investigation it was determined that this would lead to a significant difference
     # between SCF -> NonSCF even without Hubbard U enabled. Thanks to Andrew Rosen for
     # investigating and reporting.
-    if "LMAXMIX" not in incar_settings.keys():
```
arosen93, line 951: @esoteric-ephemera: Can you undo this change re: the LMAXMIX logic? See https://github.com/materialsproject/pymatgen/pull/3041#issuecomment-1650336881 for details.

mkhorton:
Context on `LELF`: Ryan was originally generating this for analysis and benchmarking, however we thought it was a good idea in general for MP production with the logic that (1) it’s cheap(ish) to generate at the end of a calculation but expensive to go back and regenerate for a lot of structures, if we decide we want a high-throughput analysis, (2) since MP now offers a charge density endpoint, the same infrastructure could be reused to offer an ELF endpoint.

Since the ELF contains a lot of useful information this might be beneficial, but MP was unlikely to do a dedicated calculation campaign just for this; if the data were already available however, it would significantly lower the barrier.

However, I realise the constraint it puts on NPAR might make this not suitable as a default. Above info just for context.

(Apologies if this is repost, don’t think my original comment went through)

rkingsbury:
> 

Thanks for the reminder @mkhorton ! I had forgotten about that discussion re: possibly future utility of the data. At the time, we weren't aware that LELF constrained KPAR at all as far as I remember.

src/atomate2/vasp/sets/BaseMPGGASet.yaml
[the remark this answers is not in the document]
esoteric-ephemera: I think this is what we want to use moving forward, right? There's also no Yb_3 in the old PBE PSPs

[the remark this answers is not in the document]
arosen93: Tagging @munrojm but that would require a complete recalculation of the entire MP database since they are not comparable. Maybe that's what you all want to do (it'd be great), but then this atomate2 workflow would not be compatible with reproducing results on MP as it stands right now.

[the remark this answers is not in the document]
arosen93: For instance, if I wanted to calculate some GGA-compatible formation energies and I used the PBE_54 pseudos, I would have values that aren't comparable with MP as produced today.

mkhorton:
> Thanks for the reminder @mkhorton ! I had forgotten about that discussion re: possibly future utility of the data. At the time, we weren't aware that LELF constrained KPAR at all as far as I remember.

I think we were aware, there was a Custodian rule(?) but my memory is not perfect. In either case, maybe a bit academic now.

src/atomate2/vasp/sets/BaseMPGGASet.yaml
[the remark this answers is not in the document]
arosen93: The atomate2 workflow isn't just for MP --- it's also for the general public and for other groups. Until all of MP is switched to .54 pseudos, I think this should be made into a separate YAML file altogether.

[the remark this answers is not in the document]
esoteric-ephemera: Re recomputing all of MP, I think this was ultimately the goal but @munrojm can comment. I'm happy supporting that effort and we can discuss that

Maybe good to have an MPGGACompatSet and MPCompatMaker that uses the OG MP design? The GGA WF also by default always does GGA + U and initializes the second relax and final static from CHG/WAVE CAR, so that deviates from the OG WF too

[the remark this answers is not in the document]
arosen93: > Re recomputing all of MP, I think this was ultimately the goal but @munrojm can comment. I'm happy supporting that effort and we can discuss that

I definitely think a giant recompute would be great but of course that's not my place 😄 

> Maybe good to have an MPGGACompatSet and MPCompatMaker that uses the OG MP design? The GGA WF also by default always does GGA + U and initializes the second relax and final static from CHG/WAVE CAR, so that deviates from the OG WF too

There's an important distinction to be made between the workflow itself having changes and the calculations being compatible. For the end-user, it's the latter that is crucial. After all, there are always updates to Custodian and Pymatgen such that the workflow is never really exactly the same. But if a user wants to get the energy above hull of a new material, the level of theory (including pseudopotentials, functional, etc.) are going to need to be identical.

So, with that being said, I agree that there needs to be some mechanism for computing GGA(+U)-compatible properties with how MP exists right now. That doesn't necessarily mean the workflow is exactly the same (e.g. CHG/WAVECAR differences are fine IMO), but it does mean pseudos are compatible. Otherwise things are going to get problematic very fast from folks using atomate2 😅

matthewkuner:
Wait, so current calcs in the MPDB all use 5.2 pseudos? Or just the R2scan ones?

arosen93:
@janosh:

I'm fairly certain that is not true. All PBE calcs use the "regular" (no version number) pseudos!

janosh:
@arosen93 [You're right](https://github.com/materialsproject/pymatgen/blob/313eae0b3d5ef30b48eac97a5458554d372c993c/pymatgen/io/vasp/MPRelaxSet.yaml#L83-L84), apologies!

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,184 @@
...
+                structure=output.structure, prev_vasp_dir=output.dir_name
+            )
+            output = final_static.output
+            jobs += [final_static]
+
+        return Flow(jobs, output, name=self.name)
+
+
+@dataclass
+class MPMetaGGARelax(Maker):
```
utf, line 111: Again with this name.

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,184 @@
...
+    Parameters
+    ----------
+    name : str
+        Name of the flows produced by this maker.
+    initial_relax_maker : .BaseVaspMaker
+        Maker to generate the first relaxation.
+    initial_static_maker : .BaseVaspMaker
+        Maker to generate the static calculation before the relaxation.
+    final_relax_maker : .BaseVaspMaker
+        Maker to generate the second relaxation.
```
utf, line 124: Missing copy_vasp_files docstring.

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,184 @@
...
```
utf, line 48: Missing copy_vasp_files docstring.

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,217 @@
...
+        Additional data to write to the current directory. Given as a dict of
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    name: str = "MP GGA Relax"
+    input_set_generator: VaspInputGenerator = field(
+        default_factory=lambda: MPGGARelaxGenerator(auto_ismear=False)
```
utf, line 52: If the default value of `auto_ismear` for this calculation is False, then you can set that directly in the `MPGGARelaxGenerator` rather than here.

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,217 @@
...
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    name: str = "MP GGA Static"
+    input_set_generator: VaspInputGenerator = field(
+        default_factory=lambda: MPGGARelaxGenerator(
+            user_incar_settings={"NSW": 0, "ISMEAR": -5, "LREAL": False},
```
utf, line 88: Rather than using a Relax generator, can you make a `MPGGAStaticSetGenerator` with these adjustments. You can see what we did for the regular static set generator.

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,217 @@
...
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    name: str = "MP pre-relax"
+    input_set_generator: VaspInputGenerator = field(
+        default_factory=lambda: MPMetaGGARelaxGenerator(
+            user_incar_settings={
```
utf, line 126: Again, these settings should be defined in the input set generator.

src/atomate2/vasp/jobs/mp.py
```diff
@@ -0,0 +1,217 @@
...
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    name: str = "MP meta-GGA static"
+    input_set_generator: VaspInputGenerator = field(
+        default_factory=lambda: MPMetaGGARelaxGenerator(
+            user_incar_settings={
```
utf, line 207: Again, make a static set generator and define these in the `get_incar_updates` function.

src/atomate2/vasp/sets/mp.py
```diff
@@ -0,0 +1,123 @@
...
+            A vasprun from a previous calculation.
+        outcar
+            An outcar from a previous calculation.
+
+        Returns
+        -------
+        dict
+            A dictionary of updates to apply.
+        """
+        return {}
```
utf, line 68: You need to set these properly so that it will configure the necessary settings for a relaxation. E.g., see here: https://github.com/materialsproject/atomate2/blob/18d014abcf2a986b78653acca205e8713107b165/src/atomate2/vasp/sets/core.py#L71

Although if using the subclassing approach detailed in my previous comment this shouldn't be necessary.

src/atomate2/vasp/sets/mp.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        bandgap = self.bandgap_override or bandgap or 0
+
+        if bandgap < self.bandgap_tol:  # metallic
+            return {"KSPACING": 0.22, "ISMEAR": 2, "SIGMA": 0.2, "GGA": None}
+
+        rmin = 25.22 - 2.87 * bandgap
+        kspacing = 2 * np.pi * 1.0265 / (rmin - 1.0183)
+        return {
+            "KSPACING": np.clip(kspacing, 0.22, 0.44),
```
utf, line 118: Again, this is missing the tags to actually make this a relaxation if continuing from a static calculation.

src/atomate2/vasp/sets/mp.py
```diff
@@ -0,0 +1,123 @@
...
+)
+
+
+_BASE_MP_GGA_RELAX_SET = loadfn(
+    resource_filename("atomate2.vasp.sets", "BaseMPGGASet.yaml")
+)
+
+
+@dataclass
+class MPGGARelaxGenerator(VaspInputGenerator):
```
utf, line 34: A few comments:
- The standard notation is for input set generators to have `SetGenerator` at the end. So this (and the other generator in this file) would become `MPGGARelaxSetGenerator`. It is verbose but consistent with the other generators.
- I think rather than defining this whole class, you can just do:

```python
@dataclass
class MPGGARelaxSetGenerator(RelaxSetGenerator):
    config_dict: dict = field(default_factory=lambda: _BASE_MP_GGA_RELAX_SET)
```

Note that we are subclassing the default RelaxSetGenerator. Everything should then work out of the box.

Similarly, you can make a static set generator using:
```python
@dataclass
class MPGGAStaticSetGenerator(StaticSetGenerator):
    config_dict: dict = field(default_factory=lambda: _BASE_MP_GGA_RELAX_SET)
```

src/atomate2/vasp/sets/mp.py
```diff
@@ -0,0 +1,123 @@
...
+        -------
+        dict
+            A dictionary of updates to apply.
+        """
+        return {}
+
+
+@dataclass
+class MPMetaGGARelaxGenerator(VaspInputGenerator):
+    """Class to generate MP-compatible VASP metaGGA relaxation input sets."""
```
utf, line 73: Can you describe bandgap_tol etc in the docstring rather than with code comments.

tests/vasp/flows/test_mp.py
```diff
@@ -0,0 +1,109 @@
+import pytest
+from pymatgen.core import Structure
+
+from atomate2.vasp.flows.mp import MPMetaGGARelax
+from atomate2.vasp.jobs.mp import (
+    MPMetaGGARelaxMaker,
+    MPPreRelaxMaker,
+)
+
```
utf, line 9: All inputs need to be within the tests themselves otherwise it breaks the tests. This is mentioned at the bottom of this page but perhaps we need to make this more clear elsewhere: https://materialsproject.github.io/atomate2/dev/vasp_tests.html

utf (commented):
Hi @janosh, I've added some comments. 

One potential reason for the tests failing could be that module imports for the makers is not inside the test functions themselves. This is necessary for the monkey patching to work properly. I added a comment about this with more details.

src/atomate2/vasp/sets/mp.py
janosh: Where does `auto_ismear` come into this? Like so?

```py

@dataclass
class MPGGARelaxSetGenerator(RelaxSetGenerator):
    """Class to generate MP-compatible VASP GGA relaxation input sets."""

    config_dict: dict = field(default_factory=lambda: _BASE_MP_GGA_RELAX_SET)

    def __post_init__(self):
        self.auto_ismear = False
        return super().__post_init__()
```

utf: Even simpler, just:
```
@dataclass
class MPGGARelaxSetGenerator(RelaxSetGenerator):
    """Class to generate MP-compatible VASP GGA relaxation input sets."""

    config_dict: dict = field(default_factory=lambda: _BASE_MP_GGA_RELAX_SET)
    auto_ismear: bool = False
```

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,184 @@
...
+    from pymatgen.core.structure import Structure
+
+    from atomate2.vasp.jobs.base import BaseVaspMaker
+
+
+__all__ = ["MPGGARelax", "MPMetaGGARelax"]
+
+
+@dataclass
+class MPGGARelax(Maker):
```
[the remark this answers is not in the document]
janosh: Are you suggesting `MPGGADoubleRelaxStatic`? That doesn't fit either given the initial relax and static at the end are optional and might not be there.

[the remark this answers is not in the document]
utf: I think that is a better name. The initial relax and static are there by default so I think that is reasonable.

tests/vasp/flows/test_mp.py
janosh: I tried moving all imports except `pytest` into the test function yesterday which made no difference, tests were failing either way.

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,184 @@
...
+    initial_relax_maker : .BaseVaspMaker
+        Maker to generate the first relaxation.
+    initial_static_maker : .BaseVaspMaker
+        Maker to generate the static calculation before the relaxation.
+    final_relax_maker : .BaseVaspMaker
+        Maker to generate the second relaxation.
+    """
+
+    name: str = "MP GGA Relax"
+    initial_maker: BaseVaspMaker | None = field(default_factory=MPGGARelaxMaker)
```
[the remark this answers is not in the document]
janosh: What do you think about having a more versatile `mp_compat_converter()` function that takes in any existing `atomate2` `Maker` and applies a `config_dict` of the user's choosing which defaults to `src/atomate2/vasp/sets/BaseMPGGASet.yaml`?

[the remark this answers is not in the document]
utf: I think that would be a nice feature. I think I suggested that at some point. Although I think for the canonical MP workflows we should define them explicitly. I image the function you describe (I think it should be a powerup) would be available for other people to convert non-MP workflows into MP compatible ones. If we use it for defining the canonical workflows it will be a bit confusing for users reading the code trying to understand what is going on.

src/atomate2/vasp/flows/mp.py
```diff
@@ -0,0 +1,223 @@
...
+    from pymatgen.core.structure import Structure
+
+    from atomate2.vasp.jobs.base import BaseVaspMaker
+
+
+__all__ = ["MPGGARelax", "MPGGADoubleRelaxMaker", "MPMetaGGARelax", "MPMetaGGADoubleRelaxMaker"]
+
+
+@dataclass
+class _MPGenericThreeStep(Maker):
```
utf, line 38: Personally, I would be against defining the flows like this. It is quite complicated for users trying to understand what is going on. I don't feel that there is much boilerplate in a DoubleRelax + Static Maker.

esoteric-ephemera: I've rewritten these: the hidden generic maker has been replaced by MPGGADoubleRelaxMaker. This maker features an optional final static, which is disabled by default. The other flows (MPGGARelax, MPMetaGGARelax, and MPMetaGGADoubleRelaxMaker) all inherit from MPGGADoubleRelaxMaker

janosh:
@arosen93 @esoteric-ephemera I just noticed `kspacing` jumps from its max value of 0.44 for large band gaps down to 0.22 if the band gap exceeds 8.78 eV where we actually want it to stay at 0.44. Could you confirm the fix in 4bf816a?

esoteric-ephemera:
It should work correctly now. Also setting GGA = None here is still unfortunately overriden by the previous INCAR, so we should update it in the relax and static meta-GGA makers through user_incar_settings

I'm not sure if the following is more transparent, basically defining the kspacing as a discontinuous piecewise function (KSPACING = 0.22 for gap < bandgap_tol, but 0.26 for gap = bandgap_tol):

    def kspacing_from_gap(bandgap,
        bandgap_tol = 1.e-4,
        min_kspacing = 0.22,
        max_kspacing = 0.44
    ):
        
        rmin_intcpt = 25.22
        rmin_slope = - 2.87
        rmin_b = 1.0183
        rmin_a = 2*np.pi*1.0265

        max_bandgap = (rmin_a/max_kspacing + rmin_b - rmin_intcpt)/rmin_slope

        updt_pars = {"KSPACING": 0.44, "ISMEAR": 0, "SIGMA": 0.05, "GGA": None}
        
        if bandgap < bandgap_tol:
            updt_pars = {"KSPACING": min_kspacing, "ISMEAR": 2, "SIGMA": 0.2, "GGA": None}
        elif bandgap < max_bandgap - bandgap_tol:
            rmin = rmin_intcpt + rmin_slope * bandgap
            updt_pars['KSPACING'] = rmin_a / (rmin - rmin_b)
        else:
            updt_pars['KSPACING'] = max_kspacing

        return updt_pars

utf:
Rather than setting things in the user_incar_settings, can you set them in the actual `get_incar_updates` function on the input set generators themselves. I made some comments about it on the PR which should still be visible.

janosh:
@utf #504 and this PR were mostly overlapping, i.e. only only should be merged. I think we need to revert this merge.

utf:
Ok, are you able to handle the unmerging?

janosh:
Yes, do you prefer a revert commit or rewriting git history like it never happened? The latter would require people who pulled `main` in the last 20 minutes to do a `git reset --hard @{u}`.

utf:
Can you overwrite the history? Should be ok I think.

PR #375, opened 2023-06-08
Fix gh 374

### LINKED ISSUE #374: `update_user_incar_settings` does not correctly set MAGMOMS for VASP jobs
https://github.com/materialsproject/atomate2/blob/0433cd8fc83f7bec9b2dc22de63d9431ad6f4064/src/atomate2/vasp/powerups.py#L14

If a previous structure has magmoms, those magmoms will be used with a higher priority than the magmoms set via the powerup `update_user_incar_settings`. I believe this may stem from [this code snippet](https://github.com/materialsproject/atomate2/blob/9c99ccd7b1e54b39dfddcc798a98d90f60e48475/src/atomate2/vasp/sets/base.py#L851) , but I could be mistaken.

@utf

janosh:
Closes #374.

This PR gives `magmoms`  passed to `_get_magmoms()` precedence over magmoms already in the structure.

It also adds a test for `_get_magmoms()`.

### DIFF
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -860,14 +860,14 @@ def _get_magmoms(magmoms, structure):
         "magmom on the site directly to ensure correct initialization."
     )
     for site in structure:
-        if hasattr(site, "magmom"):
-            mag.append(site.magmom)
-        elif hasattr(site.specie, "spin"):
-            mag.append(site.specie.spin)
-        elif str(site.specie) in magmoms:
+        if str(site.specie) in magmoms:
             if site.specie.symbol == "Co" and magmoms[str(site.specie)] <= 1.0:
                 warnings.warn(msg, stacklevel=2)
             mag.append(magmoms.get(str(site.specie)))
+        elif hasattr(site, "magmom"):
+            mag.append(site.magmom)
+        elif hasattr(site.specie, "spin"):
+            mag.append(site.specie.spin)
         else:
             if site.specie.symbol == "Co":
                 warnings.warn(msg, stacklevel=2)
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -598,13 +598,13 @@ class VaspInputGenerator(InputGenerator):
 
     def _get_incar(
         self,
-        structure,
+        structure: Structure,
         kpoints: Kpoints,
         previous_incar: dict = None,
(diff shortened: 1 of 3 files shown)


utf:
Thanks @janosh. I think this might need a bit more thought.

Previously the magmom precedence was:
- magmoms in input struct (which in the below case is the opt1_job.output.structure, which contains magmoms)
- spins in input struct
- user incar settings
- config dict
- set all magmoms to 0.6

The actual precedence for magmoms should be:
- user incar settings
- magmoms in input struct (which in the below case is the opt1_job.output.structure, which contains magmoms)
- spins in input struct
- config dict
- set all magmoms to 0.6

However, this PR has the order
- user incar settings
- config dict
- magmoms in input struct (which in the below case is the opt1_job.output.structure, which contains magmoms)
- spins in input struct
- set all magmoms to 0.6

That's because the config dict is updated with `user_incar_settings`, and the magmom settings are then passed to `_get_magmoms`. E.g., see here: 
https://github.com/materialsproject/atomate2/blob/6c541cdf1a09ad37e4f3753b45a455d1387c0071/src/atomate2/vasp/sets/base.py#L611C8-L614

And here:
https://github.com/materialsproject/atomate2/blob/6c541cdf1a09ad37e4f3753b45a455d1387c0071/src/atomate2/vasp/sets/base.py#L620

### Solution

To get the right precedence, one thing we could do is:
```
1. Before https://github.com/materialsproject/atomate2/blob/6c541cdf1a09ad37e4f3753b45a455d1387c0071/src/atomate2/vasp/sets/base.py#L614 copy the config dict magmoms to a separate variable.
2. Update the `_get_magmom` function to also accept the original config dict, and update the logic with the correct precedence.

It would also be nice to update the precedence of how magmoms are set in both the docstring `get_magmom` and as a note in the docstring of the VaspInputGenerator class.

**Comment by janosh** [comment]
@utf Thanks for the quick feedback. By config dict magmoms, you mean `incar_magmoms = incar_settings.get("MAGMOM", {})`?

**Comment by utf** [comment]
Yes, although this should be before the `user_incar_settings` are applied. E.g.,

```python
incar_settings = dict(self.config_dict["INCAR"]) 
config_magmoms = incar_settings.get("MAGMOM", {})

# apply user incar settings to SETTINGS not to INCAR 
_apply_incar_updates(incar_settings, self.user_incar_settings) 
```

janosh:
Lunch now! I'll do the docs update later.

PR #376, opened 2023-06-08
Update Lobster documentation

JaGeo:
## Summary
I would like to refer to the new jobflow tutorials on fireworks addedin https://github.com/materialsproject/jobflow/pull/338 directly in the documentation of the Lobster workflow part. I think it is especially useful there.


utf:
Good idea. Thank you.

PR #378, opened 2023-06-08
Update FireWorks section of docs

arosen93:
This is to match and leverage the updates in the Jobflow docs.


arosen93:
@utf, this is ready to go.

utf:
Thanks Andrew

PR #381, opened 2023-06-09
Document architectural difference between atomate 1 and 2

janosh:
Thanks @arosen93, this is based on your comments from the email exchange with @computron. i think this might be helpful background and historical context for newcomers to understand the most significant architectural difference between `atomate` 1 and 2.

@utf Consider this a request for comments. Feel free to close if you don't think this helpful.


JaGeo:
I think it would be useful to add how our plans for the future look in this regard, e.g. which support is already planned. I think it might also be worth repeating parts of the readme, i.e. that atomate2 wants to cover all use cases from 1 job to 10.000s.

janosh:
@JaGeo Sounds good. Feel free to suggest changes to the PR. Although maybe you want to wait for @utf's feedback.

utf:
@JaGeo yes, please add any and all comments. I'll have a look at this too later today.

docs/user/atomate-1-vs-2.md
```diff
@@ -0,0 +1,25 @@
+# Atomate 1 vs 2
+
+This document contains introductory context for people coming from atomate 1.
+
```
JaGeo, line 4:
```suggestion
One of atomate2's core ideas is to allow scaling from a single material, to 100 materials, or 100,000 materials. Therefore, both local submission options and a connection to workflow managers such as FireWorks exist. We plan to support more workflow managers in the future to further ease job submission.
```

matthewkuner:
As someone who has not used Atomate1, could you elaborate on what particular feature(s) enable Atomate2 to better "allow scaling from a single material, to 100 materials, or 100,000 materials" ?

JaGeo:
You can use it without a db.

arosen93:
> As someone who has not used Atomate1, could you elaborate on what particular feature(s) enable Atomate2 to better "allow scaling from a single material, to 100 materials, or 100,000 materials" ?

A lot of it is due to Jobflow. By using Jobflow as the backend, you could choose to run the job locally (e.g. submit it as a standard Python SLURM job) or with fireworks (for large-scale calculations). Both approaches use the same exact Atomate2 recipes.

janosh:
> @JaGeo yes, please add any and all comments! I'll have a look at this too later today.

@utf Have you had a chance to review this?

Zhuoying (commented):
Maybe we can use two executable modes to distinguish atomate 1 and 2:
(i) single material--local submission--no db needed--no job manager (e.g.: Fireworks) as a dependency
(ii) vast materials--connection to job managers --connection to db

Previously in atomate 1, we only have mode (ii) even for a single material. Now in atomte2, we can do both and try to support more job managers (future plan) and various db types (not only MongoDB).

After telling the main architectural difference, we can leave some blank space or to-be-added for **job managers** and list currently supported db types in the **connection to db** part.

janosh:
@Zhuoying Good summary! Feel free to tweak `docs/user/atomate-1-vs-2.md`.

janosh:
I'll merge this for now. Future PRs that update/tweak `atomate-1-vs-2.md` are welcome (from my side).

PR #395, opened 2023-06-21
`ruff` select `perflint` `flake8-slots`

janosh:
[PEP 484](https://peps.python.org/pep-0484/) prohibits implicit `Optional`. To be a good Python citizen, this PR removes implicit optionals from all type hints.


utf:
Personally, I'm not in favour of this. I really don't see the point of having to add the optional type when the default value is `None`. Mypy understands it fine. It just seems like extra boilerplate.

However, if there are other strong opinions I will hear them.

janosh:
@utf Don't feel strongly. I agree it's boilerplate even if more rigorous. I reverted the `Optional` changes and ignored the `ruff` rule `RUF013` that applied them.

utf:
Great, thank you.

PR #398, opened 2023-06-23
Phonons for forcefields

### LINKED ISSUE #371: Connection of force-field models to phonon workflow
@QuantumChemist and I plan to work on connection between the force field methods and a phonon workflow.

JaGeo:
## Summary

Closes #371  (in the future). It's a work in progress. 

Plans: 
- move more phonon job parts in common
- generalize the phonon jobs and schema

This is a first output for Si of the workflow with CHGNet:
image

We will also validate the results carefully.


JaGeo:
@gpetretto FYI. @QuantumChemist and I will work on this implemention in the next 1-2 months.

JaGeo:
Closes #409 as well.

JaGeo:
I have added the force constants to the data store as well. I think mostly tests are missing and an assessment if we can provide a common workflow.

JaGeo:
For Mg3Sb2, the phonon band structures show imaginary modes. However, this might potentially be expected. Overall still close results to DFT:
image

JaGeo:
@utf and maybe also @gpetretto: maybe you have some thoughts with regard to generalization. As mentioned above, BORN charges and potentially also relying on previous DFT computations are an issue for generalization of the workflow.

E.g., this line: https://github.com/JaGeo/atomate2/blob/ae043eb0d321ad630e4593dd68bc6a0042ebbe6e/src/atomate2/vasp/flows/phonons.py#L314

or this line:
https://github.com/JaGeo/atomate2/blob/ae043eb0d321ad630e4593dd68bc6a0042ebbe6e/src/atomate2/vasp/flows/phonons.py#L240

JaGeo:
I adapted my work plan a bit. Some parts could also be included within another PR.

gpetretto:
> @utf and maybe also @gpetretto: maybe you have some thoughts with regard to generalization. As mentioned above, BORN charges and potentially also relying on previous DFT computations are an issue for generalization of the workflow.

Hi @JaGeo, thanks for the great work with this PR and sorry for the delay in commenting this.
Concerning the generalization, this will likely require some global discussion. For the first example, we can indeed open a discussion about how to expose some output values in a standard way across different codes.  
In general it would be really nice to have a subsection of the TaskDoc to insert standard physical properties. Output documents from different code would then be expected to expose them in order to be used in generic workflows. In your example, accessing born like
```python
born = born_job.output.properties.born
```
instead of 

```python
born = born_job.output.calcs_reversed[0].output.outcar["born"]
```
Other properties that could be in this subsection could be energies, forces and stress.

For the `prev_vasp_dir`, I am not sure how common of a use case will be to start a relaxation of a phonon workflow from a `prev_vasp_dir`, but an option could be exposing a `relax_maker_kwargs` in the general workflow. Do you think this could be a viable option?

JaGeo:
I very much like the idea of exosing the `relax_maker_kwargs`. We might want to do this for all current existing flows. (Which might be too much for this PR)

JaGeo:
Thanks to @yuanbinliu for the test files.

Once the linting and tests finally pass, this is ready for review (@utf). It's not urgent. We can build our current work on the current state of the pull request.

JaGeo:
only the docs fail, as I have no key for deployment.

src/atomate2/common/jobs/phonons.py
```diff
@@ -279,7 +284,7 @@ def run_phonon_displacements(
         "forces": [],
         "uuids": [],
         "dirs": [],
-        "displaced_structures": [],
+        # "displaced_structures": [],
```
utf, line 287: Remove?

utf:
Thanks very much for this @JaGeo. I think we can leave the discussion for merging the vasp and forcefield phonon workflows to a separate PR. I have some ideas for this which I can raise in an issue.

I put a couple of very minor comments on the PR. Can you also rename `atomate2/forcefields/flows/flows.py` to `atomate2/forcefields/flows/relax.py` and import the two classes in that file in `atomate2/forcefields/flows/__init__.py` to avoid any breaking changes.

JaGeo:
Thanks. 
Could you make the comments public? I don't see them yet.

utf:
Fantastic. Thanks very much1

PR #408, opened 2023-06-27
Docs: add basic workflow tutorial

rkingsbury:
## Summary

Draft "how to write a workflow tutorial" ported over from the WIP MP Contributor Guide


arosen93:
Looks great, @rkingsbury! One small comment, which may or may not be relevant for you.

There is a package called [cclib](https://cclib.github.io/) that is a community-driven file I/O parser for molecular DFT calculations. It is an optional dependency of Atomate2, and I made a `TaskDocument` backed by cclib that can be used, as shown [here](https://github.com/materialsproject/atomate2/blob/main/src/atomate2/common/schemas/cclib.py). So, if a user wants to develop a workflow for a new code that doesn't have I/O parsing in `emmet` or in `Pymatgen`, they can use this pretty much out-of-the-box. A good example might be an ORCA calculation, for instance.

It might be nice to add this somewhere just so people know about it, but you also don't need to if you don't want to. I wanted to bring it up.

rkingsbury:
Thanks @arosen93 ! I made some updated per your suggestions, including adding language about `cclib`.

I will leave further refinements to others that have more hands-on experience with `atomate2`. Feel free to take what I have here and edit/modify/delete as you see fit!

arosen93:
@rkingsbury: Thanks!! Looks good to me! My $0.02 is that it looks good to merge and then people can add to it!

utf:
Thanks @rkingsbury, this looks great.

Issue #361, opened 2023-06-01
About zipping files at the end of a Job


gpetretto:
I have always had troubles with the fact that atomate (even v1) zips all the files in a folder. This typically ends up zipping the cluster queue files, preventing the writing from queue manager of additional informations (at least this is what happens in the clusters I have used). Other files can also end up zipped even if they are not supposed to.

I propose that 

1. the zipping can be disabled though an option in `Atomate2Settings` 
2. the preferred option would be not to zip the entire folder but just the files related to the executed code. Or maybe even just those that are expected to be large.

For VASP these lists can be used: https://www.vasp.at/wiki/index.php/Category:Output_files https://www.vasp.at/wiki/index.php/Category:Input_files. For Abinit I am still not sure if the zipping is easily manageable and we will need to see in future PR. I don't know if list of files are available for CP2K and Lobster.

If there is an interest in this I can open a PR.

JaGeo:
@gpetretto I would be certainly interested to have this option for VASP as well. It makes debugging especially hard if all files are zipped (including error files). 

Within the Lobster workflow, I think we also zip everything at the moment: https://github.com/materialsproject/atomate2/blob/c2d098a38a41c181be2c483f8adb83db0223b76d/src/atomate2/lobster/jobs.py#L90

Within atomate, we actually did it differently and only zipped all files related to Lobster.

JaGeo:
@gpetretto The typical output files for Lobster are available here: https://github.com/materialsproject/atomate2/blob/c2d098a38a41c181be2c483f8adb83db0223b76d/src/atomate2/lobster/files.py#L15

Input file is only "lobsterin" at the moment.

arosen93:
A PR for this would likely look like the following:

Add a new variable to `settings.py`:

```python
 ZIP_FILES: bool = Field(True, description="Whether generated files should be zip'd.")
 ```
 
 Then in Atomate2, whenever a function zips files, it could take an optional kwarg of `zip_files: bool = SETTINGS.ZIP_FILES` with a `from atomate2 import SETTINGS` at the top.

utf:
This was fixed in https://github.com/materialsproject/atomate2/pull/414

919607:
May I ask how to set the VASP file for compression? For example, if the WAVECAR is relatively large after calculation, I want to compress this file while not compressing the others

PR #414, opened 2023-06-30
Optionally zip files at the end of jobs

gpetretto:
## Summary

Implementation for the option to tune the zip of files at the end of (vasp) Jobs, as discussed in #361.
3 options are available: zipping all the files, zipping only code specific files or no zipping at all. 

For vasp I gathered inputs and outputs from the links in #361, and added a few more that were not mentioned there. 

## TODO

* tests
* adding the option for other codes


utf:
Thanks @gpetretto. I think this is a great change, but can you use the existing gzip functionality in `FileClient`? https://github.com/materialsproject/atomate2/blob/0955d70374b97704035917f98350bba1020cd9d9/src/atomate2/utils/file_client.py#L351

utf:
Actually, one does not typically use the file client directly but instead the helper functions in `atomate2/common/files.py`. We already have a gunzip files function in there. But could you add a gzip_files function with a consistent set of arguments: https://github.com/materialsproject/atomate2/blob/0955d70374b97704035917f98350bba1020cd9d9/src/atomate2/common/files.py#L225

gpetretto:
Thanks for pointing it out. Since it was originally using the monty gzip_dir I did not realize there was already an alternative in atomate2. Sorry. 

I tried to simply switch to `gzip_files`, but I incurred in an issue and I am not sure if this should be considered the expected behavior or a bug. I tried to pass my list of files to zip, but due to this part of the code 
https://github.com/materialsproject/atomate2/blob/0955d70374b97704035917f98350bba1020cd9d9/src/atomate2/common/files.py#L324-L326
the execution fails, as it still adds the files in the `include_files` list, even if they don't exist.
I am not sure in which case it would be useful to have the file in the list of filtered files, even when it does not exist. May I remove that `else` condition in `find_and_filter_files`?  Otherwise I will need to perform the check with glob before calling `gzip_files`.

utf:
I believe you can just set allow_missing to True?

utf:
> Otherwise I can add a test folder with a minimal set of files already unzipped.

I think this would be fine. Might be best to just `touch` the files, as their contents don't actually matter.

gpetretto:
> I think this would be fine. Might be best to just `touch` the files, as their contents don't actually matter.

I don't think I can do that while making a standard test, since the mocked execution still parses the outputs. Is there already a way to bypass the parsing? A minimal working test seems to need at least real vasprun.xml, OUTCAR and CONTCAR files.

Otherwise an overall easier way would be to move this portion of the code:
https://github.com/materialsproject/atomate2/blob/5711de0db84eb34a1b71f3f9d646c9c399857811/src/atomate2/vasp/jobs/base.py#L223-L227
to a `zip_vasp_files` function and test it on its own.

utf:
I see. I'm not sure of the best way to do it. Unless we add some additional logic to the tests for specifically checking the gzip status.

gpetretto:
I tried to move the logic to a common function and test only its behavior. 
I also added the same option for Lobster and CP2K. 
For Lobster I followed @JaGeo's suggestion, but maybe it is better if she checks this.  
For CP2K I only added `cp2k.in` and `cp2k.out` in the list of files to zip. However, since I have little knowledge of CP2K and I don't know if there are other large files that need to be zipped in the folder, I left the default of the setting `CP2K_ZIP_FILES` to True. Is there anyone that should be pinged and could provide a better list of files for CP2K?

utf:
I will merge this for now. Thanks very much @gpetretto

919607:
May I ask how to set the VASP file for compression? For example, if the WAVECAR is relatively large after calculation, I want to compress this file while not compressing the others

PR #415, opened 2023-06-30
[FIX] fix elastic tensor flow 

### LINKED ISSUE #333: BUG: Incorrect elastic tensor of FCC Si due to bug in _expand_deformations
**Describe the bug**
The elastic tensor of FCC Si calculated using ElasticMaker is incorrect with `sym_reduce=True` (default). The `C_44` is only 1/3 of the correct value.
The bug originates from incorrect exclusion of some rotated deformation in `_expand_deformations` of `atomate2.common.schemas.elastic`. It could also affect other calculations with moderate symmetry.

**To Reproduce**
The behavior can be reproduced using the following simple example.
```
from atomate2.vasp.flows.elastic import ElasticMaker
from atomate2.vasp.powerups import update_user_incar_settings
from atomate2.vasp.powerups import update_user_kpoints_settings
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
import numpy as np

# construct an FCC silicon structure
si_structure = Structure(
    lattice=[[0, 2.73, 2.73], [2.73, 0, 2.73], [2.73, 2.73, 0]],
    species=["Si", "Si"],
    coords=[[0, 0, 0], [0.25, 0.25, 0.25]],
)
si_structure = SpacegroupAnalyzer(si_structure).get_conventional_standard_structure()

# job with different sym_reduce
for sym_reduce_bool, sym_reduce_str in zip([True,False],['true','false']):
    elastic_job = ElasticMaker(bulk_relax_maker=None,sym_reduce=sym_reduce_bool).make(si_structure)
    elastic_job = update_user_incar_settings(elastic_job, {"ENCUT": 200, "ENAUG": None}, name_filter="relax")
    elastic_job = update_user_kpoints_settings(elastic_job, {"grid_density": 100}, name_filter="relax")

    responses = run_locally(elastic_job, create_folders=True)
    elastic_output = responses[elastic_job.jobs[-1].uuid][1].output

    ieee_tensor = elastic_output.elastic_tensor.ieee_format
    raw_tensor = elastic_output.elastic_tensor.raw

    np.savetxt('sym_reduce_{}_ieee.txt'.format(sym_reduce_str),ieee_tensor, fmt='%10.4f')
    np.savetxt('sym_reduce_{}_raw.txt'.format(sym_reduce_str),raw_tensor, fmt='%10.4f')
```

**Expected behavior**
The symmetrized tensor should not be too far away from raw data, and `sym_reduce` should not significantly affect the results. But using `sym_reduce = True`, the raw tensor is
```
  152.3951    62.4105    62.4105   -39.0322     0.0000     0.0000
   62.4105   152.3951    62.4105   -39.3736     0.0000     0.0000
   62.4105    62.4105   152.3951   -39.3736     0.0000     0.0000
  -39.0322   -39.3736   -39.3736    76.1099     0.0000     0.0000
    0.0000     0.0000     0.0000     0.0000     0.0000     0.0000
    0.0000     0.0000     0.0000     0.0000     0.0000     0.0000
```
while the IEEE-standard-symmetrized tensor is 
```
  152.3951    62.4105    62.4105     0.0000     0.0000    -0.0000
   62.4105   152.3951    62.4105     0.0000     0.0000    -0.0000
   62.4105    62.4105   152.3951     0.0000     0.0000    -0.0000
    0.0000     0.0000     0.0000    25.3700    -0.0000     0.0000
    0.0000     0.0000     0.0000    -0.0000    25.3700     0.0000
   -0.0000    -0.0000    -0.0000     0.0000     0.0000    25.3700
```
They are very different and this `C_44=25.37` is very far from the the experimental value `76.9`.

With `sym_reduce=False`, they are very close
raw
```
  152.3951    62.4096    62.3477    -0.0000    -0.0000    -0.0000
   62.4096   152.3932    62.3468     0.0000     0.0000    -0.0000
   62.3477    62.3468   152.2648     0.0000    -0.0000     0.0000
   -0.0000     0.0000     0.0000    76.1504     0.0000    -0.0000
   -0.0000     0.0000    -0.0000     0.0000    76.1504    -0.0000
   -0.0000    -0.0000     0.0000    -0.0000    -0.0000    76.1503
```
IEEE
```
  152.3511    62.3680    62.3680     0.0000     0.0000    -0.0000
   62.3680   152.3511    62.3680     0.0000     0.0000    -0.0000
   62.3680    62.3680   152.3511     0.0000     0.0000    -0.0000
    0.0000     0.0000     0.0000    76.1504    -0.0000     0.0000
    0.0000     0.0000     0.0000    -0.0000    76.1504     0.0000
   -0.0000    -0.0000    -0.0000     0.0000     0.0000    76.1504
```
And this `C_44=76.15` is very close to experimental value `76.9`.

**Origin of BUG**
The following lines in `_expand_deformations` of `atomate2.common.schemas.elastic` incorrectly excluded some `rotated_deformation` that should be included,
```
            if not Deformation(rotated_deformation).is_independent():
                continue
```
For example in FCC Si, the pure shear strain `epsilon_xy` or `epsilon_6`
```
 0.000  0.005  0.000
 0.005  0.000  0.000
 0.000  0.000  0.000
```
can be realized by deformation matrix
```
 0.99995  0.00000  0.00000
 0.01000  1.00000  0.00000
 0.00000  0.00000  1.00000
```
But this deformation is excluded in `_expand_deformations` when rotated from the computed `epsilon_yz` or `epsilon_4`. This caused zero C_55 and C_66 in the raw data, and therefore incorrect final IEEE-standard tensor.

**Proposed Solution**
Instead of checking whether the Deformation is "independent", we should check whether the strain is independent and not seen before. So I added this method
```
def _strain_is_independent(deformation: Deformation, tol: float=1e-8):
    # get strain in Voigt form from deformation
    strain_voigt = deformation.green_lagrange_strain.voigt
    # check how many strain components are involved
    n_perturbed = (abs(strain_voigt) > tol).astype(int).sum()
    # the strain is independent only if there is exactly one non-zero Voigt component
    return n_perturbed == 1
```
to replace the above-mentioned lines. The `mapping` also should record `Strain` instead of `Deformation` because the same strain may be realized with different deformation.

Using the modified code, both raw and symmetrized tensors are correct:
Raw
```
  152.3951    62.4105    62.4105    -0.0000    -0.0000     0.0000
   62.4105   152.3951    62.4105    -0.0000    -0.0000     0.0000
   62.4105    62.4105   152.3951    -0.0000    -0.0000     0.0000
   -0.0000    -0.0000    -0.0000    76.1504     0.0000     0.0000
   -0.0000    -0.0000    -0.0000     0.0000    76.1504    -0.0000
    0.0000     0.0000     0.0000     0.0000    -0.0000    76.1504
```
IEEE
```
  152.3951    62.4105    62.4105     0.0000     0.0000    -0.0000
   62.4105   152.3951    62.4105     0.0000     0.0000    -0.0000
   62.4105    62.4105   152.3951     0.0000     0.0000    -0.0000
    0.0000     0.0000     0.0000    76.1504    -0.0000     0.0000
    0.0000     0.0000     0.0000    -0.0000    76.1504     0.0000
   -0.0000    -0.0000    -0.0000     0.0000     0.0000    76.1504
```

If the fix looks reasonable to you and is of interest, I can make a PR or attach the modified code. But it could break the test `tests/vasp/flows/test_elastic.py ` whose C_44 seems also incorrect.

Best Regards,
Tianqi Deng
MSE&HIC
Zhejiang University

mjwen:
### Closes #333. 

The elastic tensor flow currently cannot generate the appropriate number of (24) deformation tasks to fit the tensor if  `sym_reduce` is used. The primary reason is using the deformation gradient to check independent calculations as detailed nicely in #333. This PR fixes it by using strain to check for independence.  

In a comment on #333, I mentioned that in `emmet`, we have a more complex way to deal with it. That is not needed here. This is because, in atomate2, the deformation tasks are generated based on the relaxed structures, while in atomate1, it is based on the initial structure. The atomate1 approach can have an issue when the symmetry changes after relaxation, so we add additional guards there. In atomate2, there is no such issue. 

### Fix conventional cell bug 

If `conventional` is set to `True`,  the flow only uses a conventional structure for [generating the deformations](). But the [structure](https://github.com/materialsproject/atomate2/blob/ff6070822eb329b133fa4d33ee3cd1503a9d0323/src/atomate2/vasp/flows/elastic.py#L122) used for later elastic calculation and tensor fit is still the original structure. As a result, the symmops used to reduce and expand the explicit calculations can then be different, because one is determined based on the conventional structure, and the other based on the original structure. Also, one should not use [deformations](https://github.com/materialsproject/atomate2/blob/ff6070822eb329b133fa4d33ee3cd1503a9d0323/src/atomate2/vasp/flows/elastic.py#L114C9-L114C21) from a conventional cell to strain the [original structure](https://github.com/materialsproject/atomate2/blob/ff6070822eb329b133fa4d33ee3cd1503a9d0323/src/atomate2/vasp/flows/elastic.py#L123).   

This PR fixes this by converting the structure to a conventional structure and using it in all calculations. This is also how [atomate1](https://github.com/hackingmaterials/atomate/blob/f4060e55ae3a22289fde9516ff0e8e4ac1d22190/atomate/vasp/workflows/base/elastic.py#L77) does it.   

## Changes
- Replace `deformation` with `strain` to expand the set of calculations (this fixes the bug)
- Replace `deformation` with `strain` to reduce the set of structures for explicit calculation. This can have two major benefits:
  * possibly further reduce the number of deformed structures for calculation. This is because the strain is a symmetric tensor while the deformation gradient is not.   
  * consistent with the new `_expand_strains()` function 
- Correct reference values for elastic flow test
- Add comprehensive tests for using reduced structures, conventional cells, and reducing/expanding strains


mjwen:
Just added tests for reducing the deformed structures, and for recovering from them. The idea is to test the set of strains before reducing and after recovering are exactly the same. 

Tthe test is et up, but it is only testing for a single structure now. I hope to test for all 230 space groups. There are two options to move forward: 
1. create structures on the fly
2. add a test file that consists of all needed structures 
 
I prefer the first approach, but it seems not easy to create a structure from each space group using e.g., `Structure.from_space_group()`, which requires many args. For the second approach, we can simply query from MP a prototype structure for each space. 

Any suggestions @utf and @mkhorton?

utf:
> The test is set up, but it is only testing for a single structure now. I hope to test for all 230 space groups. There are two options to move forward:

I don't actually think there are examples of all 230 space groups in the materials project. It should be sufficient to pick one example from each of the 14 Bravais lattice. I already compiled a list of example structures in amset. See here: https://github.com/hackingmaterials/amset/tree/main/tests/test_data/structures

mjwen:
I've also fixed a conventional cell bug; see the Summary above. 

The tests are failing because the `"TCH",  # flake8-type-checking` stuff will autofix and add `TYPE_CHECKING` for ElasticDocument and put [Structure](https://github.com/materialsproject/atomate2/blob/24eaac417eedf22206771d6154902b4887d9fe2a/src/atomate2/common/schemas/elastic.py#L14) and other stuff inside its if block. This does not work with pydantic. 

I was trying to fix it by adding something below to `pyproject.toml`, but did not get luck. Before digging deeper, I think @janosh should be able to help with this.  
 ```toml
[tool.ruff.per-file-ignores]
"common/schemas/*" = ["TCH"]
```

After that is fixed, it is ready for review.

**Comment by janosh** [comment]
@mjwen You could put `# ruff: noqa: TCH` at the top of the file to disable `TCH` for the whole file.

**Comment by janosh** [comment]
I think the best thing to add would be

```py
[tool.ruff.per-file-ignores]
# things inside TYPE_CHECKING aren't available at runtime and so can't be used by pydantic models
"**/schemas/*" = ["TCH"]
```

to disable moving imports into `TYPE_CHECKING` in _all_ `pydantic` schema files. That should permanently solve this problem. It didn't work for you because paths are relative to repo root, so you need to add leading `**`.

mjwen:
@janosh I know you should fix it quickly. Thanks! 

Another fix you might be able to help: Ruff does the below stuff to use newer typing. it does not work with older version of Python. Any idea to disable it?  

Screenshot 2023-07-28 at 4 21 18 PM

janosh:
@mjwen I think we want to keep those `ruff` changes. `from __future__ import annotations` is a side effect import that makes builtins like `list`, `dict`, `set`, etc. work as type hints from Python v3.7 upwards. Given the lowest Python supported by `atomate2` is 3.8, it should be safe to use. What version do you have?

mjwen:
Yes, it would be great to keep these. I am using py3.10 on my laptop and everything works nicely. 

It seems py3.8 does not work, which results in the error you see in the GH action below: "testing / test (3.8) (pull_request)". 

So, I guess we will need to either upgrade the lowest supported version to py3.9?, or use `List` instead of `list`.

janosh:
@mjwen Sorry, didn't notice that we're still talking about a schema file. In that the situation is different. `pydantic` models don't support future annotations until 3.10. So seems best if we also ignore `FA` in all schema files

```toml
# flake8-type-checking (TCH): things inside TYPE_CHECKING aren't available
#     at runtime and so can't be used by pydantic models
# flake8-future-annotations (FA): future annotations only work in pydantic models in python 3.10+
"**/schemas/*" = ["FA", "TCH"]
```

mjwen:
>  In that the situation is different. `pydantic` models don't support future annotations until 3.10. 
Good to know!

mjwen:
Thanks @janosh! Now all typing and such checks are done. 

@utf ready for review.

utf:
Fantastic, thank you.

PR #422, opened 2023-07-04
Deploy docs on every commit to `main`

### LINKED ISSUE #383: Should we update the Atomate2 docs with the version from `main`?
I didn't see the changes I made reflected in the public docs. Are you doing this after every release or after every docs update push?

janosh:
Closes #383.


janosh:
@utf Do you agree with this change?

Issue #396, opened 2023-06-23
Documentation: Defect workflow


JaGeo:
I am wondering if we could document the Defect workflow in a better way. I just wanted to check the flos and we do not have anything on this yet in the documentation.  As it is a quite complex workflow, additional documentation or a link to a paper with the description of the parts of the workflow would probably help.

mattmcdermott:
Tagging @jmmshn @nwinner

jmmshn:
@JaGeo, sorry for the late reply,  I was on a pretty length family leave.
I'll add the example and tutorial very soon.

PR #430, opened 2023-07-10
Update to Defects WF

jmmshn:
## Selective dynamics and perturbation for Defects WF

- Allow selective dynamics (clamp atoms away from defect center)
- Allow perturbation of unclamped atoms


JaGeo:
You might have seen my comment: it would also be great to add more infos on the workflows to the documentation. 😃

https://github.com/materialsproject/atomate2/issues/396#issue-1770986676

PR #431, opened 2023-07-10
More details for lobster documentation

JaGeo:
## Summary

I have added another code snippet to the documentation on how to start Lobster runs.


docs/user/codes/vasp.md
```diff
@@ -258,6 +258,33 @@ VASP_CMD: <<VASP_CMD>>
...
+lobster = VaspLobsterMaker().make(structure)
+
+# update the incar
+lobster = update_user_incar_settings(lobster, {"NPAR": 4})
+# run the job
+run_locally(lobster, create_folders=True, store=store)
+```
+
+
+It is, however,  computationally very beneficial to define two different types of job scripts for the VASP and Lobster runs.
```
janosh, line 287: Not clear to me what is meant. Is it because you have double the max wall time with 2 jobs?

JaGeo: Two types of parallelisations. MPI vs. OpenMP

JaGeo: Actually something we could also fix in custodian. Currently you cannot change number of openmp threads as part of the VASP or Lobster command.

JaGeo: @janosh tried to add the explanation. Cleaner now?

janosh: Perfect!

janosh:
Thanks @JaGeo! 👍

PR #39, opened 2021-12-21
[WIP] Initial sketch of `MagneticOrderingsMaker`

mkhorton:
## Summary

Not ready for review yet, will keep PR open as I develop.

This PR will be for a magnetic orderings workflow, ported from `MagneticOrderingsWf`:

https://github.com/hackingmaterials/atomate/blob/main/atomate/vasp/workflows/base/magnetism.py#L109

## Additional dependencies introduced (if any)

* List all new dependencies needed and justify why. While adding dependencies that bring
  significantly useful functionality is perfectly fine, adding ones that add trivial
  functionality, e.g., to use one single easily implementable function, is frowned upon.
  Provide a justification why that dependency is needed. Especially frowned upon are
  circular dependencies.

## TODO (if any)

- [ ] Write minimally-working flow
- [ ] Add test case (likely NiO, small enumeration)
- [ ] Resolve TODOs in code (likely cases where I still need to understand best practices here)

## Checklist

Work-in-progress pull requests are encouraged, but please put [WIP] in the pull request
title.

Before a pull request can be merged, the following items must be checked:

- [ ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  The easiest way to handle this is to run the following in the **correct sequence** on
  your local machine. Start with running [black](
  https://black.readthedocs.io/en/stable/index.html) on your new code. This will
  automatically reformat your code to PEP8 conventions and removes most issues. Then run
  [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/), followed by [flake8](
  http://flake8.pycqa.org/en/latest/).
- [ ] Docstrings have been added in the[Numpy docstring format](
  https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [ ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.
- [ ] Tests have been added for any new functionality or bug fixes.
- [ ] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more
efficient if you already fix most errors prior to submitting the PR. It is highly
recommended that you use the pre-commit hook provided in the repository. Simply run
`pre-commit install` and a check will be run prior to allowing commits.

Issue #145, opened 2022-06-16
FEATURE: Converge on (part) of the TaskDocument


davidwaroquiers:
In order to be able to use "code-agnostic-like" workflows, we should try to converge on some parts of the TaskDocument's. One obvious example is the finite-difference elastic constants workflow where the deformations could be performed using "any" code, provided the response follows some convention, i.e. it should have the energy, the forces, stress, ...
This needs some discussions with @utf on how to deal with that. We might consider having some general pydantic models that are subclassed for the different codes (somewhat similar to StructureMetadata).

PR #432, opened 2023-07-11
Add magnetic orderings workflow

mattmcdermott:
## Summary

This PR includes the addition of a flow and builder for collinear magnetic ordering enumeration/calculations. This flow is based on the original [`MagneticOrderingsWF` in atomate1](https://github.com/hackingmaterials/atomate/blob/387938da7c578481c41aa4cfeb65a166778e6372/atomate/vasp/workflows/base/magnetism.py#L108C27-L108C27). This extends the first sketch by @mkhorton in #39.

The final postprocessing step is included as a job and as a builder.

Most of the workflow logic is written within `common`.  In this PR I have implemented everything with VASP.

## TODO (if any)

- [x] Finalize and merge PR into pymatgen for broken magnetic code https://github.com/materialsproject/pymatgen/pull/3070
- [x] Fix possible bugs related to orderings changing between static and relax
- [x] Explore moving even more code into common (especially builder code)
- [x] Add tests

## Checklist

  [ruff](https://ruff.rs).
  type check your code.


mkhorton:
Amazing! I will review but don't expect major comments.

mattmcdermott:
Thanks! Yes feedback especially on schemas would be useful -- I did base them on your previous sketches though.

mattmcdermott:
FYI: I moved the builder into `common`. Of course, this opens up the possibility of the builder mixing outputs from different DFT codes; I addressed this by adding a property that is implemented in child classes. This property, `_dft_code_query`, is tacked onto the `get_items` query to ensure that only calcs from one DFT code are used.

Since this is the first builder in `common`, maybe we can standardize an approach like this for future common builders?

JaGeo:
Approach for `common` sounds very good!

src/atomate2/common/builders/magnetism.py
```diff
@@ -0,0 +1,248 @@
...
+    grouped_tasks = [[tasks[0]]]
+
+    for task in tasks[1:]:
+        parent_structure = task["metadata"]["parent_structure"]
+
+        match = False
+        for group in grouped_tasks:
+            group_parent_structure = task["metadata"]["parent_structure"]
+
+            # parent structure should really be exactly identical (from same workflow)
```
mkhorton, line 229: I like this function. I wonder if we might benefit from the same code but in an `exact_match()` utility function somewhere instead, it seems like it might be useful more generally.

src/atomate2/common/schemas/magnetism.py
```diff
@@ -0,0 +1,207 @@
...
+        None, description="Final structure from the calculation."
+    )
+    ordering: Ordering = Field(
+        None,
+        description=(
+            "The magnetic ordering of the output structure, "
+            "as defined in pymatgen.analysis.magnetism.analyzer."
+        ),
+    )
+    magmoms: list[float] = Field(None, description="Magnetic moments of the structure.")
```
mkhorton, line 86: Not a request to modify, but adding a note for reference: it would be useful to round magnetic moments somewhat, since getting a primitive structure (with respect to the magnetic moments, i.e. without losing magnetic order) will not work as intended unless equivalent moments are exactly equal. Typically, the noise in the moment is sufficient such that this is not the case. I did not get around to adding this however.

src/atomate2/common/schemas/magnetism.py
```diff
@@ -0,0 +1,207 @@
...
+            "Specifies whether or not the symmetry changed during the calculation."
+        ),
+    )
+    energy_above_ground_state_per_atom: float = Field(
+        None, description="Energy per atom above the calculated ground state ordering."
+    )
+    relax_output: MagneticOrderingRelaxation | None = Field(
+        None, description="Relaxation output, if relaxation performed."
+    )
+    energy_diff_relax_static: str | float | None = Field(
```
mkhorton, line 122: Might be worth noting this field was only really present for the purpose of benchmarking (i.e., is the static calculation required).

Since I wrote the original workflow, I have since concluded that the static step should--at minimum--have a much smaller EDIFF, however I have not benchmarked this. The original benchmark results speak for themselves, however the difference in energies between orderings is sometimes very very small, such that I would have more confidence if we had a tighter tolerance.

src/atomate2/vasp/builders/magnetism.py
```diff
@@ -0,0 +1,23 @@
...
+from atomate2.common.builders.magnetism import (
+    MagneticOrderingsBuilder as MagneticOrderingsBuilderBase,
+)
+from atomate2.vasp.schemas.magnetism import (
+    MagneticOrderingOutput,
+    MagneticOrderingRelaxation,
+)
+
+
+class MagneticOrderingsBuilder(MagneticOrderingsBuilderBase):
```
mkhorton, line 14: It's really nice to see how few additions were required to make this work with VASP. I wonder if this is a pattern that we should use elsewhere in atomate2? (If it's not already?)

src/atomate2/vasp/flows/magnetism.py
```diff
@@ -0,0 +1,100 @@
...
+    name : str
+        Name of the flows produced by this Maker.
+    static_maker : StaticMaker
+        Maker used to peform static calculations for total energy.
+    relax_maker : RelaxMaker | None
+        Maker used to perform relaxations of the enumerated structures. If None,
+        relaxations will be skipped (i.e., only static calculations).
+    default_magmoms : dict | None
+        Optional default mapping of magnetic elements to their initial magnetic moments
+        in µB. Generally these are chosen to be high-spin, since they can relax to a
```
mkhorton, line 58: One thing that might be useful to verify is the default spin on Co. Since the original workflow was written, the pymatgen default changed to low-spin Co, since unlike other elements, Co seemed happier to be trapped in a high-spin configuration even if less stable. Strictly, I think Co should probably be run twice with both configurations.

mkhorton (approved):
Great PR! Approved by me, although noting I'm not the maintainer here :) I added some comments mostly for context/gotchas.

mattmcdermott:
Btw I haven't forgotten about this; just a little preoccupied. Need to address a bug still and write tests!

src/atomate2/common/schemas/magnetism.py
mattmcdermott: Should we enable `round_magmoms=True` in the `CollinearMagneticStructureAnalyzer`?

mattmcdermott:
Bug was addressed. See Issue #505 where AFM orderings were being destroyed between relax and static calcs. Fix was provided in PR #506.

Still need to write tests and then we should be about ready to merge!

mattmcdermott:
@utf I just implemented your suggestion in #466. This workflow now contains a postprocessing job that shares much of the doc-constructing code used in the builder.

In addition to feedback about the postprocessing implementation, your feedback on how this is implemented within `common` would be very helpful, as this workflow might set the precedence for future additions to atomate2.

I'll be working on tests in the meantime and will let you know when they're ready :)

mattmcdermott:
Test for magnetic ordering workflow has been added (using the developer guide).  Since writing this code, @mcgalcode and I have also run the workflow and builder on 50+ structures, and it seems stable overall. 

@utf Is there anything else I should add? It would be great if you could review Matt H's comments above as well; one discussion point is whether we should make the static jobs have a smaller EDIFF by default (e.g., 1e-6 or 1e-7). Instead, we could leave it the way it is (i.e., default `StaticMaker`) and let the user change this if desired.

I see now that the new test is failing on GitHub Actions because `enumlib` is a required dependency to enumerate magnetic orderings. What do you think we should do here?

src/atomate2/vasp/builders/magnetism.py
mattmcdermott: I think it's a nice pattern, although it was tricky to figure out. I hope that this addition will be a nice example for future workflows :)

src/atomate2/common/schemas/magnetism.py
mattmcdermott: This is an important (and still unresolved point). Right now, the default static calculations in the workflow are made with the VASP `StaticMaker`, which I believe is `EDIFF=1e-5`.

mkhorton:
> have also run the workflow and builder on 50+ structures

This seems more than sufficient to approve. Would recommend leaving the EDIFF to the default unless benchmarking suggests a tighter tolerance is required (although I agree it might help).

Re. `enumlib`, it's an easy compile but it's not consistently versioned. The one on conda-forge does seem like it was last updated 9 months ago, so fairly recent--perhaps we could try installing this in CI?

src/atomate2/common/schemas/magnetism.py
mkhorton: Yes, I'd do that. Eventually the underlying algorithm in `round_magmoms` should be updated however. Simple rounding (e.g. `np.round`) will sometimes round two similar magmoms in opposite directions. I tried instead doing a KDE of the magmoms with a suitable Gaussian width and taking the peak positions, but this felt like it was over-complicating things!

utf:
@Zhuoying please can you take a look at this PR and the comments.

@mattmcdermott once @Zhuoying has had a look over, I will do a final pass.

src/atomate2/common/schemas/magnetism.py
Zhuoying: The default `StaticMaker` (EDIFF=1e-5) is looser than `TightRelaxMaker` (ENCUT=700, EDIFFG=1e-7). Maybe we could use that parameter and this `energy_diff_relax_static` term to benchmark and see if we need to improve the EDIFF in `StaticMaker`.

src/atomate2/vasp/flows/magnetism.py
```diff
@@ -0,0 +1,109 @@
+"""VASP-specific flows for magnetism-related calculations."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Literal, Sequence
+
+from atomate2.common.flows.magnetism import (
+    MagneticOrderingsMaker as MagneticOrderingsMakerBase,
```
Zhuoying, line 9: DFT-independent `BaseMaker` are put in `common` and callable under `vasp/flows`. A good example to follow for future workflow writing.

Zhuoying (commented):
Nice structured workflow. Can be used as a good example for new developers when they need a standardized postprocessing. The remaining issue might be:
(i) how strict some incar in static calculations
(ii) double-check Co default spin (changed to low-spin in pymatgen) 
(iii) Pydantic v2 update (if any)
Thanks @mattmcdermott for the nice PR.

src/atomate2/vasp/builders/magnetism.py
mattmcdermott: Thanks @Zhuoying, agreed :)

JaGeo: This would be great! I think we could then easily connect the forcefields as well 😃

mattmcdermott:
Hi @Zhuoying, thanks for the nice feedback and code review; this is very helpful!

> The remaining issue might be: (i) Is the default incar settings (EDIFF=1e-5) in StaticMaker sufficient (may need some benchmark evidence). 

I do think EDIFF for the static calc should be stricter by default, per @mkhorton's recommendation. So far, I have run relax+static calcs on 285 different orderings (from 26 structures) with the current defaults (EDIFF=1e-5).  I made a histogram of all their (normalized) `energy_diff_relax_static` values. 

Screenshot 2023-10-18 at 1 15 04 PM

The mean difference is -0.0016 eV/atom (i.e. the static calculation tends to be that much lower in energy than the relax), but there are some outliers! I don't have a good sense on what this means with regards to picking an EDIFF, but I imagine this supports the idea that we should be stricter... Thoughts?

> (ii) Double-check Co default spin (default changed to low-spin in pymatgen) 

While the default Co is now low-spin, the `MagneticStructureEnumerator` still uses the older default magnetic moments by loading them from file here: 

https://github.com/materialsproject/pymatgen/blob/cfd47add8958af795dabcfe1661cf8890ad56415/pymatgen/analysis/magnetism/default_magmoms.yaml#L1-L4

This means that right now the default for the magnetic orderings workflow will be to initialize Co to high-spin. @mkhorton recommended running both low-spin and high-spin. Is this something we should implement in pymatgen or implement here?

> (iii) Pydantic v2 update (if any) 

Will look into!

JaGeo:
> I do think EDIFF for the static calc should be stricter by default, per @mkhorton's recommendation. So far, I have run relax+static calcs on 285 different orderings (from 26 structures) with the current defaults (EDIFF=1e-5). I made a histogram of all their (normalized) `energy_diff_relax_static` values.
> 
> Screenshot 2023-10-18 at 1 15 04 PM
> 
> The mean difference is -0.0016 eV/atom (i.e. the static calculation tends to be that much lower in energy than the relax), but there are some outliers! I don't have a good sense on what this means with regards to picking an EDIFF, but I imagine this supports the idea that we should be stricter... Thoughts?

`1e-7` or at least `1e-6`

src/atomate2/common/flows/magnetism.py
```diff
@@ -0,0 +1,171 @@
...
+    @property
+    @abstractmethod
+    def prev_calc_dir_argname(self):
+        """Name of argument informing static maker of previous calculation directory.
+
+        As this differs between different DFT codes (e.g., VASP, CP2K), it
+        has been left as a property to be implemented by the inheriting class.
+
+        Note: this is only applicable if a relax_maker is specified; i.e., two
+        calculations are performed for each ordering (relax -> static)
```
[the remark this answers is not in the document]
mattmcdermott: This is correct; only a static calculation will be performed by default if `relax_maker=None`.

The `prev_calc_dir_argname` is simply to keep track of the name of the argument referencing a previous directory in your DFT code's maker. This is basically a patch/workaround while we have differently named args (e.g., `prev_vasp_dir`, `prev_cp2k_dir`, etc.). Does that answer your question? Not sure if I understood

src/atomate2/common/flows/magnetism.py
```diff
@@ -0,0 +1,171 @@
...
+    default_magmoms : dict | None
+        Optional default mapping of magnetic elements to their initial magnetic moments
+        in µB. Generally these are chosen to be high-spin, since they can relax to a
+        low-spin configuration during a DFT electronic configuration. If None, will use
+        the default values provided in pymatgen/analysis/magnetism/default_magmoms.yaml.
+    strategies : tuple[str]
+        Different ordering strategies to use. Choose from ferromagnetic,
+        antiferromagnetic, antiferromagnetic_by_motif, ferrimagnetic_by_motif,
+        ferrimagnetic_by_species, or nonmagnetic. Here, "motif", means to use a
+        different ordering parameter for symmetry inequivalent sites.
```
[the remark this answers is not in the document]
mattmcdermott: @mkhorton can you shed some light on this? Not very familiar with this terminology as it is primarily handled in pymatgen. From looking at the `MagneticStructureEnumerator` code, it seems that `ferrimagnetic_by_motif` only is added as a strategy when there is more than one type of wyckoff symbol for the magnetic site, otherwise `ferrimagnetic_by_species` would be the only strategy. Not sure about how results differ

JaGeo:
Just to shortly add to my last comment, @mattmcdermott : I think one would need to compare the energies resulting from two static runs. comparing the optimization with a static run could lead to wrong conclusions as the volume might change during the optimization which has an influence on the basis set quality.

Zhuoying:
@mattmcdermott Thanks for providing the benchmarking histogram! It gives some quantitative evidence on whether we should choose a more strict EDIFF. 

Two follow-up comments: 
(i) EDIFF in `static_maker`: From the histogram, there are more than 10% of orderings (~30/285) Ediff >10 meV/atom. **If the difference in energies between orderings is sometimes very very small (<5meV/atom)**, the current default (`1e-5`) needs to be modified to a smaller value (i.e., 1e-7) rather than leaving it for users to decide IMO.

(ii) `relax_maker=None`: I suppose when you have confidence in your input structure (well-relaxed), you can turn the `relax_maker` off. I have not come up with a situation where you can do this. -If magnetism for different orderings is set well and done with relaxation, it sounds like manually doing the first half of the workflow. If not and skip the relaxation step, I might be worried about the accuracy since magnetism does matter for the optimized structures. So if you have any idea when we can turn it off, please let me know. Otherwise, we might consider adding a warning for `relax_maker=None` and turning it on as the default?

@utf The discussions and replies to my review are done. Now it is ready for your final review.

mkhorton:
> This means that right now the default for the magnetic orderings workflow will be to initialize Co to high-spin. @mkhorton recommended running both low-spin and high-spin. Is this something we should implement in pymatgen or implement here?

@mattmcdermott This is a personal judgement call. Co has historically been tricky, and seems more likely to get stuck in a high-spin configuration than other elements. The decision to switch to a low-spin by default was simply because it seemed more common (i.e., what would be the most sensible _default_ value?) but of course plenty of high-spin Co materials exist, and this is all anecdotal. The most exhaustive option would be to consider multiple states for all elements, but this would likely be quite inefficient.

For the purposes of this workflow, if you can predict the oxidation state ahead of time, that's helpful, and the default magnetic moment can be set differently according to that.

I would not recommend any changes to this workflow at this time, there isn't enough evidence either way (that I know of). At the end of the day, porting this workflow to atomate2 at all is super helpful, and it doesn't mean further improvements can't be made later.

Modifying the workflow/enumerator to allow multiple spin states per element would be an option, but I think this might be an awkward change. Perhaps the best option is simply giving the user a warning if they're calculating a material including Co and suggest they attempt both?

src/atomate2/common/flows/magnetism.py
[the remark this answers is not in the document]
mkhorton: My own memory is dim at this point, but I think it's like this:

* Ferrimagnetic orderings require defining at least two distinct sub-lattices (I realize I'm using this term imprecisely), where the spin orientation on the second sub-lattice is opposite to that on the first.
  * `ferrimagnetic_by_species` defines the sub-lattices by saying, e.g., one element is the one sub-lattice and the other element is another sub-lattice. This can work well for materials where there are multiple magnetic species.
  * `ferrimagnetic_by_motif` defines the sub-lattices by taking into account the actual site symmetries, for example consider Fe3O4: in this material there is only a single magnetic element, but there are two distinct sites, an octahedral and a tetrahedral, and the ferrimagnetic order is defined with opposing spins in those two sites.

The default, again, is simply based on how exhaustive you want to be, since the more enumerations the more expensive the calculation. I think the default was set to make it more appropriate for high-throughput.

mattmcdermott:
1. @Zhuoying and @JaGeo Thank you for the helpful discussion on EDIFF values. I agree with your assessment and will implement `EDIFF=1e-7` as the new default.
2. @Zhuoying Regarding the optional relaxation -- note that the VASP-implemented workflow includes a RelaxMaker by default (it only defaults to None in the common flow). I will add a warning when None is supplied, though.
3. @mkhorton Providing the warning you described for Co-containing structures seems like the best solution for now. 

Thanks all for your help. I will update the PR today, and then we should be good for final review by @utf

src/atomate2/common/flows/magnetism.py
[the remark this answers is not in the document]
mattmcdermott: Fantastic response! Thank you for clearing that up

src/atomate2/vasp/flows/magnetism.py
mattmcdermott: I think we're all good here!

fraricci:
If I can chime in here. I just posted a bug issue in pmg which is affecting the atomate MagWF. https://github.com/materialsproject/pymatgen/issues/3431

It could be useful to test if that bug affects the current MagWF in atomate2 as well.

utf:
This looks fantastic. Thanks so much @mattmcdermott. Once tests pass I'm happy to merge.

mattmcdermott:
Thanks for the nice feedback @utf. All the tests are passing :) 

The failure is a CI issue only (can't reproduce on my laptop). It seems related to calculating coverage. I'll figure this out today so we can merge the PR.

```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/_pytest/main.py", line 271, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/_pytest/main.py", line 325, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/pluggy/_hooks.py", line 493, in __call__
INTERNALERROR>     return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/pluggy/_manager.py", line 115, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/pluggy/_callers.py", line 130, in _multicall
INTERNALERROR>     teardown[0].send(outcome)
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/pytest_cov/plugin.py", line 298, in pytest_runtestloop
INTERNALERROR>     self.cov_controller.finish()
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/pytest_cov/engine.py", line 44, in ensure_topdir_wrapper
INTERNALERROR>     return meth(self, *args, **kwargs)
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/pytest_cov/engine.py", line 254, in finish
INTERNALERROR>     self.cov.combine()
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/coverage/control.py", line 832, in combine
INTERNALERROR>     combine_parallel_data(
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/coverage/data.py", line 176, in combine_parallel_data
INTERNALERROR>     data.update(new_data, aliases=aliases)
INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/coverage/sqldata.py", line 6[60](https://github.com/materialsproject/atomate2/actions/runs/6884676495/job/18727615337?pr=432#step:7:61), in update
INTERNALERROR>     raise DataError("Can't combine line data with arc data")
INTERNALERROR> coverage.exceptions.DataError: Can't combine line data with arc data
```

Zhuoying:
It is glad that @utf gave a final approval. 
@mattmcdermott Thanks for all the hard work to make this PR look amazing! I just updated the branch and retriggered the tests, let's see if the CI issue still exists or not.

mattmcdermott:
Thanks @Zhuoying. Now getting some failures on the tests for forcefields flows?

Zhuoying:
@mattmcdermott Yeah, the failures seem not related with this PR.
Let me ping @JaGeo, do you have any thoughts on it?
Screenshot 2023-11-16 at 12 13 27 PM

JaGeo:
@janosh merged something new related to forcefields yesterday (https://github.com/materialsproject/atomate2/pull/611) where there were issues with other forcefield outputs. I fear it might be related.

janosh:
I suspect (could be wrong) `matgl` is the culprit here. I just tried and failed to publish a new release (which may or may not fix this) over at https://github.com/materialsvirtuallab/matgl/releases/tag/v0.8.6. It's also affecting @esoteric-ephemera in https://github.com/materialsproject/emmet/pull/892 with a clearer error message (assuming it's the same issue here). 

> FAILED emmet-core/tests/test_ml.py::test_ml_doc[M3GNet-MP-2021.2.8-PES-prop_kwargs1] - ValueError: Bad serialized model or bad model name. It is possible that you have an older model cached. Please clear your cache by running `python -c "import matgl; matgl.clear_cache()"`

@shyuep Can someone in your group take a look?

mattmcdermott:
@Zhuoying @utf tests are passing now. Codecov check is warning about the diff but seems to only apply to some lines in the builder (as far as I know, we do not have any tests for builders in atomate2).

src/atomate2/vasp/schemas/magnetism.py
```diff
@@ -0,0 +1,105 @@
...
+            input_structure=input_structure,
+            output_structure=structure,
+            output_energy=energy,
+            uuid=uuid,
+            dir_name=dir_name,
+            relax_output=relax_output,
+        )
+
+
+class MagneticOrderingsDocument(MagneticOrderingsDocumentBase):
```
utf, line 90: Thinking about this a bit more, I'm personally not too keen on these subclasses for the schemas. While the implementation is clean, it ultimately adds a bit more boilerplate to implement if you want to configure the magnetism workflow for another code. The other issue is that if you need the full task document to construct the schemas, this will pull a lot of information from the database adding unnecessary strain.

Ideally, the inputs needed for the schemas would be configured in the workflow itself, e.g., see the code for constructing the elastic schemas here: https://github.com/materialsproject/atomate2/blob/main/src/atomate2/common/jobs/elastic.py#L167

and here: https://github.com/materialsproject/atomate2/blob/9066774f4615f48d091ca80dcbbefa3ba1527f9a/src/atomate2/common/flows/elastic.py#L136-L144

Only the specific information needed to construct the documents are passed and there is no VASP specific code.

Is it possible to do something similar in this workflow? That would enable us to remove the VASP specific schemas, and the also the extra class function for building the schemas.



Only the specific information

utf:
Thanks @mattmcdermott, I just had one thought about the schemas and VASP specific code needed.

src/atomate2/vasp/schemas/magnetism.py
```diff
@@ -0,0 +1,105 @@
...
+        This does not include the uuid, which must be set separately.
+        """
+        dir_name = task_document.dir_name
+        structure = task_document.structure
+        input_structure = task_document.input.structure
+        if not input_structure.site_properties.get("magmom"):
+            input_structure.add_site_property(  # input struct likely has no magmoms
+                "magmom", task_document.input.parameters["MAGMOM"]
+            )
+        energy = task_document.output.energy
```
mattmcdermott, line 78: @utf These lines prevent the simpler implementation as you suggest. If magnetic moments were stored in the task document in a universal way (independent of the DFT code), then we would not need to do this, and we could remove the VASP schema subclasses. 

Do you know if there is a code-independent way to access magnetic moments from task documents? They're not always stored in the site properties of the pymatgen structure (hence the logic of the highlighted line).

mattmcdermott: (sorry the highlighted lines should be 74-77)

utf: Thanks for the clarification. This is a good point and I'm not sure of the best way forward.

Does it make sense to store the magmoms on the input structure in the task doc? Naively I think it does but perhaps @munrojm or @tschaume can spot some issues.

Alternatively, I propose that we add a new field to the schema, let's say `task_document.input.magnetic_moments` and then make this standard for VASP, CP2K, AIMs etc. 

I think this extra bit of effort will make it much nicer for generalisable workflows longer term.

utf: Hi @mattmcdermott, have you had a chance to work on the above at all?

mattmcdermott: @utf No, sorry. It wasn't clear to me which change we should go with and whether I should make the changes in emmet. Are you still wanting to go with the new input field approach?

utf: I think the easiest approach for now is to add the new input field. Would you be happy to implement that?

mattmcdermott: @utf Looking into the emmet code further.... if we added a magnetic moments property to the input schema, would this actually be enough? I thought so at first but there is more code here that is VASP specific.

To completely get rid of a DFT code-specific `MagneticOrderingRelaxation.from_task_document` implementation, we would need a universal `TaskDoc` base class that all DFT codes share. As far as I can tell, we don't have this yet. It seems that most task doc implementations for other DFT codes are subclassing `StructureMetadata`, which doesn't necessitate that there is a specific `InputDoc` or `OutputDoc` format, and we also need to ensure that those are in there and follow the same type of structure. Can you confirm that my understanding of this is correct?

If the goal is to write a universal base TaskDoc that all the different DFT-codes share, then I would support that... but it seems like a massive change, and is outside the scope of a single atomate2 feature contribution IMO. (This would likely be something discussed among the full team of developers and implemented as a followup PR right?)

mattmcdermott: Looping in @janosh since he has a better understanding than me of how the subclassing from `emmet-core` works

utf: Strictly yes, we should create a universal base TaskDoc, however, I do not expect you do to that as part of this PR. Currently, everything is held together in an ad-hoc way. We don't have a base TaskDoc and yet we do have universal workflows (such as the elastic tensor) due to common fields in the existing task docs. In order to get this merged, I'd be happy with the following:
- Add the requested field for the VASP task doc.
- Raise an issue on atomate2 noting that this field should be added for CP2K/aims.

utf: Actually, one is already open from 2022! https://github.com/materialsproject/atomate2/issues/145

mattmcdermott: Oh wow, haha. It agree it should be a priority, and ideally, we can get input from many others to ensure that it does not restrict the addition of support for future codes.

> Add the requested field for the VASP task doc.

And to confirm, once I add the requested field in a PR in `emmet-core`, I can still leave the current VASP-specific implementation for this workflow as is? I think it would have to remain that way unless you want to get rid of the typing (since there is no base TaskDoc we can use to set the type). I guess that would be fine but also seems a little sketchy...

utf: Personally, I think it could be better to remove the VASP implementation but add a note in the flow that this workflow currently only works for Task Documents that have this particular field. 

We can remove this note once the universal base is adopted.

mattmcdermott: Alright, got it. I think that will be okay and I'll add a warning.

I'll link to the PR in emmet once it's done.

mattmcdermott: See https://github.com/materialsproject/emmet/pull/945

mattmcdermott: Implemented this change!

mattmcdermott:
@utf this is now implemented only in common, with defaults for VASP and a note about ensuring compatible `TaskDoc` implementation for alternate DFT codes.

Note I bumped the `emmet-core` version to include the TaskDoc change I made there (https://github.com/materialsproject/emmet/pull/945) 

The failing tests are from other modules.

No rush to pull but let me know if you are envisioning any other changes!

utf:
This is fantastic. Thanks very much!

The tests are failing because of a recent change in emmet. I can merge this once materialsproject/emmet#974 is merged and a new version released.

munrojm:
Just merged, it should be released soon.

JaGeo:
@utf I am merging as now all tests pass and you said it is good to merge as soon as the emmet version is fixed (only codecov slightly decreased now)

PR #460, opened 2023-08-02
[BUG FIX] Edge case for Magmoms

jmmshn:
There's a strange interaction between charge-decorated structures and magmoms. 

Charge-decorated structures can `site.species.spin == None` which would lead to a result of `[None, ....]` for the `_get_magmoms` function and then lead to failure when writing the INCAR


jmmshn:
@utf, @janosh, this was causing failures with structures grabbed with oxidation states.

PR #475, opened 2023-08-14
Use GitHub's `deploy-pages` action to deploy docs

### LINKED ISSUE #454: Migrate to 1st party GH Pages Deploy action
GitHub added a [1st party Pages deploy action](https://github.com/actions/deploy-pages) a while back. See [e.g. here for usage example](https://github.com/janosh/workflows/blob/becb92eff4b140d78d094e45dc29684ef37861bf/.github/workflows/nodejs-gh-pages.yml).

We're currently using `peaceiris/actions-gh-pages`

https://github.com/materialsproject/atomate2/blob/59fd0b46b434d563b1621ee037fc8c4a55e6618e/.github/workflows/deploy.yml#L38-L42

but may want to consider migrating.

janosh:
Closes #454.

## TODO (if any)

Needs a test run to make sure it works.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,8 +5,15 @@ on:
   push:
     branches: [main]
 
+# set GITHUB_TOKEN permissions to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
 jobs:
   build-docs:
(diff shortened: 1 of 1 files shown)


utf:
Thanks! Let's give it a whirl.

utf:
Great, thanks @janosh. I didn't have to change the branch protections. It seems like everything is working now.

PR #476, opened 2023-08-15
Small change to `gunzip` to allow better restarting

jmmshn:
# Small change to `gunzip` to allow better restarting

Proposed change to make resuming of long-running calculations easier

If we want to resume a calculation
`VaspJob` in `custodian` has an `auto_continue` argument.  This makes sure that `CONTCAR` is copied to `POSCAR` whenever possible.
https://github.com/materialsproject/custodian/blame/26bf28a4b15b1253c955f3f8d6bd427bbc398e3a/custodian/vasp/jobs.py#L209-L229

So setting something like this at job creation or setting it in a restart script/CLI will be useful.
For a given maker:
```python 
BaseVaspMaker.run_vasp_kwargs = {"vasp_job_kwargs" : {"auto_continue" : True}}
```
However, most of the  VASP jobs in `atomate2` will copy from `prev_dir` first and thus overwrite the current state of the directory.  In most cases, this will just error out because `force_overwrite=False` by default.
For proper restarting: the `copy_vasp_files` function should make sure it does not overwrite existing files.
This is currently not possible.  This PR changes the `gunzip` function to accept `force="skip"` as an argument.

### DIFF
--- a/src/atomate2/utils/file_client.py
+++ b/src/atomate2/utils/file_client.py
@@ -369,7 +369,7 @@ class FileClient:
         force : bool
             How to handle writing a gzipped file if it already exists. Accepts
             either a string or bool:
-            
+
             - `"force"` or `True`: Overwrite gzipped file if it already exists.
             - `"raise"` or `False`: Raise an error if file already exists.
             - `"skip"` Skip file if it already exists.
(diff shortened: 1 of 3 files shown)


utf:
I just made some edits to add the same feature to gzip files as well, to make sure the APIs are consistent. Would you be able to add a corresponding test for the force mode for gzipping files?

I also butchered the linting, sorry about that.

utf:
Actually, nvm. This looks good!

Issue #441, opened 2023-07-19
BUG: auto_ismear is not overriden by user_incar_settings


esoteric-ephemera:
Two primary issues here:
- The auto_ismear tag in the VaspInputSetGenerator class is not overriden by the user's choice of ISMEAR in the user_incar_settings. Since auto_ismear defaults to True, the user's choice of ISMEAR / SIGMA will always be overwritten by the defaults in VaspInputSetGenerator
- Also, the default smearing width for Gaussian smearing (ISMEAR = 0) set for (auto_ismear = True) is very large compared to what is usually recommended (SIGMA = 0.2 vs. 0.03 - 0.05)

matthewkuner:
@esoteric-ephemera this is related to this: #412

PR #487, opened 2023-08-22
fix auto ismear 0 sigma value

matthewkuner:
`auto_ismear`, when no previous calc's bandgap is available, sets ISMEAR = 0 (already being done, all is well there). According to the VASP wiki, a SIGMA of 0.03-0.05 should be used in such cases (not being done).

This changes the SIGMA in such cases from 0.2 to 0.05.

### DIFF
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -668,7 +668,7 @@ class VaspInputGenerator(InputGenerator):
             if bandgap is None:
                 # don't know if we are a metal or insulator so set ISMEAR and SIGMA to
                 # be safe with the most general settings
-                auto_updates.update({"ISMEAR": 0, "SIGMA": 0.2})
+                auto_updates.update({"ISMEAR": 0, "SIGMA": 0.05})
             elif bandgap == 0:
                 auto_updates.update({"ISMEAR": 2, "SIGMA": 0.2})  # metal
             else:


utf:
If no band gap is available, I think we assume the system is a metal and therefore set a large sigma.

matthewkuner:
@utf as it says in the VASP manual, if you have no prior knowledge of a system, it would be best to set SIGMA = 0.03-0.05 (https://www.vasp.at/wiki/index.php/ISMEAR). I think setting it to 0.2 would make a calculation that is supposed to be good for all materials no longer serve that primary function. I believe that @esoteric-amoeba would agree based on his comments in #441

esoteric-ephemera:
Agreed with @matthewkuner here - the large sigma = 0.2 is OK for Methfessel-Paxton smearing (ISMEAR = 1 or 2), but isn't the best choice for Gaussian smearing (ISMEAR = 0). Would be easy to run some tests showing this, @utf let me know if that's of interest to have data for

utf:
My hunch is that SIGMA = 0.2 with ISMEAR = 0 will give reasonable results for both metals and any semiconductor with a gap above ~0.5 eV.

But if you're willing to run some tests @esoteric-ephemera that would be fantastic. It would be good to include a mix of metals, non-metals and some small gap materials too.

matthewkuner:
@utf but what about semiconductors with a bandgap below 0.5 eV? Especially given how DFT under-estimates bandgaps, I expect this to be a decent fraction of materials

utf:
Agreed. That is why having some benchmarking would be useful. The question is which is worse:
- Running metals with SIGMA = 0.05
- Running small gap materials with SIGMA = 0.2

matthewkuner:
What do you mean by “worse”? A smaller SIGMA can only make the results better for all materials (though it may make convergence somewhat slower)

utf:
>  A smaller SIGMA can only make the results better for all materials (though it may make convergence somewhat slower)

That's not quite right. A small sigma means you need denser k-points sampling for metals. If you use a small sigma with metals you can easily get the wrong energy and forces because the Fermi level will not be set to the correct position.

Quantifying the smearing methods against: 1. Static total energy calculations with ISMEAR = -5, and 2. Relaxed lattice parameters from ISMEAR = -5 (semiconductors) and MP smearing (metals) would be sufficient.

utf:
I'm going to close this, since this will come into effect when we switch to the pymatgen input sets.

PR #504, opened 2023-09-01
Materials Project GGA and meta-GGA workflows

### LINKED ISSUE #352: Port SCAN workflows to atomate2

### LINKED ISSUE #363: BUG: `LMAXTAU` should be auto-set when there are f electrons and `LASPH` is True for meta-GGA calculations

### LINKED ISSUE #364: BUG: `LMAXMIX` should be set based on the presence of s, p, d, f electrons not based on Z

janosh:
This PR branches off from #362. It implements the future Materials Project PBE and r2SCAN relaxations as 3-step workflows consisting of a double-relaxation followed by a higher-quality static job (all as in #362 so far) but builds these 3 steps out of `MP(Meta)GGADoubleRelaxMakers` that are exported for standalone use, as well as called inside `MP(Meta)GGADoubleRelaxStatic` (the main workflows intended for public use).

The difference to #362 is intended to address https://github.com/materialsproject/atomate2/pull/362#discussion_r1298853478 and more closely align the MP workflow API with the existing `DoubleRelaxMaker` in `atomate2.vasp.flows.core`.

---

Closes #352. Closes #363. Closes #364.

- #352
- #363
- #364

This has been a team effort with @esoteric-ephemera and @Andrew-S-Rosen. 🙏


esoteric-ephemera:
Looks good! Can we add back in the GGA_plus_U option for MPGGADoubleRelaxStatic? There's an argument in the maker for it, but the logic like we had before to alter the config_dict's of the makers isn't

src/atomate2/vasp/sets/mp.py
```diff
@@ -44,6 +43,45 @@ class MPGGAStaticSetGenerator(StaticSetGenerator):
...
+            A vasprun from a previous calculation.
+        outcar
+            An outcar from a previous calculation.
+
+        Returns
+        -------
+        dict
+            A dictionary of updates to apply.
+        """
+        return {"ALGO": "FAST"}
```
janosh, line 82: @esoteric-ephemera @arosen93 Do you remember any reason why we don't set this in the `_BASE_MP_R2SCAN_RELAX_SET` yaml config?

janosh:
> Looks good! Can we add back in the GGA_plus_U option for MPGGADoubleRelaxStatic? There's an argument in the maker for it, but the logic like we had before to alter the config_dict's of the makers isn't

@esoteric-ephemera That argument was left over by accident. Maybe @utf's can chime in about this but if I understood correctly in yesterday's meeting, there's an existing idiomatic way of enabling +U?

src/atomate2/vasp/sets/BaseMPGGASet.yaml
```diff
@@ -111,13 +111,15 @@ INCAR:
 KPOINTS:
   reciprocal_density: 64
   reciprocal_density_metal: 200
-POTCAR_FUNCTIONAL: PBE
+POTCAR_FUNCTIONAL: PBE_54
```
utf, line 114: I don't think this should be changed to keep the input sets consistent with pymatgen?

src/atomate2/vasp/sets/mp.py
arosen93: Presumably because we want to use ALGO = "All" with R2SCAN and only use ALGO = "Fast" with GGA. So, we have ALGO = "All" in the set but override it just for the GGA step. Note: This is potentially something where inheritance can be tricky again...

arosen93:
@janosh: Shall I close #362?

janosh:
I leave that up to @utf to decide. Just didn't want to overwrite @esoteric-ephemera's changes but still bring this slight change in implementation up for discussion, hence the new PR.

src/atomate2/vasp/sets/mp.py
```diff
@@ -131,19 +134,4 @@ def get_incar_updates(
...
-        return {
-            "LWAVE": True,
-            "LCHARG": True,
-            "KSPACING": kspacing if 0.22 < kspacing < 0.44 else 0.44,
-            "ISMEAR": 0,
-            "SIGMA": 0.05,
-            "GGA": None,  # VASP 6.4+ errors if both GGA and METAGGA tag are set
-            # GGA tag might come from prev job INCAR inheritance, unset it to be safe
-        }
+        return {"LCHARGE": True, "LWAVE": True}
```
janosh, line 136: @utf Is there a better way to override the `"LCHARG": False` from `RelaxSetGenerator` here?

utf: No, this is the recommended way to do it. All the particular settings should always be specified in one place in the input set. `get_incar_updates` so it is always clear where to find them

src/atomate2/vasp/sets/mp.py
utf: But this isn't GGA, this is the SCAN calculation.

janosh: I'll move `ALGO=FAST` into `src/atomate2/vasp/sets/BaseMPR2SCANRelaxSet.yaml` then

janosh: Actually, I guess we wanted `ALGO=ALL` in `MPMetaGGARelaxSetGenerator` and only set `ALGO=FAST` in `MPMetaGGAStaticSetGenerator`?

src/atomate2/vasp/sets/BaseMPGGASet.yaml
janosh: We'll bring this up with @kristinpersson when we discuss potential input set changes but I think the idea was to break backwards compat in favor of more future-proof settings if we decide to do a complete rerun of all non-deprecated MP calcs.

janosh:
Wow, 1st time ever this PR has had green CI! 🎉 Only took a few months. 😅

src/atomate2/vasp/sets/base.py
```diff
@@ -398,20 +408,16 @@ def get_input_set(  # type: ignore
...
-            outcar=outcar,
-        )
-        kpoints_updates = self.get_kpoints_updates(
-            structure,
-            prev_incar=prev_incar,
-            bandgap=bandgap,
-            vasprun=vasprun,
-            outcar=outcar,
-        )
+        prev_incar = prev_incar if self.inherit_incar else {}
```
janosh, line 411: @utf To get the tests passing, we needed to disable INCAR inheritance. This PR (2177b0f1) adds a new keyword `inherit_incar: bool = None` to `VaspInputGenerator` which disables INCAR inheritance by default.

src/atomate2/settings.py
```diff
@@ -83,6 +83,12 @@ class Atomate2Settings(BaseSettings):
         "all the files are compressed. If 'atomate' only a selection of files related "
         "to the simulation will be compressed. If False no file is compressed.",
     )
+    VASP_INHERIT_INCAR: bool = Field(
+        False,
+        description="Whether to inherit INCAR settings from previous calculation. "
+        "This might be useful to port Custodian fixes to child jobs but can also be "
+        "dangerous e.g. when switching from GGA to meta-GGA or relax to static jobs.",
+    )
```
janosh, line 91: As advocated for by @Andrew-S-Rosen during the atomate2 maintainer meeting, INCAR inheritance can not only be controlled at the `VaspInputGenerator` level but also using the new `VASP_INHERIT_INCAR` global flag.

src/atomate2/vasp/sets/BaseMPGGASet.yaml
mkhorton: It seems like a large omission if atomate2 is unable to generate calculations compatible with current MP data. Even with refreshing the input set, presumably this should be a new, additional configuration? Especially if MP has not yet released a new database version with the new settings.

JaGeo: I agree with @mkhorton here.

janosh: Thanks all for taking a look! Reverted in 36c25294.

src/atomate2/settings.py
utf: This has changed the default behaviour for all input sets. Please can you change this back to True and we can assess this as a separate issue. This discussion should probably happen on the pymatgen issues since the default pymatgen behaviour is to inherit incar settings and we will be merging the atomate2 inputs with those in pymatgen soon.

src/atomate2/vasp/sets/base.py
```diff
@@ -305,6 +311,7 @@ class VaspInputGenerator(InputGenerator):
     symprec: float = SETTINGS.SYMPREC
     vdw: str = None
     config_dict: dict = field(default_factory=lambda: _BASE_VASP_SET)
+    inherit_incar: bool = None
```
utf, line 314: Please add docstring for this option.

src/atomate2/vasp/sets/base.py
utf: Why do you need to disable inheritance for the tests to pass? Surely the test data can be changed?

src/atomate2/settings.py
janosh: Good point. `VASP_INHERIT_INCAR` now defaults to `True`.

src/atomate2/vasp/sets/base.py
janosh: Done.

utf:
This looks good to me!

Issue #505, opened 2023-09-02
Default magmom behavior causes magnetic ordering to reset between relax -> static


mattmcdermott:
The default handling of magmoms (#375) results in an unexpected behavior that causes magnetic orderings to be reset to FM in between VASP calculations in a flow.

Consider a flow consisting of a relax followed by a static (`RelaxMaker` -> `StaticMaker`) for an input structure with an AFM magnetic ordering. After the relaxation, when the structure is copied and the input set is being generated for the static calculation, one of two cases can happen for each species in the structure: 

- A) the magmom is replaced if its value is defined in the input set  (e.g., `BaseVaspSet.yaml`). This is always a positive value (no negative magmoms, hence the AFM ordering is destroyed).
- B) the magmom is copied from the previous calculation's output structure. This happens for non-magnetic species.

This annotated snippet of code from [_get_magmoms()](https://github.com/materialsproject/atomate2/blob/0978f5de2c16cdb1746d2f1f8faaad6b8a4f22e2/src/atomate2/vasp/sets/base.py#L872) shows why:

```python
for site in structure:
    specie = str(site.specie)
    if specie in magmoms:
        mag.append(magmoms.get(specie))   # Case A: magnetic species get replaced from input set value
    elif hasattr(site, "magmom"):
        mag.append(site.magmom)  # Case B: non-magnetic species get copied
    elif hasattr(site.specie, "spin") and site.specie.spin is not None:
        mag.append(site.specie.spin)
```

The ordering of these statements leads to AFM orderings being destroyed between calculations.

@utf @janosh -- is this behavior an unintended consequence?

utf:
Yes, this is the intended behaviour in that it is consistent with the pymatgen input sets. E.g., see here:

https://github.com/materialsproject/pymatgen/blob/a5b82b5a2c6dd66fef6bfea87f39d3b6656770cb/pymatgen/io/vasp/sets.py#L456-L460

utf:
Actually, sorry I see your point. Yes, this looks like it is wrong.

utf:
[...]

PR #506, opened 2023-09-02
Enforce magmom precedence in INCAR creation

mattmcdermott:
## Summary

* Fixes issue in #505. Incar MAGMOM setting precedence now follows prescribed order.
* Adds test for `StaticSetGenerator` to catch this (method is in base class so should apply to other generators)

## TODO (if any)

  [ruff](https://ruff.rs).
  type check your code.


utf:
Great, thanks!

Issue #502, opened 2023-09-01
Structure metadata and additional properties in output documents


gpetretto:
Flows like the ones calculating elastic tensors or phonons usually produce a final document containing the parsed results but no default metadata about the structure or how the calculation was performed. This makes somewhat difficult to retrieve the outputs, as one does not have any obvious way to filter them. I think it would be usuful to convert the output documents (e.g. `ElasticDocument` and `PhononBSDOSDoc` ) to be subclasses of `StructureMetadata`. I don't think this will add much data to the DB, but could be useful.

Another related point is the fact that the final document does not contain the input parameters used to perform the simulation. To retrieve that information one would first need to match the final output to the perturbation jobs and extract the information from there. While this is not too complex, it might be still be a bit involved for a casual user. It could thus be interesting to store the input parameters of one of the main steps of the flow in the final document (e.g. one of the perturbations for elastic or phonon flow). However this would be code dependent and not necessarily representative of all the steps of the flow, so it may be less straightforward than the structure metadata.

If any of these options seem acceptable I can try to work on it.

utf:
I think making those documents a subclass of `StructureMetadata` makes sense.

Personally, I'm less keen on adding calculation parameters for the reason that you mentioned - these are code specific so creating a general schema will some work and could change in the future, e.g., with ML force fields. Both the `ElasticDocument` and `PhononBSDOSDoc` store the `uuids` of the calculations used to generate them. E.g., `ElasticDocument.fitting_data.uuids` and `PhononBSDOSDoc.uuids`. So you shouldn't need to do any matching. It is just one extra query to get those parameters.

JaGeo:
Maybe,however, we could add information to the documentation on how to retrieve the static computations belonging to an elastic or phonon computation? It might help clarifying this for all users.

utf:
Agreed, that would be very useful.

gpetretto:
Thanks for the feedback. Indeed I was using a needlessly complicated approach to query the input parameters.
I will open a PR.

mkhorton:
+1 for `StructureMetadata`, in fact I believe the `ElasticDoc` in `emmet` already does this via sub-classing `PropertyDoc`, sub-classing `StructureMetadata`. I don't recall specific details of `PropertyDoc` but I do like that it enforces some level of standardization between documents for different properties.

I strongly support keeping calculation details out of the document model, except maybe in a very abstract way (e.g. something like an originating task_id key, or a generic method field or similar).

[...]

PR #514, opened 2023-09-04
Add StructureMetadata as baseclass for output documents

gpetretto:
As discussed in #502 I added StructureMetadata to some of the final documents, i.e. `ElasticDocument`, `PhononBSDOSDoc`, `LobsterTaskDocument`, `ElectronPhononRenormalisationDoc`. I wanted to add it for the output of the defect flow as well, but I am not sure if there is a meaningful place to add it.


utf:
Great, thank you!

PR #529, opened 2023-09-22
Update lobster task schema

naik-aakash:
## Todo

- [x] reorder the lobstertask doc schema
- [x] include optional JSON data generation
- [x] update `LobsterTaskDocument` to have calc quality summary dict 
- [x] add calc quality summary text to `LobsterTaskDocument`
- [x] make coop, cohp and cobicar storing in `LobsterTaskDocument `optional (can be very large)
- [x] add tests
- [x] Fix non gzip  vasp files in lobster jobs ( Makes lobster schema fail : issue introduced after some refactoring in earlier PR https://github.com/materialsproject/atomate2/commit/726a34f133d54308b2aee2c000ad66e44cb449f5#diff-c29012eae961867ce5ed6ab6c32e0d63ef2de72e6830e7bec862dedcc5df76f7  ) 
- [x]  make schema work for non gzip files as well ( at the moment it is hard-coded to work only with gzip files)
- [x] Add model for the dicts within `CalcQualitySummary ` model
- [x] Add models for dicts nested down the schema,
- [x] add missing type hints/defaults


naik-aakash:
Hi @JaGeo , I think , I am done adding the desired changes, would be great, if you can have a look at it as well 😃  and would be happy to address if any more changes are needed

tests/vasp/lobster/schemas/test_lobster.py
```diff
@@ -165,3 +173,165 @@ def test_LobsterTaskDocument(lobster_test_dir):
...
+
+    from pymatgen.electronic_structure.cohp import Cohp, CompleteCohp
+    from pymatgen.electronic_structure.dos import LobsterCompleteDos
+
+    from atomate2.lobster.schemas import (
+        LobsterTaskDocument,
+        read_saved_json,
+    )
+
+    # Test cba saved jsons and reading
```
JaGeo, line 192: Could you expand documentation and comments here?

tests/vasp/lobster/flows/test_lobster.py
```diff
@@ -141,11 +144,14 @@ def test_lobstermaker(mock_vasp, mock_lobster, clean_dir, memory_jobstore):
     job = VaspLobsterMaker(
         lobster_static_maker=LobsterStaticMaker(),
         lobster_maker=LobsterMaker(
+            task_document_kwargs={
+                "calc_quality_kwargs": {"potcar_symbols": ["Si"], "n_bins": 10}
+            },
```
JaGeo, line 149: Is this a fix for the test?

src/atomate2/lobster/schemas.py
```diff
@@ -312,6 +315,75 @@ def from_directory(
             return (None, None, None, None, None)
 
 
+class CalcQualitySummary(BaseModel):
+    """Model describing the calculation quality of lobster run."""
+
+    minimal_basis: bool = Field(
+        None,
+        description="Denotes whether the calculation used minimal basis for the LOBSTER"
```
JaGeo, line 323: Could you add articles here as well?

src/atomate2/lobster/schemas.py
```diff
@@ -629,21 +743,268 @@ def from_directory(
...
+
+        if save_cba_jsons:
+            cba_json_save_dir = dir_name / "cba.json.gz"
+            with gzip.open(cba_json_save_dir, "wt", encoding="UTF-8") as f:
+                f.write("[")
+                if (
+                    doc.lobsterpy_data_cation_anion is not None
+                ):  # check if cation-anion analysis failed
+                    lobsterpy_analysis_type = (
+                        doc.lobsterpy_data_cation_anion.which_bonds.replace("-", "_")
```
JaGeo, line 769: Why are we doint this?

tests/vasp/lobster/flows/test_lobster.py
naik-aakash: Yes, no POTCARS and very low res lobster computations

src/atomate2/lobster/schemas.py
naik-aakash: to keep the keys consistent as in our manuscript for lightweight jsons

tests/vasp/lobster/schemas/test_lobster.py
naik-aakash: Done

src/atomate2/lobster/schemas.py
JaGeo: Yeah. I see. Might be good to change it in general

src/atomate2/lobster/schemas.py
naik-aakash: Done

src/atomate2/lobster/schemas.py
JaGeo: Yes. I think this is a weird fix.

naik-aakash: Okay will make the suggested change 😄

Zhuoying:
image
Hi @naik-aakash Let me know when you fix this.

Thanks for reviewing this @JaGeo! I am helping to push the PRs moving forward as needed. When it is done, I will leave a comment here as **ready-to-merge**

JaGeo:
@Zhuoying Thanks! Yes, we are still working on this. We were just releasing a new LobsterPy version today so hat we can make the final changes.

Zhuoying:
@JaGeo Sounds great! Thanks for letting me know. Ping me anytime when you think it is ready to merge.

naik-aakash:
Hi @JaGeo , I have now updated the lobsterpy version and tests also has been adapted accordingly. If you think, it is fine, I can ask for it to be merged 😃

JaGeo:
@naik-aakash Yes, looks good!

naik-aakash:
Hi @Zhuoying , this PR is ready to be merged. 😄  Thanks !!

src/atomate2/lobster/schemas.py
```diff
@@ -621,21 +750,268 @@ def from_directory(
...
+                            ):  # add comma separator between two dicts
+                                f.write(",")
+                        else:
+                            data = {attribute: doc_here.__getattribute__(attribute)}
+                            json.dump(data, f)
+                            if (
+                                attribute != list(doc_here.__fields__.keys())[-1]
+                            ):  # add comma separator between two dicts
+                                f.write(",")
+                f.write("]")
```
utf, line 977: What is going on here? Can't you just use `jsanitize` to get a dict representation of the data and then use `dumpfn`?

src/atomate2/lobster/schemas.py
```diff
@@ -775,3 +1151,76 @@ def _get_strong_bonds(
                         }
                     )
     return bond_dict
+
+
+def read_saved_json(
```
utf, line 1156: Why do you need this function. Can't you just use `monty.serialization.loadfn`?

src/atomate2/lobster/schemas.py
JaGeo: (Should have catched this. Sorry.)

naik-aakash: Here as well, I wrote this function as most of the json saving libraries tend to create Python representation of the entire taskdoc in the memory which results in too much memory usage. The way I am m writing the JSONS, I am able to avoid excessive memory usage (writing the file incrementally by converting one field at a time to dict representation).

Maybe I am overlooking something here. Any suggestions are welcome

utf: Out of interest, how big are the generated files?

src/atomate2/lobster/schemas.py
```diff
@@ -621,21 +750,268 @@ def from_directory(
...
+            # include additional fields for all bonds
+            cohp_data=cohp_obj,
+            coop_data=coop_obj,
+            cobi_data=cobi_obj,
         )
+
+        if save_cba_jsons:
+            cba_json_save_dir = dir_name / "cba.json.gz"
+            with gzip.open(cba_json_save_dir, "wt", encoding="UTF-8") as f:
+                f.write("[")
```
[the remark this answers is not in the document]
JaGeo: Just to add this: this is mostly relevant when DOSs and COHPs with many points for benchmarking are processed (as we did for our Lobster database). Typically, these files are not as large.

src/atomate2/lobster/schemas.py
utf: Is it really that bad? I'd imagine this would only cause an issue if the files are above 1GB.

utf: The alternative would be to store these files in an alternative format like hdf5

src/atomate2/lobster/schemas.py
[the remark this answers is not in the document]
JaGeo: With regard to size: 
800 of such files amount to 4 GB. Thus, it must be much less than 1 GB per file. The smaller files for all 1500 compounds amount to 700 MB. See https://zenodo.org/record/8091844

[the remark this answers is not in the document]
naik-aakash: > With regard to size: 800 of such files amount to 4 GB. Thus, it must be much less than 1 GB per file. The smaller files for all 1500 compounds amount to 700 MB. See https://zenodo.org/record/8091844

You ~40GB for 800 of such files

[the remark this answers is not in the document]
JaGeo: Oh, yes. Sorry. 40 GB but still much less than 1 GB per file. Very sorry for the typo!

src/atomate2/lobster/schemas.py
naik-aakash: > Is it really that bad? I'd imagine this would only cause an issue if the files are above 1GB.

Yeah, the size usually ends up of JSONS above 1 GB (on average 2-4 GB: just checked a few generated files now) in the uncompressed state when using a larger number of points say 10000 points for writing COHPCAR, COOPCAR, COBICAR, in LOBSTER runs. 

To add to this, the database we constructed had fairly simpler and smaller structures, It will be much larger for cases when moving to complicated and larger structures 

Thus I also added an option to skip storing this larger cohpcar objects in the db

naik-aakash:
Hi @utf, First of all, thank you for your suggestion to use monty encoders. It certainly has simplified the code a lot. However, using `monty.serialization.dumpfn` is not memory efficient. I tested this for a simple SiO2 structure. I have attached the memory profiler outputs from the old, `dumpfn` and new implementation.

# Old
old

# Via `monty.serialization.dumpfn`
monty_mp7000

# New : Current
new_mp7000_iter_implementation

Let me know if any more changes are desired.

For the cba_json.gz file, I need to have data in the specific keys, as per our publication; thus, the code is a bit longer, but have added comments to make it clear what happens in each step.

src/atomate2/lobster/schemas.py
naik-aakash: We need to definitely have this function `read_saved_json` as you can see below, the memory requirements can be super large to load these files for larger structures.

naik-aakash:
Hi @munrojm, we have updated the schema to include the JSON data generation (Both large and small JSON). It would be great if you could have a look at it and let us know what you think about this.

Now the smaller jsons also have LOBSTER DOS objects in them.

src/atomate2/lobster/schemas.py
```diff
@@ -304,6 +308,90 @@ def from_directory(
...
+    band_overlaps: dict = Field(
+        None,
+        description="Dict summarizing important information from the "
+        "bandOverlaps.lobster file to evaluate the quality of the projection, "
+        "namely whether the file is generated during projection (i.e., larger "
+        "deviations exist), the maximum deviation observed, percent of k-points "
+        "above the threshold set in the pymatgen parser (during data generation the "
+        "value was set to 0.1)",
+    )
+    dos_comparisons: dict = Field(
```
tschaume, line 336: Do the four `dict` fields in `CalcQualitySummary` have an expected schema or are they free-form? If the former, we should probably enforce it in the model.

naik-aakash: Thanks! They have an expected form, I can create a model for that as well 😃

naik-aakash:
Hi @JaGeo, I pushed the changes to make schema work even if gzipping is switched off. Can you have a look and let me know if this change is fine? (I do not expect you to look into it on the weekend 😉 ). 

I have added a test as well, and it seems to work.

PS:
I tried using the existing atomate2 utils, but the code seems to get a bit messy. Thus, I came up with writing this FileNames class that seemed much simpler.

tests/vasp/lobster/schemas/test_lobster.py
```diff
@@ -173,11 +173,13 @@ def test_lobster_task_document_non_gzip(lobster_test_dir):
     # Simulate settings when gzip is turned off in atomate2 settings
     SETTINGS.LOBSTER_ZIP_FILES = False
 
+    # copy test files to temp path
+    copy_files(src_dir=lobster_test_dir / "lobsteroutputs/mp-2534", dest_dir=tmp_path)
     # Unzip test files to check if schema still works
-    gunzip_files(lobster_test_dir / "lobsteroutputs/mp-2534")
+    gunzip_files(tmp_path)
```
JaGeo, line 179: So, you are unzipping all files?

src/atomate2/lobster/schemas.py
naik-aakash: @tschaume , have now added models for the same as per your suggestion 😄

src/atomate2/lobster/schemas.py
```diff
@@ -301,7 +318,196 @@ def from_directory(
...
+        description="Bool indicating whether atoms classification as cations "
+        "or anions based on Loewdin charge signs of LOBSTER "
+        "agree with BVA analysis",
+    )
+
+
+class BandOverlapsComparisons(BaseModel):
+    """Model describing the Band overlaps field in the CalcQualitySummary model."""
+
+    file_exists: bool = Field(
```
munrojm, line 383: Any field like this with a default value of None should probably be made Optional, particularly in support of pydantic2. If you don't want it optional, then I would set it to ... if you also don't know what the default value should be.

src/atomate2/lobster/schemas.py
munrojm, line 1: Any remaining fields that are just type hinted as a `list` should also include what is in the list so it isn't read as `list[Any]`.

src/atomate2/lobster/schemas.py
munrojm: I will add that any other remaining fields that are validated as a raw `dict` should also have an explicit model if the intention is to query the data down the line.

src/atomate2/lobster/schemas.py
naik-aakash: Done

src/atomate2/lobster/schemas.py
naik-aakash: Done, adapted the schema now

naik-aakash:
Hi @munrojm and @JaGeo, I am done addressing the comments. Let me know if anything more needs to be done in this PR.

src/atomate2/lobster/schemas.py
naik-aakash: Done

munrojm:
> Hi @munrojm and @JaGeo, I am done addressing the comments. Let me know if anything more needs to be done in this PR.

Things look good to me.

naik-aakash:
Hi @utf, I think this PR is ready to be merged if there are no more concerns related to code requirements.

JaGeo:
Could someone review this as this is central to getting the Lobster data into the MP databas? 😃 Thanks in advance

utf:
Thanks for all your work on this. @naik-aakash @JaGeo

utf:
As a separate thought, if lobster will be on the MP database, should we move these schemas to emmet?

JaGeo:
I assume once everything works well, we probably need to do this. 

(I would also like to implement a workflow compatible with the current workflow in atomate, i.e. same exchange correlation functional etc )

PR #537, opened 2023-09-29
removed py38 support, and add py 3.11 support

### LINKED ISSUE #530: Problem with python 3.8 support with newer releases of dependencies
# Issue

`pymatgen` and `lobsterpy` no longer support python 3.8. This leads to failing tests with added new functionality., as atomate2 still supports python3.8

Can the python 3.8 support safely removed from atomate2 as well? Or is there any other work around to avoid this issue @utf ?

Thank you for your help :smile:  and looking forward to your thoughts on it

naik-aakash:
# Changes

1. Remove Python 3.8 support
2. Add python 3.11 

Closes #530

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "3.9"
           cache: pip
           cache-dependency-path: pyproject.toml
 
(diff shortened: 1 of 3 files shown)


naik-aakash:
Hi @janosh, I have tried to remove py38 support, but I cannot figure out why the workflows still point to py3.8 as the required test. Can you please point out what I am missing here?

janosh:
That's not something you can change. @utf will have to update which CI is required when he merges this PR.

utf:
Thanks!

PR #540, opened 2023-09-30
Remove `__all__` from all modules

### LINKED ISSUE #539: Remove `__all__` to discourage wildcard imports
I think wildcard imports are a bad practice and a feature I wish Python didn't support.

By defining `__all__` across atomate2 modules, we're encouraging this practice. It's additional maintenance burden to keep the list of strings in sync with the actual symbol names in a module. It would also increase ecosystem consistency to remove them since other MP repos do not use `__all__`.

janosh:
Closes #539.


utf:
Cheers!

Issue #549, opened 2023-10-03
`mock_vasp` should default to assert all INCAR tags match reference files


janosh:
The `mock_vasp` pytest fixture

https://github.com/materialsproject/atomate2/blob/468fb57b4a51a212ea385bed9e9ecae9b97dba4f/tests/vasp/conftest.py#L25-L33

by default doesn't check that the INCAR tags returned by the set generator being tested match the reference files. You have to explicitly pass per-job `incar_settings` to have them checked for agreement:

https://github.com/materialsproject/atomate2/blob/468fb57b4a51a212ea385bed9e9ecae9b97dba4f/tests/vasp/jobs/test_core.py#L148-L150

Maybe a better default behavior would be to check _all_ INCAR tags and allow passing `{"incar_settings": ["ISMEAR", "LREAL", "NSW"]}` if you only want to check that subset?

Pinging @esoteric-ephemera @JaGeo @Andrew-S-Rosen @mkhorton @computron to get a broad range of opinions.

utf:
The issue with checking all input settings is that any time an input set changes you need to update all the test files which is a lot of work.

So the default has been to check only the settings that are important for that specific calculation (e.g., a relaxation should have NSW != 0.

janosh:
Those cases you could still explicitly pass in a subset of tags to check. I feel like following the principle of least surprise here means checking all INCAR tags by default.

Maybe another option would be defining `__all__` as a special value for comparing the full generated INCAR to its reference. As it seems like a lot of boilerplate having to manually create a list of all INCAR tags that are present.

3rd option would be to define "ignore_incar_settings: list[str] | None = None" which when passed makes the test check all except the ignored INCAR tags.

JaGeo:
I actually think it's a good idea to make sure more incar flags are tested. If @janosh did not realize all incar flags are tested by default, this will happen to many developers. To be honest, I had nearly overlooked this as well.

I would vote for option 1 proposed by @janosh, where one can check subsets of flags that need to pass.

utf:
I'm not definitely not averse to imposing stricter checking. Especially if the input sets are settling down. But I just want to highlight that once the input sets are moved to pymatgen, if anyone makes an edit, this will cause a lot of atomate2 tests to fail. Whose responsibility is it to fix those failing tests? If we try and enforce that the person editing pymatgen also fixes atomate2, that will be fine. But otherwise, it will always come down to one of us fixing the tests and that can be a continued effort.

[...]

PR #550, opened 2023-10-03
Fix wrong INCAR values in MP workflows

janosh:
Follow up PR to #504 that fixes some unexpected INCAR values after noticing in #549 that they aren't actually being checked to match the reference INCARs.


janosh:
I just noticed the MP workflow reference files in #504 accidentally included POTCARs. Just deleted them in dc7c4df.

@JaGeo I noticed another 2 POTCAR in the Lobster reference files

- `tests/test_data/lobster/lobsteroutputs/mp-754354/POTCAR.gz`
- `tests/test_data/lobster/lobsteroutputs/mp-2534/POTCAR.gz`

which I deleted as well. Is that ok with you?

JaGeo:
@janosh Yes, please. 🙃

utf:
Great, thanks!

PR #562, opened 2023-10-11
atomate2 aims

tpurcell90:
## Summary

Create an FHI-aims interface into atomate2 (moved from: https://github.com/tpurcell90/atomate2-fhi-aims). This was originally started as a project to incorporate FHI-aims into jobflow as a part of the NOMAD-CoE, working with @davidwaroquiers and @ansobolev. Originally we did not know if this could be incorporated into atomate2, which is why we had a separate repository, but everyone now agrees that we should move it to atomate2.

* FHI-aims interface for base jobs, input generators and schemas
* ASE interface that is aims specific (can be moved into a general framework if everyone agrees)
* Generalization of the phonon workflow originally made for VASP

* ASE: This is needed because currently the only way to interact with FHI-aims in python is through ASE. We are in the process of discussing a new FHI-aims controlled python interface, but I have not gotten any type of general support to start to write that. Additionally the SocketIO calculators are useful for the phonon calculations as it can speed up the SCF convergence in FHI-aims by a factor of ~2. The largest problem with this however, is the delay in getting ASE 3.23 out, which is needed for this code interface
* seekpath: This is used in the common phonon workflows, but I did not see it in the dependencies

Note I did not add these to the pyproject file before discussing
## TODO (if any)

* Discussion of ASE dependency
* @JaGeo can you check that my modifications to the phonon workflow did not break the VASP workflow? I don't have a VASP license to check this
* run ruff
* run mypy

## Checklist

title.

  [ruff](https://ruff.rs).
  type check your code.


JaGeo:
@tpurcell90 @utf I am happy to help. I left one immediate comment there already. For the rest, I need more time.

src/atomate2/common/jobs/phonons.py
```diff
@@ -117,7 +124,7 @@ def get_supercell_size(
     return transformation.transformation_matrix.tolist()
 
 
-@job(data=[Structure])
```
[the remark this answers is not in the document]
tpurcell90: Okay I must have pulled the workflow from an earlier version with out this change, I'll correct this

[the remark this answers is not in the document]
JaGeo: Yes, this is only relevant for very large structures that can be run with ML potentialy

utf:
Thanks for the PR. In general this looks really high quality and hopefully there won't be much of a barrier to getting it merged. From a quick scan through, it looks like merging the phonon workflows will be the main challenge. If you're happy to coordinate that aspect @JaGeo that would be great.

JaGeo:
> * seekpath: This is used in the common phonon workflows, but I did not see it in the dependencies

We have additional dependencies for the phonon runs. It is part of it.

src/atomate2/common/flows/phonons.py
```diff
@@ -0,0 +1,357 @@
...
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from atomate2.vasp.jobs.base import BaseVaspMaker
+    from emmet.core.math import Matrix3D
+    from pymatgen.core.structure import Structure
+
+    from atomate2.aims.jobs.base import BaseAimsMaker
+
+__all__ = ["BasePhononMaker"]
```
JaGeo, line 29: We got rid of this recently.

src/atomate2/common/flows/phonons.py
```diff
@@ -0,0 +1,357 @@
...
+          with this workflow. Otherwise, we must rely on seekpath
+        - "conventional" will enforce to start the phonon computation
+          from the conventional standard structure
+          according to Setyawan, W., & Curtarolo, S. (2010).
+          High-throughput electronic band structure calculations:
+          Challenges and tools. Computational Materials Science,
+          49(2), 299-312. doi:10.1016/j.commatsci.2010.05.010.
+          We will however use seekpath and primitive structures
+          as determined by from phonopy to compute the phonon band structure
+    bulk_relax_maker : .BaseAimsMaker or None
```
JaGeo, line 95: I would add the other options here as well. See also below.

src/atomate2/common/flows/phonons.py
```diff
@@ -0,0 +1,357 @@
...
+        Make flow to calculate the phonon properties.
+
+        Parameters
+        ----------
+        structure : .Structure
+            A pymatgen structure. Please start with a structure
+            that is nearly fully optimized as the internal optimizers
+            have very strict settings!
+        prev_dir : str or Path or None
+            A previous vasp calculation directory to use for copying outputs.
```
JaGeo, line 169: Update necessary on the documentation

src/atomate2/common/jobs/phonons.py
```diff
@@ -230,7 +238,7 @@ def generate_frequencies_eigenvectors(
         Additional parameters that are passed to PhononBSDOSDoc.from_forces_born
 
     """
-    return PhononBSDOSDoc.from_forces_born(
+    phonon_doc = PhononBSDOSDoc.from_forces_born(
```
JaGeo, line 241: you could directly return. I changed this recently.

src/atomate2/common/jobs/phonons.py
```diff
@@ -246,34 +254,39 @@ def generate_frequencies_eigenvectors(
...
     structure: Structure object
-        Fully optimized structure used for phonon computations.
+        Fully optimized structure used for phonon computations
     supercell_matrix: Matrix3D
         supercell matrix for meta data
-    phonon_maker : .BaseVaspMaker
-        A VaspMaker to use to generate the elastic relaxation jobs.
-    prev_vasp_dir : str or Path or None
-        A previous vasp calculation directory to use for copying outputs.
+    phonon_maker : BaseVaspMaker or ForceFieldStaticMaker or BaseAimsMaker
```
JaGeo, line 283:
```suggestion
    phonon_maker : .BaseVaspMaker or .ForceFieldStaticMaker or .BaseAimsMaker
```

src/atomate2/common/jobs/phonons.py
```diff
@@ -282,84 +295,49 @@ def run_phonon_displacements(
         "dirs": [],
     }
 
-    for i, displacement in enumerate(displacements):
-        if prev_vasp_dir is not None:
-            phonon_job = phonon_maker.make(displacement, prev_vasp_dir=prev_vasp_dir)
+    if socket:
+        if prev_dir is not None:
+            phonon_job = phonon_maker.make(displacements, prev_dir=prev_dir)
```
JaGeo, line 300: @utf I am wondering if we should in general switch to prev_dir.

src/atomate2/common/schemas/phonons.py
```diff
@@ -27,8 +26,18 @@
...
 logger = logging.getLogger(__name__)
 
+__all__ = [
+    "PhononBSDOSDoc",
+    "PhononComputationalSettings",
+    "PhononUUIDs",
+    "PhononJobDirs",
+    "ThermalDisplacementData",
+]
+
```
JaGeo, line 40: can be removed now.

src/atomate2/common/schemas/phonons.py
```diff
@@ -84,16 +93,16 @@ def __init__(self, force_constants: List[List[Matrix3D]]):
...
-    static_run_job_dir: Optional[str] = Field(
+    static_run_job_dir: str = Field(
         None, description="Directory where static run was performed."
     )
-    born_run_job_dir: Optional[str] = Field(
+    born_run_job_dir: str = Field(
         None, description="Directory where born run was performed."
     )
-    optimization_run_job_dir: Optional[str] = Field(
+    optimization_run_job_dir: str = Field(
```
JaGeo, line 105: There were some recent updates. Maybe make sure this is included

JaGeo (changes requested):
@tpurcell90 I was too curious to not look at it.
I left several comments. One major comment that I have: The phonon forcefield workflow should probably also inherit from this new base class if possible: https://github.com/materialsproject/atomate2/blob/main/src/atomate2/forcefields/flows/phonons.py

src/atomate2/common/jobs/phonons.py
JaGeo: I agree that will be a breaking change and additional work, but as more and more codes are coming and we work on "common" workflows, it is probably something we have to do sooner or later. We might also need to discuss how to connect jobs from different codes in the best way.

src/atomate2/vasp/flows/phonons.py
```diff
@@ -1,95 +1,26 @@
...
-        Name of the flows produced by this maker.
-    sym_reduce : bool
-        Whether to reduce the number of deformations using symmetry.
-    symprec : float
-        Symmetry precision to use in the
-        reduction of symmetry to find the primitive/conventional cell
-        (use_primitive_standard_structure, use_conventional_standard_structure)
-        and to handle all symmetry-related tasks in phonopy
-    displacement: float
-        displacement distance for phonons
```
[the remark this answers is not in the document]
tpurcell90: Adding it back now

src/atomate2/vasp/flows/phonons.py
```diff
@@ -102,36 +33,9 @@ class PhononMaker(Maker):
...
 
-    name: str = "phonon"
-    sym_reduce: bool = True
-    symprec: float = 1e-4
-    displacement: float = 0.01
-    min_length: float | None = 20.0
-    prefer_90_degrees: bool = True
-    get_supercell_size_kwargs: dict = field(default_factory=dict)
-    use_symmetrized_structure: str | None = None
+    code: str = "vasp"
```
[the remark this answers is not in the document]
tpurcell90: Adding it back now with the changed items at the bottom of the list (applies to VASP and Aims)

src/atomate2/common/flows/phonons.py
tpurcell90: removing

src/atomate2/common/flows/phonons.py
```diff
@@ -0,0 +1,357 @@
...
+        epsilon_static: Matrix3D | None = None,
+        total_dft_energy_per_formula_unit: float | None = None,
+        supercell_matrix: Matrix3D | None = None,
+    ):
+        """
+        Make flow to calculate the phonon properties.
+
+        Parameters
+        ----------
+        structure : .Structure
```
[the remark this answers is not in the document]
tpurcell90: I added it in

src/atomate2/common/flows/phonons.py
tpurcell90: Corrected and will do another round of docstring checks once the force fields are included

src/atomate2/common/jobs/phonons.py
tpurcell90: This was fixed by one of the linters. I was not sure if that was a style thing atomate2 did or not

src/atomate2/common/jobs/phonons.py
tpurcell90: When I made the change I was not sure what the overall plan was for a generalization like this. I am happy to help in the discussions about connecting jobs from different codes. 

Let me know if you want me to make a change here in the VASP jobs for `prev_dir` vs `prev_vasp_dir`

src/atomate2/common/schemas/phonons.py
tpurcell90: made this change will do a diff check once the forcefields are adapted

src/atomate2/common/phonon_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    float
+        The correct conversion factor
+
+    Raises
+    ------
+    ValueError
+        If code is not defined
+    """
+    # TARP: This is based on self.code == "vasp" in the old forcefields workflow
+    if code in ["forcefields", "vasp"]:
```
tpurcell90, line 27: Is this correct? In the original forcefields workflow the code was still set to VASP so I assumed that the factors were the same.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "numpy",
     "pydantic",
     "pymatgen>=2023.1.9",
+    "ase@git+https://gitlab.com/ase/ase.git@793179f4#egg=ase",
```
JaGeo, line 37: Maybe also a part of the optional dependencies and to be added to strict?

src/atomate2/common/phonon_utils.py
JaGeo: Yes, exactly.

src/atomate2/common/schemas/phonons.py
tpurcell90: I ran a diff on all of the common phonon files and double checked the VASP ones I think everything is consistent, but I will do another check in the morning to be sure

pyproject.toml
tpurcell90: I think that would be better. I was not sure what level to put it on, @utf do you have a prefrence?

tpurcell90:
> @tpurcell90 I was too curious to not look at it. I left several comments. One major comment that I have: The phonon forcefield workflow should probably also inherit from this new base class if possible: https://github.com/materialsproject/atomate2/blob/main/src/atomate2/forcefields/flows/phonons.py

The forcefields now have the same format as VASP/aims

JaGeo:
> The forcefields now have the same format as VASP/aims
@tpurcell90 Awesome! Thank you.

JaGeo:
@utf `prev_dir` is now breaking the VASP workflow. Do you maybe have a good suggestion to solve this? Should we define `prev_dir` in addition to `prev_vasp_dir` or rename completely?

utf:
For now, can we follow the approach taken in the Magnetic ordering maker ([see here](https://github.com/materialsproject/atomate2/blob/5127b108c4a76002d24a4c9b018a6dcfc304d7ec/src/atomate2/common/flows/magnetism.py#L163C7-L163C7)) from this PR: https://github.com/materialsproject/atomate2/pull/432

I will make a separate issue about renaming `prev_vasp_dir` which we can then implement after some consultation.

tpurcell90:
I used that approach for the phonon workflows as well. I set the `forcefields` to `None` since it does not look like that used a `prev_dir` keyword at all

src/atomate2/common/jobs/phonons.py
tpurcell90: See below for the solution

pyproject.toml
utf: Yes, put it in the optional dependencies under aims (you will need to create this) and also add it to struct with a specific version pinned.

JaGeo:
> I used that approach for the phonon workflows as well. I set the `forcefields`' argname to `None` since it does not look like that used a `prev_dir` keyword at all

Yes, it does not. No files need to be copied as none are generated.

src/atomate2/common/schemas/phonons.py
tpurcell90: Everything should be consistent now

JaGeo:
Could someone approve the workflow? @janosh @utf ?

JaGeo:
This approval thing is always a bit exhausting 😅. I think tge ase dependency looks fixed now. Hopefully, it works now.

tpurcell90:
No worries, I took the opportunity to double check pytest results work locally and found a few more minor bugs that I fixed. Everything should pass now. I also asked @ansobolev to review the documentation one last time so if everything works out this should be good to go by the end of the week.

src/atomate2/aims/flows/phonons.py
Zhuoying, line 1: I saw this commit of following the same approach as in the magnetism workflow: 
adding `def prev_calc_dir_argname(self)`.
But `abc` and `@abstractmethod` were only added to [src/atomate2/common/flows/phonons.py](url) and missing in other files (aims/forcefields/vasp)?

JaGeo: I  definitely see it being implemented in phonons.vasp, phonons.aims and phonons.forecfield. I am not aure understand the comment

src/atomate2/vasp/flows/phonons.py
```diff
@@ -133,3 +133,15 @@ class PhononMaker(BasePhononMaker):
     phonon_displacement_maker: BaseVaspMaker = field(
         default_factory=PhononDisplacementMaker
     )
+
+    @property
+    def prev_calc_dir_argname(self):
```
Zhuoying, line 138: The same issue, should we add `@abstractmethod` decoration here as it is in [src/atomate2/common/flows/phonons.py](url)?

src/atomate2/common/flows/phonons.py
```diff
@@ -354,3 +365,15 @@ def make(
 
         # create a flow including all jobs for a phonon computation
         return Flow(jobs, phonon_collect.output)
+
+    @property
+    @abstractmethod
```
Zhuoying, line 373: `@abstractmethod` was added here and it looks correct to me. But it seems missing in vasp/aims/forcefields.

src/atomate2/aims/flows/phonons.py
```diff
@@ -138,3 +138,15 @@ def __post_init__(self):
                 self.phonon_displacement_maker = PhononDisplacementMakerSocket()
             else:
                 self.phonon_displacement_maker = PhononDisplacementMaker()
+
+    @property
```
Zhuoying, line 141: Should we add `@abstractmethod` here?

Zhuoying (commented):
Just want to confirm, should we add `@abstractmethod` decorator before all the `def prev_calc_dir_argname(self):` in aims/common/forcefields/vasp?

src/atomate2/vasp/flows/phonons.py
tpurcell90: I don't think we need the `@abstractmethod` decorator there since in the child classes those are not abstract. `@abstractmethod` tells python that a method  will be overwritten by the child class.

src/atomate2/aims/flows/phonons.py
Zhuoying: Sorry, my bad for not referring to the lines of codes. See the comments below. 
If there is any reason for not adding the decoration of `@abstractmethod` in some cases, just ignore my comments.

src/atomate2/aims/flows/phonons.py
JaGeo: But why should it be still abstract there? As far as I understand the decorator, it indicates that inheriting classes have to implement the method.

src/atomate2/vasp/flows/phonons.py
tpurcell90: If someone were to make `PhononMakerVaspSpecialFeature` inheriting from `PhononMakerVasp` then `prev_calc_dir_argname` could become abstract in `PhononMakerVasp` if `PhononMakerVaspSpecialFeature` were to change that property.

Zhuoying: okay, just want to confirm it is only needed in some cases.
@tpurcell90 Thanks for the point-by-points fixes. It is very clear and easy to track your commits.
I suppose it is ready to merge now.
Since it is such a heavy PR, I would like to wait for @utf to click the final button.

tpurcell90: The only two open things are the ASE dependency linking to gitlab vs an exact version (waiting for the new release) and that I want my colleague/coauthor @ansobolev of the code base to do one final review to make sure that he agrees to it.

JaGeo: Just wanted to add that I mainly looked at the phonon parts in the review. Thus, reviewing the other parts and that they agree with atomate2 as well, might be good.

I had one additional question: do the units from the output schema agree with the VASP units? E.g., if we wanted to build a common elastic workflow, would we need to convert? I think we have such a conversion for the forcefields. (@utf, what overall philosophy does atomate2 follow here?)

tpurcell90: @ansobolev can correct me, but I think they are the same with the caveat that FHI-aims uses CODATA 2002 for all of its units so we may differ at the 6-9th digit after the decimal.

tpurcell90: For clarification internally FHI-aims uses Hartrees and then converts to eV using CODATA 2002 constants, which can lead to small differences

JaGeo: Thanks!

Stresses as well? I just remember a conversion that was necessary for the stresses coming from the forcefields in ase. Maybe, I am mistaken. Just wamt to make sure.

tpurcell90: What units does VASP use for stresses? FHI-aims uses ev/A^3, but I think there was a discussion at some point about a CODATA 2002 discrepancy between vibes and FHI-aims. I forget the issue, but that is when we learned ASE uses the modern CODATA and FHI-aims uses the 2002 version and that led to issues at the 6-9th decimal.

JaGeo: It's kbar (see https://www.vasp.at/wiki/index.php/PSTRESS): 
https://github.com/materialsproject/atomate2/blob/afcce840c7a86be527efbfc09c2b0864906f50a5/src/atomate2/forcefields/schemas.py#L59

tpurcell90: Okay we can do the same conversion in our schema as well so they are all consistent withe each other

JaGeo: Yes. And let's wait for @utf 's suggestions :).

tpurcell90: I made the change in the parser, but that can be easily reverted (one multiplication factor). I also noticed that I copied over one too many tests for the parser so I removed it

JaGeo:
Thanks. Now we wait for someone turning on the workflows. :D

janosh:
@tpurcell90 could submit another PR to for example be added to the list of contributors, which would remove the first time contributor status, and avoid the need to approve workflow runs.

tpurcell90:
@janosh I made that PR just now. I was not sure why all of the workflows needed to wait for approval.

janosh:
> I was not sure why all of the workflows needed to wait for approval.

It's a GitHub abuse prevention feature to avoid cryptomining. See [this blog post](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors).

> Based on conversations with several maintainers, we feel this step is a good balance between manual approval and existing automated workflows. This will be the default setting and, as of now, **there is no way to opt out of the behavior**, but we do have plans to implement a few additional settings to provide more flexibility.

To my knowledge, the last part of adding customization settings for this behavior hasn't happened (yet). I wish they would only require 1 approval per PR, not demand re-approval for each commit.

tpurcell90:
@ansobolev Can you confirm you agree with the Optional tags I changed in the schemas?

ansobolev:
> @ansobolev Can you confirm you agree with the Optional tags I changed in the schemas?
Yes, everything is fine I guess

tpurcell90:
@utf is there anything more we need to do for this? For the ASE dependency is it okay to use the gitlab repo or do we need to reframe it for ASE 3.22.1?

utf:
@tpurcell90, I will have a final detailed pass through the PR. Due to teaching commitments I won't get to this until Friday this week.

In terms of ase, we shouldn't use the GitLab repo - this will break our automated dependency updating system (dependabot). Does ase 3.22.1 support the features in this PR? If so, then just update the dependency to this. Otherwise, we'll have to wait until a new ase version is released before we can merge this PR.

tpurcell90:
Okay no rush, I just wanted to know the status of it.

Right now it does, but I can try to strip out the 3.23 dependencies. Waiting for a new version of ASE is not ideal since it is unclear when that will actually happen.

JaGeo (approved):

janosh (commented):
I think it would help with repo size and PR review-ability to compress all the `outputs` files

src/atomate2/aims/flows/gw.py
```diff
@@ -0,0 +1,91 @@
...
+
+from atomate2.aims.jobs.base import BaseAimsMaker, ConvergenceMaker
+from atomate2.aims.jobs.core import BandStructureMaker, GWMaker, StaticMaker
+from atomate2.aims.sets.bs import BandStructureSetGenerator, GWSetGenerator
+from atomate2.aims.sets.core import StaticSetGenerator
+from atomate2.aims.utils.msonable_atoms import MSONableAtoms
+
+
+@dataclass
+class PeriodicGWConvergenceMaker(BaseAimsMaker):
```
utf, line 18: I'm a bit conflicted by this class. This subclasses the `BaseAimsMaker` which is a `job` maker, but then this class itself is a `Flow` maker. So I think this will result in confusion. I thing the reason you've done it like this is to only have one instance of the `input_set_generator.parameters` that can be shared between all subjobs. However, I don't think this is the best way to do it.

My suggestion would be:
1. Don't subclass `BaseAimsMaker`.
2. Add two fields to the dataclass `static_maker` and `gw_maker`. These can have default values and use these makers inside the `make` function.
3. If you need to update the parameters for the convergence, then you can just use a powerup or the `update_maker_kwargs` function. This will update all parameters in the flow simultaneously.
4. If you want to support both molecules and crystals, I would make two subclasses of this maker, each with different defaults for the `static_maker`. Alternatively, you can add support inside one class, but the static will need to be added dynamically. See my previous comment about using `Response(replace=flow)`.

src/atomate2/aims/flows/gw.py
```diff
@@ -0,0 +1,91 @@
...
+
+        convergence = ConvergenceMaker(
+            maker=gw_maker,
+            epsilon=self.epsilon,
+            criterion_name=self.criterion_name,
+            convergence_field=self.convergence_field,
+            convergence_steps=self.convergence_steps,
+        )
+
+        gw = convergence.convergence_iteration(
```
utf, line 87: Why is this called `convergence_iteration` and not `make`?

src/atomate2/aims/io/aims_output.py
```diff
@@ -0,0 +1,201 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from emmet.core.math import Matrix3D, Vector3D
+
+    from atomate2.aims.utils.msonable_atoms import MSONableAtoms
+
+
+class AimsOutput(MSONable):
```
utf, line 19: I think we should move the contents of `atomate2/aims/io` to `pymatgen/io/aims`.

src/atomate2/aims/jobs/base.py
```diff
@@ -0,0 +1,313 @@
...
+    criterion_name: str = "energy_per_atom"
+    epsilon: float = 0.001
+    convergence_field: str = field(default_factory=str)
+    convergence_steps: list = field(default_factory=list)
+
+    def __post_init__(self):
+        """Set the value of the last index."""
+        self.last_idx = len(self.convergence_steps)
+
+    def make(self, atoms):
```
utf, line 164: What is this make function for? Can you just rename `convergence_iteration` to `make` to make this consistent with other `Makers`.

src/atomate2/aims/jobs/base.py
```diff
@@ -0,0 +1,313 @@
...
+                < self.epsilon
+            )
+
+        split_job_dir = str(job_dir).split(":")[-1]
+        convergence_file = Path(split_job_dir) / CONVERGENCE_FILE_NAME
+        with open(convergence_file, "w") as f:
+            json.dump(convergence_data, f)
+
+    @job(name="Getting the results")
+    def get_results(self, prev_dir: Path | str) -> dict[str, Any]:
```
utf, line 299: Do you need a separate `job` for `get_results`. Couldn't the file contents just be returned from `update_convergence_file`? And then you only need two jobs in the flow.

src/atomate2/aims/jobs/base.py
```diff
@@ -0,0 +1,313 @@
...
+
+            replace_flow = Flow(
+                [next_base_job, update_file_job, next_job], output=next_base_job.output
+            )
+            return Response(detour=replace_flow, output=replace_flow.output)
+        task_doc = AimsTaskDoc.from_directory(prev_dir_no_host)
+        return ConvergenceSummary.from_aims_calc_doc(task_doc)
+
+    @job(name="Writing a convergence file")
+    def update_convergence_file(
```
utf, line 246: Typically, `Makers` don't have more than one function that produces a job. Unless there is a very good reason to, I would refactor this out to just be a function rather than an instance method.

src/atomate2/aims/run.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Run FHI-aims.
+
+    Parameters
+    ----------
+    aims_cmd : str
+        The command used to run FHI-aims (defaults to ASE_AIMS_COMMAND env variable).
+    """
+    if aims_cmd is None:
+        aims_cmd = os.getenv("ASE_AIMS_COMMAND", "aims.x")
```
utf, line 34: What do you think about adding an AIMS_COMMAND option to the atomate2 settings, which is used in preference to the ASE_AIMS_COMMAND?

src/atomate2/aims/run.py
```diff
@@ -0,0 +1,113 @@
...
+    aims_cmd: str
+        The aims command to use
+    """
+    with open("parameters.json") as param_file:
+        parameters = json.load(param_file, cls=MontyDecoder)
+
+    if aims_cmd:
+        parameters["aims_command"] = aims_cmd
+    elif "aims_command" not in parameters:
+        aims_cmd = os.getenv("ASE_AIMS_COMMAND", "aims.x")
```
utf, line 95: If you add the atomate2 AIMS_COMMAND then this should be updated here.

src/atomate2/common/jobs/phonons.py
```diff
@@ -250,115 +249,83 @@ def generate_frequencies_eigenvectors(
...
-        A VaspMaker to use to generate the elastic relaxation jobs.
-    prev_vasp_dir : str or Path or None
-        A previous vasp calculation directory to use for copying outputs.
+    phonon_maker : .BaseVaspMaker or .ForceFieldStaticMaker or .BaseAimsMaker
+        A maker to use to generate dispacement calculations
+    prev_dir: str or Path
+        The previous working directory
+    prev_dir_argname: str
+        argument name for the prev_dir variable
+    socket: bool
```
utf, line 278: Does it make sense having this int he common phonon maker?

utf (commented):
Hi @tpurcell90, thanks again this PR. Overall, it looks great!

I have one potentially unpopular comment. It seems like the ase dependency is only needed to write the input files and run aims. From what I can tell, the actual code to do that is quite short (e.g., looking at the [`AimsTemplate`](https://gitlab.com/ase/ase/-/blob/master/ase/calculators/aims.py) code in ase).

What do you think about converting that code to atomate2? It would then allow us to remove the use of ase `Atoms` (unless that is a desired feature) and would make the `aims` workflows more inline with the rest of atomate2.

If keeping the `ase` components is desirable, I think we should add first class support for ase `Atoms` objects to `monty` itself. Specifically, the `jsanitize` and `MontyEncoder/Decoder` functions. That way, we can remove all the custom `AtomsMSONable` from this PR, and any code can make use of serialising atoms objects. Happy to chat about this further.

tpurcell90:
> I think it would help with repo size and PR review-ability to compress all the `outputs` files

The test data files have now all been compressed, we can also do this for the files generated by the scripts it that is standard practice.

tpurcell90:
> Hi @tpurcell90, thanks again this PR. Overall, it looks great!
> 
> I have one potentially unpopular comment. It seems like the ase dependency is only needed to write the input files and run aims. From what I can tell, the actual code to do that is quite short (e.g., looking at the [`AimsTemplate`](https://gitlab.com/ase/ase/-/blob/master/ase/calculators/aims.py) code in ase).
> 
> What do you think about converting that code to atomate2? It would then allow us to remove the use of ase `Atoms` (unless that is a desired feature) and would make the `aims` workflows more inline with the rest of atomate2.
> 

Hi @utf for the I/O and running aspect of the ASE dependence, ideally we'd strip that out of both ASE and atomate2 and keep that in a separate package run by FHI-aims. This is something that I was thinking about doing for awhile now, but never saw the need for it until working on this PR. We are currently gearing up to do this, but before @ansobolev  and I can write that we need to discuss with the coordinators a bit more first. This can be delayed somewhat because I am moving back to the USA in ~1.5 months.

My general plan was to keep the ASE dependency in place for now and then remove it once the new package was ready, but we can also bring most of that into either atomate2 or pymatgen directly depending on where you think it would be more appropriate.

The biggest question for me would be what to do with the socket calculator since that can reduce the phonon calculation time significantly.

> If keeping the `ase` components is desirable, I think we should add first class support for ase `Atoms` objects to `monty` itself. Specifically, the `jsanitize` and `MontyEncoder/Decoder` functions. That way, we can remove all the custom `AtomsMSONable` from this PR, and any code can make use of serialising atoms objects. Happy to chat about this further.

This would probably be the smarter way of doing it. The `AtomsMSONable` solution was a bit hackish of me to get everything started working.

src/atomate2/common/jobs/phonons.py
tpurcell90: In theory the socket can be used for more codes than just FHI-aims, which is why I have it in common now. However, I see your point where at the moment it is an aims specific keyword.

src/atomate2/aims/run.py
tpurcell90: Ya I can make that change

src/atomate2/aims/run.py
tpurcell90: I'll update this here as well

src/atomate2/aims/files.py
```diff
@@ -0,0 +1,149 @@
...
+        include_files=zipped_files,
+        allow_missing=True,
+        file_client=file_client,
+    )
+
+    logger.info("Finished copying inputs")
+
+
+def write_aims_input_set(
+    structure: MSONableAtoms,
```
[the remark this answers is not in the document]
tpurcell90: I thought I caught all of those in a previous commit, I'll update them here as well.

utf:
> My general plan was to keep the ASE dependency in place for now and then remove it once the new package was ready

That sounds like a good plan to me. No point wasting the effort moving things to atomate2 if it's only going to be removed later.

src/atomate2/aims/files.py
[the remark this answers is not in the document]
tpurcell90: This should be fixed now

src/atomate2/common/jobs/phonons.py
tpurcell90: This is now done

src/atomate2/aims/io/aims_output.py
tpurcell90: I don't disagree, but I was not sure if that should occur now or after the more in-depth changes are made

src/atomate2/aims/jobs/base.py
ansobolev: Yes, this job s not needed, left over from the previous iteration of logic

src/atomate2/aims/flows/gw.py
ansobolev: Because my IDE is too smart, I guess :)

src/atomate2/aims/jobs/base.py
ansobolev: But I need class attributes for the job. After refactoring I'll have to present them as arguments to the function, which will bloat the code and contradict the DRY principle.

src/atomate2/aims/jobs/base.py
utf: Can you see my comment at the top of this class. I think there is a better way to do this

src/atomate2/aims/flows/gw.py
ansobolev: I also didn't like it, it's there just because I was specifically asked for it. Ended up subclassing it from the `ConvergenceMaker` class.

src/atomate2/aims/flows/gw.py
```diff
@@ -0,0 +1,91 @@
...
+            The structure to calculate
+        prev_dir : str or Path or None
+            A previous FHI-aims calculation directory to copy output files from.
+        """
+        parameters = self.input_set_generator.user_parameters
+        parameters["elsi_restart"] = ["read_and_write", 1]
+
+        # the first calculation
+        if all(structure.pbc):
+            input_set = BandStructureSetGenerator(user_parameters=deepcopy(parameters))
```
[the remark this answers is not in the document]
ansobolev: this is rewritten

src/atomate2/aims/flows/gw.py
```diff
@@ -0,0 +1,91 @@
...
+        structure : .MSONableAtoms, Structure, or Molecule
+            The structure to calculate
+        prev_dir : str or Path or None
+            A previous FHI-aims calculation directory to copy output files from.
+        """
+        parameters = self.input_set_generator.user_parameters
+        parameters["elsi_restart"] = ["read_and_write", 1]
+
+        # the first calculation
+        if all(structure.pbc):
```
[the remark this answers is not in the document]
ansobolev: this is rewritten

src/atomate2/aims/jobs/base.py
utf: I don't think it will result in that much code bloat, it just means you will pass 2 more arguments to the function.

src/atomate2/aims/flows/gw.py
utf: I don't have an issue with the concept of the class, just how it has been implemented. I think all the issues can be resolved following the steps I outlined above.

src/atomate2/aims/jobs/base.py
utf: Is this make function actually used somewhere?

src/atomate2/aims/jobs/base.py
utf: I don't think it will result in bloat, it is simply that two more arguments will get added to the function.

ansobolev:
It seems I have addressed all the comments. There were some questions on `ConvergenceMaker` where I did not change the code; the rationale is explained in the comments. If you still want me to rewrite it, please let me know.

src/atomate2/aims/jobs/base.py
ansobolev: >    Can you see my comment at the top of this class. I think there is a better way to do this

Sorry, there is no comment at the top of this class

>    Is this make function actually used somewhere?

Yes, you can only use make for this to work (see eg the test in `test_makers/test_convergence.py`). If you rename `convergence_iteration` to `make`, you'll not be able to replace jobs in the flow, because there will be no flow

tpurcell90: I'll make some changes to clarify what is going on here and then both of you can see if you agree with them

tpurcell90: I modified this in my last commit, I basically moved `convergence_iteration` into a separate job function independent of the `Maker`. I'll now test to see if `convergence_iteration` can be removed entirely with that logic included in make of the Convergence workflow

tpurcell90: I removed the `convergence_iteration` function entirely and moved it to main for the Convergence maker and made that `make` function a job and not a flow.

@utf and @ansobolev do you both agree with the current implementation.

tpurcell90: I kept the `detours` functionality over the `replace` one. Is this okay or should we modify this to use `replace?

src/atomate2/aims/jobs/convergence.py
```diff
@@ -0,0 +1,199 @@
...
+        A name for the convergence criterion. Must be in the run results
+    epsilon: float
+        A difference in criterion value for subsequent runs
+    convergence_field: str
+        An input parameter that changes to achieve convergence
+    convergence_steps: Iterable
+        An iterable of the possible values for the convergence field.
+        If the iterable is depleted and the convergence is not reached,
+        that the job is failed
+    last_idx: int
```
utf, line 44: Is this needed? Can't you just pass a smaller `convergence_steps` iterable if you want to do fewer steps?

src/atomate2/aims/jobs/base.py
ansobolev: Y

> I removed the `convergence_iteration` function entirely and moved it to main for the Convergence maker and made that `make` function a job and not a flow.
> 
> @utf and @ansobolev do you both agree with the current implementation.

Yes, that's ok

src/atomate2/aims/flows/gw.py
```diff
@@ -0,0 +1,32 @@
...
+    Parameters
+    ----------
+    name : str
+        A name for the job
+    maker: .GWMaker
+        A maker for the run
+    criterion_name: str
+        A name for the convergence criterion. Must be in the run results
+    epsilon: float
+        A difference in criterion value for subsequent runs
```
utf, line 26: Can you duplicate the additional docs from `ConvergenceMaker` here? Just to make it easier for people reading the documentation to understand how to use the is class.

src/atomate2/aims/jobs/convergence.py
```diff
@@ -0,0 +1,199 @@
...
+        The index of the last allowed convergence step (max number of steps)
+    """
+
+    name: str = "convergence"
+    maker: BaseAimsMaker = field(default_factory=BaseAimsMaker)
+    criterion_name: str = "energy_per_atom"
+    epsilon: float = 0.001
+    convergence_field: str = field(default_factory=str)
```
utf, line 52: Does it make sense to have default values for this and `convergence_steps`? It seems like these need to be set to make this class useful.

src/atomate2/aims/run.py
```diff
@@ -0,0 +1,114 @@
...
+def run_aims(
+    aims_cmd: str = None,
+):
+    """
+    Run FHI-aims.
+
+    Parameters
+    ----------
+    aims_cmd : str
+        The command used to run FHI-aims (defaults to ASE_AIMS_COMMAND env variable).
```
utf, line 33: Docstring is out of date.

src/atomate2/common/phonon_utils.py
```diff
@@ -0,0 +1,30 @@
+"""Define utilities for the phonon workflow."""
+from phonopy.units import VaspToTHz
+
```
utf, line 3: Can you move this code to `src/atomate2/common/jobs/phonons.py`. Seems like we don't need a whole file for this function.

utf (commented):
Some minor comments. Overall, it's looking very nice.

I think we should go ahead and move the AIMS io code to pymatgen at this point. If we merge it into atomate2 now it will take more effort to move it to pymatgen later on.

src/atomate2/aims/jobs/convergence.py
```diff
@@ -0,0 +1,199 @@
...
+
+            next_job = self.make(
+                structure,
+                prev_dir=next_base_job.output.dir_name,
+            )
+
+            replace_flow = Flow(
+                [next_base_job, update_file_job, next_job], output=next_base_job.output
+            )
+            return Response(detour=replace_flow, output=replace_flow.output)
```
utf, line 128: I think you should change it back to `replace`. We are thinking of removing the `detour` option.

src/atomate2/aims/jobs/convergence.py
```diff
@@ -0,0 +1,199 @@
...
+            if convergence_file.exists():
+                with open(convergence_file) as f:
+                    data = json.load(f)
+                    idx = data["idx"] + 1
+                    # check for convergence
+                    converged = data["converged"]
+        else:
+            prev_dir_no_host = None
+
+        if idx < self.last_idx and not converged:
```
utf, line 96: If removing `self.last_idx` you can just change this to:
```suggestion
        if idx < len(self.convergence_steps) and not converged:
```

src/atomate2/aims/flows/gw.py
ansobolev: done

src/atomate2/forcefields/flows/phonons.py
```diff
@@ -131,191 +117,31 @@ class PhononMaker(Maker):
...
     use_symmetrized_structure: str | None = None
-    bulk_relax_maker: BaseVaspMaker | ForceFieldRelaxMaker | None = field(
+    create_thermal_displacements: bool = True
+    generate_frequencies_eigenvectors_kwargs: dict = field(default_factory=dict)
+    kpath_scheme: str = "seekpath"
+    store_force_constants: bool = True
+
+    code: str = "forcefields"
+    bulk_relax_maker: ForceFieldRelaxMaker | None = field(
         default_factory=lambda: CHGNetRelaxMaker(relax_kwargs={"fmax": 0.00001})
```
JaGeo, line 127:
```suggestion
        default_factory=lambda: CHGNetRelaxMaker(relax_cell=True, relax_kwargs={"fmax": 0.00001})
```

While looking at the code in a different context again: could we add a volume relaxation here. Sorry, probably something I had overlooked.

src/atomate2/aims/jobs/convergence.py
ansobolev: I put them there as the most common use case for the class - tighten the inputs until the energy per atom converges up to 1 meV/atom. However, it makes sense also to remove them completely

src/atomate2/aims/jobs/convergence.py
tpurcell90: removing

JaGeo (commented):
I made one additional comment. I looked at the code for a different purpose again and would appreciate if we could directly fix this as well. Thanks and sorry!

src/atomate2/aims/run.py
tpurcell90: fixing

src/atomate2/aims/jobs/convergence.py
tpurcell90: done

src/atomate2/aims/jobs/convergence.py
tpurcell90: done

src/atomate2/aims/jobs/convergence.py
utf: Sorry for the confusion. I wasn't talking about `criterion_name` or `epsilon`. I just meant that `convergence field` and `convergence_steps` have default values as `field(default_factory=str)` etc. This means that you can technically initialise this class without setting a value for these fields. It could be better to remove the `= field(...)` parts entirely to make sure the user sets the correctly.

tpurcell90:
> I made one additional comment. I looked at the code for a different purpose again and would appreciate if we could directly fix this as well. Thanks and sorry!

No worries, I accepted the requested change let me know if you want me to try to include aims into #581 or include that here depending on what gets merged first

src/atomate2/aims/jobs/convergence.py
ansobolev: ah, ok. Yes, these should be mandatory

JaGeo:
> > I made one additional comment. I looked at the code for a different purpose again and would appreciate if we could directly fix this as well. Thanks and sorry!
> 
> No worries, I accepted the requested change let me know if you want me to try to include aims into #581 or include that here depending on what gets merged first

I fear I will need a bit more time with this PR. Step by step. Let's get this merged first. Then, I can hopefully finish the elastic workflow for the force fields.

JaGeo:
One thing I have just noticed with the stresses from the force field schmema: maybe, we should make sure we always have a `Matrix3D` and not a stress in Voigt notation as this will be hard for common workflows. Might be relevant here as well.

tpurcell90:
For aims we use `Matrix3D` for stress and `list[Matrix3D]` for the atomic stresses. I did not modify the forcefileds ones because I did not want to break things, but I think that makes sense for common workflows

utf:
> I may take a bit longer and do the pymatgen things properly. If it works out then we can strip the atoms dependence outside of the socket io stuff.

That sounds good.

Regardless, I submitted a PR to support serializing ase Atoms objects in materialsvirtuallab/monty#582.

src/atomate2/aims/schemas/calculation.py
```diff
@@ -0,0 +1,494 @@
...
+
+        stress = None
+        if output.stress is not None:
+            stress = voigt_6_to_full_3x3_stress(output.stress).tolist()
+
+        stresses = None
+        if output.stresses is not None:
+            stresses = [
+                voigt_6_to_full_3x3_stress(st).tolist() for st in output.stresses
+            ]
```
JaGeo, line 205: @utf We might need to check units for stresses in general. ase uses eV/Angstrom^3.

tpurcell90: If I recall correctly someone mentioned that atomate2 uses kbar earlier in the PR so in the parsers I added a unit conversion to move from eV/AA^3 to kbar. FHI-aims also uses eV/AA^3 by default.

JaGeo: Great. Thanks. That was me. The force field schema had the wrong unit for stresses. This is why I wanted to make sure.

tpurcell90:
pymatgen PR is now up: https://github.com/materialsproject/pymatgen/pull/3435

Andrew-S-Rosen:
Re: serialization of ASE `Atoms` objects, I don't really see a viable way forward via ASE at this time. It's not going to be viable to patch in an `.as_dict()` and `.from_dict()` doing largely redundant behavior that currently exists in the `Atoms` object. And somewhat understandably, there is friction from the `monty` side as well.

@utf: Could you not do what I've done in `quacc` [here](https://github.com/Quantum-Accelerators/quacc/blob/dffb910a95f92e203ef0043d2ffc5f0cc79c410e/src/quacc/__init__.py#L18-L40)? I've found that has worked great without getting other parties involved. The only thing is you'll only want to do the monkeypatching if ASE is installed since it's not a required dependency, e.g. via `importlib.util.find_spec("ase")`

tpurcell90:
Also is `prev_dir` now standard? If yes I can remove the argument flag from the workflows

tpurcell90:
> EDIT: Sorry, I forgot the original context here. My suggestion doesn't help because the serialized object is represented by a JSON `str` in the database, which can be regenerated into the starting ASE `Atoms` object but isn't query-able or user-friendly.
> 
> Re: serialization of ASE `Atoms` objects, I don't really see a viable way forward via ASE at this time. It's not going to be viable to patch in an `.as_dict()` and `.from_dict()` doing largely redundant behavior that currently exists in the `Atoms` object with the only goal of making it `MSONable`. And somewhat understandably, there is friction from the `monty` side as well.
> 
> @utf: Could you not do what I've done in `quacc` [here](https://github.com/Quantum-Accelerators/quacc/blob/dffb910a95f92e203ef0043d2ffc5f0cc79c410e/src/quacc/__init__.py#L18-L40)? I've found that has worked great without getting other parties involved. You'll only want to do the monkeypatching if ASE is installed since it's not a required dependency, e.g. by checking via `importlib.util.find_spec("ase")`

Once the pymatgen PR is merged in I think I can remove the atoms objects since they are there for IO purposes, and that can be removed with pymatgen io support

tpurcell90:
It looks like the pymatgen PR should be merged once the pipelines pass. I will fix this PR to use the new aims IO interface tomorrow. Is there a procedure to update the pymatgen version inside of a PR?

Andrew-S-Rosen:
@tpurcell90: Yes, you'll want to update the `pyproject.toml` file. Increase the minimum version of pymatgen in the `dependencies` list so that users don't get blindsided with an incompatible version. And then in `strict` under `[project.optional-dependencies]`, set the pinned version of pymatgen to the new version. The `strict` dependencies are the ones used by the CI.

tpurcell90:
Tests pass locally, but I found a few minor bugs in the pymatgen interface that need to be fixed first

ansobolev:
> @ansobolev can you review these changes and make sure you agree
> 
> otherwise I think this is ready now

From the point of view of logic it seems fine; also, the tests pass, which is good. There are some discrepancies in type hinting here and there, but otherwise I also think that it's fine.

tpurcell90:
What are the discrepncies?

src/atomate2/aims/schemas/calculation.py
```diff
@@ -0,0 +1,492 @@
...
+            "direct_bandgap": output.direct_band_gap,
+        }
+
+        forces = None
+        if output.forces is not None:
+            forces = output.forces
+
+        stress = None
+        if output.stress is not None:
+            stress = voigt_to_full_stress_conv(output.stress).tolist()
```
ansobolev, line 198: `output.stress` is `Matrix3D`, but the input to `voigt_to_full_stress_conv` is `Sequence[float]`

ansobolev:
There are some more things that I found; I can clean the code up a bit

src/atomate2/aims/schemas/calculation.py
tpurcell90: This is fixed

src/atomate2/aims/schemas/calculation.py
```diff
@@ -0,0 +1,492 @@
...
+    aims_input: dict[str, Any] = Field(
+        None, description="The aims input parameters used for this task"
+    )
+
+    @classmethod
+    def from_aims_output(cls, output: AimsOutput):
+        """Initialize from AimsOutput object."""
+        return cls(
+            structure=output.initial_structure,
+            species_info=output.data.get("species_info", None),
```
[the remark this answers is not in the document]
tpurcell90: I removed this class since it is never used

ansobolev:
Ok, then I believe it can be merged now

utf (commented):
Thanks for all your hard work on this so far. I have two main comments:

1. I think we should move the input sets to pymatgen. Ultimately, atomate2 will be the home of the infrastructure needed to run the calculations, whereas all input/output parsing and set generation belongs in pymatgen.
2. I gave some suggestions for simplifying the convergence maker.

src/atomate2/aims/sets/base.py
```diff
@@ -0,0 +1,571 @@
+"""Module defining base FHI-aims input set and generator."""
+from __future__ import annotations
```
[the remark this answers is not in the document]
tpurcell90: Okay I can do this, I did not because the other codes also had sets defined here, and also in pymatgen so I was not sure where they belonged. I can fix this early next week when I have access to a properly setup computer again

src/atomate2/aims/jobs/convergence.py
```diff
@@ -0,0 +1,189 @@
...
+                output=next_base_job.output,
+            )
+
+            next_job = self.make(
+                structure,
+                prev_dir=next_base_job.output.dir_name,
+            )
+
+            replace_flow = Flow(
+                [next_base_job, update_file_job, next_job], output=next_base_job.output
```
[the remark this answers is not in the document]
tpurcell90: @ansobolev can you review this?

src/atomate2/aims/sets/base.py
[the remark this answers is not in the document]
utf: Yes, this is my fault. We are in the process of moving the VASP input sets to pymatgen and will work on moving the other codes afterwards.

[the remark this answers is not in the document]
tpurcell90: Okay then I will also make the move

src/atomate2/aims/jobs/convergence.py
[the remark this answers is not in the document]
ansobolev: This approach is nice although I'd prefer to store the current convergence information as the outputs to the last job as well, and not only the inputs of the running job - it's rather counterintuitive to find, at least for me. Also, I like files as they are independent of the DB, easy to look at and not lost when the DB gets overwritten.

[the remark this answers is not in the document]
utf: Ok, I think that should be easily accomplished just by writing the convergence data if there is a prev_dir. This also avoids the need for a separate job. I wrote up some code below:

```python
    @job
    def make(
        self,
        structure: Structure | Molecule,
        prev_dir: str | Path | None = None,
        convergence_data: dict | None = None,
        prev_output_value: float | None = None
    ):
        """Create a top-level flow controlling convergence iteration.

        Parameters
        ----------
        structure : Structure or Molecule
            a structure to run a job
        prev_dir : str or Path or None
            An FHI-aims calculation directory in which previous run contents are stored
        convergence_data : dict or None
            The convergence information to date.
        prev_output_value : float or None
            The output value being converged from the previous aims calculation.
        """
        # getting the calculation index
        idx = 0
        converged = False
        if convergence_data is not None:
            idx = convergence_data["idx"] + 1
            convergence_data["convergence_field_values"].append(self.convergence_steps[idx])
            convergence_data["criterion_values"].append(prev_output_value)
            if len(convergence_data["criterion_values"]) > 1:
                # checking for convergence
                converged = (
                    abs(prev_output_value - convergence_data["criterion_values"][-2])
                    < self.epsilon
                )
        else:
            convergence_data = {
                "criterion_name": criterion_name,
                "criterion_values": [],
                "convergence_field_name": convergence_field,
                "convergence_field_values": [],
            }
        convergence_data.update({"idx": idx, "converged": converged})
        
        if prev_dir is not None and convergence_data is not None:
            # if there is a previous directory and the convergence data is not none
            # then we have done at least one step of the convergence process and
            # we can write the convergence data in the previous directory
            split_prev_dir = str(prev_dir).split(":")[-1]
            convergence_file = Path(split_prev_dir) / CONVERGENCE_FILE_NAME
            with open(convergence_file, "w") as f:
                json.dump(convergence_data, f)

        if idx < len(self.convergence_steps) and not converged:
            # finding next jobs
            next_base_job = self.maker.make(structure, prev_dir=prev_dir)
            next_base_job.update_maker_kwargs(
                {
                    "_set": {
                        f"input_set_generator->user_parameters->"
                        f"{self.convergence_field}": self.convergence_steps[idx]
                    }
                },
                dict_mod=True,
            )
            next_base_job.append_name(append_str=f" {idx}")
            
            next_job = self.make(
                structure,
                prev_dir=next_base_job.output.dir_name,
                convergence_data=convergence_data,
                prev_output_value=getattr(next_base_job.output.output, self.criterion_name)
            )
            
            replace_flow = Flow([next_base_job, next_job], output=next_base_job.output)
            return Response(replace=replace_flow)
        
        task_doc = AimsTaskDoc.from_directory(prev_dir)
        return ConvergenceSummary.from_aims_calc_doc(task_doc)
```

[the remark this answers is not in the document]
ansobolev: Ok, I've rewritten this. The only problem I see here now is that the index in the convergence file will be greater by 1 than the actual index of the calculation, but as this file is not actually used, I guess we can be OK with it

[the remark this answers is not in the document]
utf: Can I just clarify, in the folder with the DFT calculation, would you like the convergence file to contain the convergence info after the calculation has finished or the information without the current calculation? Both should be possible with this approach?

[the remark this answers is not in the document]
ansobolev: Right now the file is added after the calculation has finished, which is fine with me. I don't know if the information without current calculation can be added to the current calculation directory, as it will require prior knowledge of the directory we're running in, and I'm afraid this is unavailable before running the next_base_job. Or is this directory somewhere in the attributes of the next_base_job instance? I looked into the docs, and didn't find it.

tpurcell90:
I created the AimsInputSet pull request here: https://github.com/materialsproject/pymatgen/pull/3482

I left the generators here @utf, should they also move?

utf:
Thanks @tpurcell90. Personally, I think the specific input sets should move too.

tpurcell90:
Right now there is only one `AimsInputSet` class, with multiple generators. I can move all of the generators over to pymatgen, but I am not sure if that makes sense since then if someone wants to add new workflows they'd have to make PRs in both repos. I am okay with this, but I don't know if that is standard.

utf:
Hi @tpurcell90, going forward the aim is to have the core input sets in pymatgen. If someone needs to create a custom input set for a specialised job that isn't broadly of interest then those input sets are fine to live in atomate2. But looking at all the input sets in this PR, I would classify them as "core" input sets and so they should live in pymatgen. 

I hope that makes sense. This is something we are actively trying to balance at the moment so apologies there hasn't been clear guidance.

tpurcell90:
No worries, I might not be clear with my question either should `AimsInputGenerator` also move or just `AimsInputSet`

utf:
Everything in `src/atomate2/aims/sets` should go in pymatgen!

tpurcell90:
`AimsInputSet` is now in pymatgen master. So I need to wait for the next version to be released

JaGeo:
@tpurcell90 You might likely run into this error. I added some hints here: https://github.com/materialsproject/atomate2/pull/693/files

tpurcell90:
Thanks I'll take a look at those once I am back from class!

tpurcell90:
So looking at this the pymatgen change is also in #693 so I think it makes more sense to wait for that to go in then deal with the merge conflicts of the two fixes.

utf:
Hi @tpurcell90, are you able to resolve the remaining conflicts? Then this should be be mergeable.

tpurcell90:
@utf This is now ready, let me know if there are any more final changes you want me to make

utf:
Fantastic, thank you @ab5424 and @tpurcell90!

PR #548, opened 2023-10-02
Update to Pydantic v2

utf:
Final step in the MP ecosystem migration to `pydantic` v2! 🎉

PR #563, opened 2023-10-11
skip validate charge test

jmmshn:
## Fixing defect tests
For #548 

Looks like the pydanctic2 PR is blocked by the defect formation energy test.

This fixes one of the problems:
- The defect charge state sanity check requires the POTCAR data but we only have POTCAR spec so we cannot assign charge automatically.  Prior versions of atomate2 did not give a CI failure for this.  The particular job to check the charge state still failled but the CI is able to pass. 

But there seems to be a second problem that I'm not sure about:
- The`bulk_summary["sc_entry"]` which comes directly from TaskDoc of the relaxation job now gives `None`
Other fields for `bulk_summary` seems to be OK. 

```python
    relax_output: TaskDoc = relax_job.output
    summary_d = {
        "uc_structure": uc_structure,
        "sc_entry": relax_output.entry,
        "sc_struct": relax_output.structure,
```
This creates the remaining error in the CI.  The strange thing here is that `relax_output.structure` is properly parsed but `relax_output.entry` is `None`.

### DIFF
--- a/src/atomate2/common/flows/defect.py
+++ b/src/atomate2/common/flows/defect.py
@@ -200,10 +200,10 @@ class FormationEnergyMaker(Maker, ABC):
         with `relax_radius`.
 
     validate_charge: bool
-        Whether to validate the charge of the defect. If True, the charge of the
-        output structure will have to match the charge of the input defect. This
-        help catch situations where the charge of the output defect is either
-        improperly set of improperly parsed before the data is stored in the
+        Whether to validate the charge of the defect. If True (default), the charge
+        of the output structure will have to match the charge of the input defect.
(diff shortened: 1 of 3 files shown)


janosh:
Thanks @jmmshn! Maybe @munrojm knows why `TaskDoc.from_directory` [gives `None` in your example](https://github.com/materialsproject/atomate2/pull/548#issuecomment-1758572768)?

```py
td = TaskDoc.from_directory("/Users/shen9/repos/mp/atomate2/tests/test_data/vasp/GaN_Mg_defect/bulk_relax/outputs")
print(td) # gives None instead of the entry
```

PR #567, opened 2023-10-12
Update to Pydantic v2

hrushikesh-s:
Final step in the MP ecosystem migration to `pydantic` v2! 🎉


hrushikesh-s:
@Zhuoying & @janosh , it seems like starting a PR to `main` branch also results in just `1 failed test`

PR #574, opened 2023-10-17
`parse_additional_json()` ignore `FW.json.gz` in output directories

janosh:
Ignore `FW.json(.gz)` so `jobflow` doesn't try to parse `prev_vasp_dir` `OutputReferences` from it.

**Reason**
This was causing MP workflows in https://github.com/materialsproject/atomate2/pull/504 to fail with 

```py
ValueError: Could not resolve reference 7f5a7f14-464c-4a5b-85f9-8d11b595be3b not in store or cache
```

if a `FW.json.gz` was present in the job's output directory.

Equivalent `emmet-core` PR: https://github.com/materialsproject/emmet/pull/864


tests/test_data/vasp/Si_mp_meta_gga_relax/r2scan_relax/outputs/FW.json.gz
janosh, line 1: Adding this file tests the above change since `test_mp_meta_gga_double_relax_static` used to fail with `Si_mp_meta_gga_relax/r2scan_relax/outputs/FW.json.gz` present before this change.

PR #577, opened 2023-10-18
Restart capabilities

ansobolev:
This PR is just removing copying of all the outputs to the next dir by default and moving this logic to the base maker. When I tried the writing a restart maker, it appeared that there is no way of finding if the previous calculation converged (well, it has to be, or else the ParseError is thrown). So it will need changing the underlined logic of the parser and the base maker, something that I would not want to do without consulting with you first.


ansobolev:
Sorry, I guess it was the wrong repo :(

PR #583, opened 2023-10-20
Fix Lobster workflow

### LINKED ISSUE #572: BUG: Schema overflow in Lobster workflow
**Describe the bug**
When we run the Lobster workflow for larger structures (unit cells larger than 15 Angstroms in each direction), some of the schema (VASP and Lobster calculation), partially overflow and not all large objects are transferred to the datastore. 
We think that it is connected to the use of the `UniformBandStructureMaker` and the storing the band structures and DOS https://github.com/materialsproject/atomate2/blob/0dad7cbb23745295a8e09d98c3bdec798f8e3898/src/atomate2/vasp/flows/lobster.py#L29
To fix the VASP run, we reduced the kpoint density for a follow-up run. However, then the Lobster schema itself breaks. Here, we still need to investigate further which objects get so big that they need to be part of the datastore or if it is another error. We will try to address the second part within https://github.com/materialsproject/atomate2/pull/529 or a follow-up pull request.

In general, we might need to think about making a robustness test for our schemas in atomate2 (e.g., test larger structures), as this seems to be a common error showing up when the first larger tests are run.

JaGeo:
## Summary

Closes #572 

It's not solving potential related issues with jobflow but a solution for the Lobster workflow. We don't need to save the band structure of DOS from VASP in this case. (Actually, only the BandStructure does not end up in the data store which is something I cannot really understand).

The gzip error was initially resolved by @naik-aakash as a part of his larger PR. However, I would very much appreciate it if this fix could be merged rather quickly.

I also fixed a tiny issue in the documentation that I had forgotten the last time.

### DIFF
--- a/src/atomate2/lobster/jobs.py
+++ b/src/atomate2/lobster/jobs.py
@@ -13,14 +13,18 @@ from pymatgen.io.lobster import Lobsterin
 
 from atomate2 import SETTINGS
 from atomate2.common.files import gzip_output_folder
-from atomate2.lobster.files import LOBSTEROUTPUT_FILES, copy_lobster_files
+from atomate2.lobster.files import (
+    LOBSTEROUTPUT_FILES,
+    VASP_OUTPUT_FILES,
+    copy_lobster_files,
+)
 from atomate2.lobster.run import run_lobster
 from atomate2.lobster.schemas import LobsterTaskDocument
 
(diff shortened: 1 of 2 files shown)


JaGeo:
@Zhuoying or @utf this would be ready to be merged.

utf:
Thanks @JaGeo

PR #585, opened 2023-10-20
Allow prv_dir to be used more in defect wf

jmmshn:
## Completing Allow Previous Bulk Calculations Features
Old code:
- The parsing of the locpot and entry data from an existing bulk SC calculation was not implemented
- The parsing of locpot was not code-independent since the code to get it from the task document still assumes the Vasp format

This PR fixes those issues and adds testing for restarting from existing bulk SC calculation (which is the optimal way to run many defects based on the same bulk SC)


src/atomate2/common/flows/defect.py
```diff
@@ -348,8 +350,8 @@ def make(
         )
 
     @abstractmethod
-    def structure_from_prv(self, previous_dir: str) -> Structure:
-        """Copy the output structure from previous directory.
+    def sc_entry_and_locpot_from_prv(self, previous_dir: str) -> Structure:
```
Zhuoying, line 353: The returned object here should be `ComputedStructureEntry` and `Locpot` as docstring indicates, right? It is not longer `Structure`.

jmmshn: correct

PR #587, opened 2023-10-20
Add setting `VASP_RUN_DDEC6: bool = False`

janosh:
Similar to existing setting `VASP_RUN_BADER`, allows saving DDEC6 charge analysis automatically in any VASP `TaskDoc`.

https://github.com/materialsproject/atomate2/blob/a3215dbe6dcd094eb4dfdff51d77dbdf71530f33/src/atomate2/settings.py#L82

Related PR https://github.com/materialsproject/emmet/pull/869

### DIFF
--- a/src/atomate2/settings.py
+++ b/src/atomate2/settings.py
@@ -84,6 +84,11 @@ class Atomate2Settings(BaseSettings):
         description="Whether to run the Bader program when parsing VASP calculations."
         "Requires the bader executable to be on the path.",
     )
+    VASP_RUN_DDEC6: bool = Field(
+        default=False,
+        description="Whether to run the DDEC6 program when parsing VASP calculations."
+        "Requires the chargemol executable to be on the path.",
+    )
 
     VASP_ZIP_FILES: Union[bool, Literal["atomate"]] = Field(
         "atomate",
--- a/src/atomate2/settings.py
+++ b/src/atomate2/settings.py
@@ -93,7 +93,7 @@ class Atomate2Settings(BaseSettings):
         default=None,
         description="Directory where the atomic densities are stored. If not set, "
         "pymatgen tries to auto-download the densities and extract them into "
-        "~/.cache/pymatgen",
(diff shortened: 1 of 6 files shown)


src/atomate2/settings.py
```diff
@@ -84,6 +84,11 @@ class Atomate2Settings(BaseSettings):
         description="Whether to run the Bader program when parsing VASP calculations."
         "Requires the bader executable to be on the path.",
     )
+    VASP_RUN_DDEC6: bool = Field(
```
Andrew-S-Rosen, line 87: It also requires an environment variable `DDEC6_ATOMIC_DENSITIES_DIR`. I mean, by default it assumes local directory but it's unlikely the user will copy the atomic densities to the local dir if not directed.

janosh: Yeah, we were discussing over at https://github.com/materialsproject/emmet/pull/869#issuecomment-1773481631. I think it would be best not to bother the user with this and just handle it automatically.

Andrew-S-Rosen: I see. Just double-check the licensing that this would be permitted, assuming you mean that you'll ship it.

janosh: Not ship in the `pymatgen` package since it would burden many non-users of `chargemol` to download those files. But once they're in the repo, we have a stable `raw.github.com` URL from where we could `wget` them.

Andrew-S-Rosen: Ah, that's smart. But one thing to note is that not all compute nodes have access to the network, so you might want to have a fallback option.

janosh: Yes, fallback option for no-network situations is on the TODO list.

JaGeo: Oh yes, please. Most computing centers in Europe don't have any outside connection on the compute nodes.

Andrew-S-Rosen: Or at Princeton... 😉

PR #593, opened 2023-10-27
Breaking: Unify previous directory `Maker` API

janosh:
As discussed in today's `atomate2` maintainer meeting, it would be great to unify the API between `Makers` for different codes and forcefields to more easily swap, say, a PBEsol pre-relax maker for a forcefield pre-relax maker and not have to worry about the flow that stitches the makers together passing in an unrecognized keyword like `prev_vasp_dir` into the forcefield maker.

### DIFF
--- a/src/atomate2/amset/jobs.py
+++ b/src/atomate2/amset/jobs.py
@@ -42,7 +42,7 @@ class AmsetMaker(Maker):
     def make(
         self,
         settings: dict,
-        prev_amset_dir: str | Path = None,
+        prev_dir: str | Path = None,
         wavefunction_dir: str | Path = None,
         deformation_dir: str | Path = None,
         bandstructure_dir: str | Path = None,
@@ -54,7 +54,7 @@ class AmsetMaker(Maker):
         ----------
         settings : dict
             Amset settings.
-        prev_amset_dir : str or Path
+        prev_dir : str or Path
             A previous AMSET calculation directory to copy output files from. The
             previous directory is also used to check for transport convergence.
         wavefunction_dir : str or Path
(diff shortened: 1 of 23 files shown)


JaGeo:
Sounds like a good change. Do the forcefield makers need a prev_dir? No files are currently written (should we write some logs?)

In addition:
Should we include a tight-binding based approach to cheaply generate electronic structres?

janosh:
3d255367 shows we could even got a step further and define a `class BaseMaker(Maker, ABC)` that inherits from `jobflow.Maker` and which all other `Maker` subclasses inherit from to enforce a consistent `def make()` method signature. 

That way, it's not up to the PR reviewer to notice violations. `mypy` will automatically flag if a `Maker` has a signature that's incompatible with other makers.

janosh:
I think we may want to hold off on the abstract `BaseMaker` to enforce method signature since `abstractmethods` are covariant, i.e. restrict subclass methods to be more generic than the parent method which is not what we want. We want subclass `Maker.make` to be able to have additional arguments not defined in the parent class. We just want to ensure the arguments `structure` and `prev_dir` are always present. Not sure if there's a good static analysis way of doing this in Python and enforcing it at runtime seems less useful.

janosh:
Either way, here's a shell one line to migrate `atomate2` Flows outside this code base to the new `prev_dir` keyword:

```sh
find ./src -type f -name "*.py" -exec sed -i'' -E 's/prev_(vasp|cp2k|amset)_dir/prev_dir/g' {} +
```

Or to replace any `prev_????_dir`:


```sh
find ./src -type f -name "*.py" -exec sed -i'' -E 's/prev_.+_dir/prev_dir/g' {} +
```

PR #600, opened 2023-11-03
Add `KspacingMetalHandler` to VASP `_DEFAULT_HANDLERS`

janosh:
In the context of https://github.com/materialsproject/custodian/pull/298, @esoteric-ephemera and I noticed that `ScanMetalHandler` (now `KspacingMetalHandler`) is [part of the default handler set in `atomate1`](https://github.com/hackingmaterials/atomate/blob/6327b07ea1ae85fbbe29772777a80dfe4a94dc7b/atomate/vasp/firetasks/run_calc.py#L139-L151) but not used in `atomate2` even though the new MP r2SCAN and all MatPES workflows use `KSPACING`.

This PR adds `KspacingMetalHandler` (requires new `Custodian` release once https://github.com/materialsproject/custodian/pull/298 is merged) to the VASP `_DEFAULT_HANDLERS`. This should be safe to do since we fixed the handler in https://github.com/materialsproject/custodian/pull/298 to not apply fixes on calculations that don't use `KSPACING`.

I also added a test for `_DEFAULT_HANDLERS`.


utf:
Hi @janosh, do you know why the tests are failing?

janosh:
Needs a `custodian` release from @shyuep (please? 😅 )

utf:
Hi @janosh, it seems like the custodian release isn't on Pypi? Are you able to push another one?

utf:
Strange that the right version can't be found...

janosh:
hmmm... `pip install custodian==2024.2.15` works locally. tried both with pip 23.3 and 24 used in CI.

janosh:
i would guess it's a cache issue. 80% sure this PR is safe to merge and will be green on main

utf:
Thanks for double checking!

Issue #584, opened 2023-10-20
BUG: BandStructure not in data store anymore


JaGeo:
@utf I think the bug described in #572  is not related to the workflow being dynamic. I just ran the tutorial example with fewer kpoints
```Python
from atomate2.vasp.flows.core import RelaxBandStructureMaker
from jobflow import run_locally
from pymatgen.core import Structure


mgo_structure = Structure(
    lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],
    species=["Mg", "O"],
    coords=[[0, 0, 0], [0.5, 0.5, 0.5]],
)

bandstructure_flow = RelaxBandStructureMaker().make(mgo_structure)

# run the job
run_locally(bandstructure_flow, create_folders=True)
```

And the band structure ends up in the MongoDB:
![image](https://github.com/materialsproject/atomate2/assets/22094846/cb513c8f-1155-4832-85c2-9cb413e23579)

But again, not the DOS:
![image](https://github.com/materialsproject/atomate2/assets/22094846/6d52c093-ac72-46f6-a331-d5053980f6d5)

I actually don't have any idea where this could come from but it's not a Lobster-workflow related issue alone.

JaGeo:
Could someone else test this independently from me?

utf:
This is so strange and I don't understand why this is happening. It should work to either specify an MSONable class OR a dictionary key. That said, I'm happy to merge the change in #605 while we investigate further.

utf:
Ah, I figured it out! `BandStructure` is not MSONable for some reason.... https://github.com/materialsproject/pymatgen/blob/535f9252c15ea112969f2dba431442027a8a7e14/pymatgen/electronic_structure/bandstructure.py#L134

But `BandStructureSymmLine` is: https://github.com/materialsproject/pymatgen/blob/535f9252c15ea112969f2dba431442027a8a7e14/pymatgen/electronic_structure/bandstructure.py#L676

I think the proper fix is to make `BandStructure` MSONable but lets get this merged for now.

PR #605, opened 2023-11-07
Resolve bandstructure datastore bug

JaGeo:
Closes https://github.com/materialsproject/atomate2/issues/584. 

I added "bandstructure" to the list of object names that should be saved in the datastore. I think it is related to how emmet saves the bandstructures:
https://github.com/materialsproject/emmet/blob/08f143833ac6ba5732abadb7beee9e424d0220cf/emmet-core/emmet/core/vasp/calculation.py#L49

For DOS, I assume it works by accident.

### DIFF
--- a/src/atomate2/vasp/jobs/base.py
+++ b/src/atomate2/vasp/jobs/base.py
@@ -41,7 +41,7 @@ _DATA_OBJECTS = [
     Trajectory,
     "force_constants",
     "normalmode_eigenvecs",
-    "bandstructure",
+    "bandstructure",  # FIX: BandStructure is not currently MSONable
 ]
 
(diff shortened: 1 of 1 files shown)


JaGeo:
It would be great if someone of the maintainers tests this as well as I cannot really write a test for this. Thanks. 

@Zhuoying @janosh ?

JaGeo:
Maybe, we also need to change all names there or directly import the data objects from emmet

utf:
Thanks very much @JaGeo. I just added a note with a reminder to fix BandStructure MSONable later.

JaGeo:
@utf Thanks for responding and resolving it so quickly.

PR #611, opened 2023-11-15
fea: add MACE RelaxMaker and StaticMaker

CompRhys:
## Summary

Adds MACE RelaxMaker and StaticMaker, adds a 2Mb model for testing, could adopt MACE pattern and train a fixture in the test

https://github.com/ACEsuit/mace added to `strict` and `forcefield`

## Checklist

  [ruff](https://ruff.rs).
  type check your code.


src/atomate2/forcefields/jobs.py
```diff
@@ -322,7 +410,6 @@ class GAPRelaxMaker(ForceFieldRelaxMaker):
     relax_cell: bool = False
     steps: int = 500
     relax_kwargs: dict = field(default_factory=dict)
-    optimizer_kwargs: dict = field(default_factory=dict)
```
CompRhys, line 325: deleted as unused, may be passed in somewhere in a flow which may necessitate reverting this.

JaGeo: It is used in Line 82, I think.

CompRhys: will it not inherit from the base class in that instance however?

src/atomate2/forcefields/jobs.py
```diff
@@ -309,6 +309,94 @@ class GAPRelaxMaker(ForceFieldRelaxMaker):
...
+    steps: int = 500
+    relax_kwargs: dict = field(default_factory=dict)
+    task_document_kwargs: dict = field(default_factory=dict)
+    potential_param_file_name: str = "MACE.model"
+    potential_device: str = "cpu"
+    potential_kwargs: dict = field(default_factory=dict)
+
+    def _relax(self, structure: Structure) -> dict:
+        from mace.calculators import MACECalculator
+
```
janosh, line 332: Try 32 or 64 might fix the failing test.

```suggestion
self.potential_kwargs.setdefault("default_dtype", "float32")
```

src/atomate2/forcefields/jobs.py
JaGeo: Yes, sure, but we typically mention those as well in atomate2.

src/atomate2/forcefields/jobs.py
janosh: Depending on what precision `MACE.model` was trained with. [Helped in MBD](https://github.com/janosh/matbench-discovery/commit/467d77746c1f7a445b035a02db97ccc832af7dd1#diff-7c0d3ddd75d72c03759aba8249fe98bd31f16bfafe5945b11c7fa3754deabe61L99).

src/atomate2/forcefields/jobs.py
CompRhys: will put back

tests/forcefields/test_jobs.py
```diff
@@ -66,6 +78,9 @@ def test_m3gnet_static_maker(si_structure):
 
 
 def test_m3gnet_relax_maker(si_structure):
+    # FIXME - brittle due to inability to adjust dtypes in M3GNetRelaxMaker
+    torch.set_default_dtype(torch.float32)
```
janosh: Do we really need these? I think 

```py
        potential_kwargs={"default_dtype": "float64"},
```

should do it.

CompRhys: Couldn't see where the dtypes can be set in chgnet codebase so need to do globally, issue is MACE does it globally to 64 so tests using chg/m3g after mace jobs test were failing

janosh: I see. Pinging @ilyes319 and @davkovacs to see if this can be solved on MACE's side without changing global state?

For now, we'll wrap the test in `try/finally` and revert the `torch.dtype` at the end.

janosh: > It would be great to make it safe to use different types of potential in one go.

Strongly agree. I think we can figure something out with Ilyes and David.

JaGeo: Thanks!

chiang-yuan: Yes dtype conversion has haunted me for a long time...

Issue #607, opened 2023-11-12
BUG: ElasticMaker breaks when fitting tensor


katnykiel:
**Describe the bug**
When I run the following code to calculate elastic constants with VASP:

```python

# Symmetrize the structure
sga = SpacegroupAnalyzer(struct)
struct = sga.get_primitive_standard_structure()

# Get the elastic constants
elastic_flow = ElasticMaker().make(struct)

# Add the workflow to the launchpad
wf = flow_to_workflow(elastic_flow)
lpad.add_wf(wf)
```

I then run `qlaunch rapidfire` and get the following error in the "fit_elastic_tensor" FireWork:

```zsh
Traceback (most recent call last):
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/fireworks/core/rocket.py", line 261, in run
    m_action = t.run_task(my_spec)
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/managers/fireworks.py", line 177, in run_task
    response = job.run(store=store)
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/core/job.py", line 572, in run
    self.resolve_args(store=store)
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/core/job.py", line 678, in resolve_args
    resolved_args = find_and_resolve_references(
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/core/reference.py", line 458, in find_and_resolve_references
    resolved_references = resolve_references(
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/core/reference.py", line 348, in resolve_references
    cache[uuid][index] = store.get_output(
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/core/store.py", line 521, in get_output
    return find_and_resolve_references(
  File "/home/knykiel/.conda/envs/2022.10-py39/atomate2/lib/python3.9/site-packages/jobflow/core/reference.py", line 458, in find_and_resolve_references
    resolved_references = resolve_references(
[...]

PR #623, opened 2023-11-21
[WIP] Equation of state (EOS) workflows

esoteric-ephemera:
## Summary

* Added three families of EOS workflows which inherit from the base classes in `atomate2.vasp.x.EOS.base`, where `x`=`jobs` or `flows`.
  * Base atomate 2 parameters (`base`)
  * Legacy MP parameters in `atomate2.vasp.x.EOS.mp_legacy`: taken [from this paper](https://doi.org/10.1038/s41524-018-0091-x) and used in the atomate 1 implementation (see [here](https://github.com/hackingmaterials/atomate/blob/main/atomate/vasp/workflows/presets/core.py#L564) and [here](https://github.com/hackingmaterials/atomate/blob/main/atomate/vasp/workflows/base/bulk_modulus.py#L21)
  * Current MP-compatible GGA and meta-GGA workflows in `atomate2.vasp.x.EOS.mp`. These represent "tight" calculations but do not strive for the same level of computational complexity as the original atomate workflow, which is less adapted for high-throughput (e.g., k-point grid_density is excessively high, as shown in the SI of the above paper)
* Tests just for the MP GGA jobs and flows

## TODO

* Because the jobs and flows are identical up to config, I'm not sure tests are needed for the base atomate2 and mp_legacy sets/jobs/flows, but I can add them as needed
* Would also love to have suggestions from maintainers regarding structure of the additions and changes for better usability


janosh (commented):
Is there a way to cut down on the 300+ new test files, many of which are 100+ kB in size? Maybe reduce the number of deformations in the reference run used for testing?

Pinging @utf or @Andrew-S-Rosen in case there's an established solution for how to construct tests for many-step workflows.

janosh:
Massive contribution!! Thanks so much @esoteric-ephemera! Super valuable to be able to run EOS with legacy and new params all in `atomate2`. Pinging @mkhorton just so he can revel in the moment. 😆

src/atomate2/vasp/flows/EOS/base.py
janosh, line 1: I would go with `flows/(EOS->eos)/base.py` for consistency since other abbreviations are lower case (see `/vasp`, `/cp2k`, ...).

esoteric-ephemera:
> Is there a way to cut down on the 300+ new test files, many of which are 100+ kB in size? Maybe reduce the number of deformations in the reference run used for testing?
> 
> Pinging @utf or @Andrew-S-Rosen in case there's an established solution for how to construct tests for many-step workflows.

Yeah only one or two should be needed, I was just lazy and put everything there! I'll pare it down today.

JaGeo:
Question from outside: can we somehow generalize the workflows to use them with different DFT codes? I assume they might also be useful in connection with phonon runs (quasi-harmonic). Nothing that needs to be done immediately, of course, but maybe as a long term goal.

esoteric-ephemera:
> Question from outside: can we somehow generalize the workflows to use them with different DFT codes? I assume they might also be useful in connection with phonon runs (quasi-harmonic). Nothing that needs to be done immediately, of course, but maybe as a long term goal.

Yeah absolutely. These basically just need three ingredients: an initial relax maker, a static maker (optional), and a postprocessing step (electronic-structure code-agnostic as of right now). I can try to port these to the common lib

JaGeo:
> > Question from outside: can we somehow generalize the workflows to use them with different DFT codes? I assume they might also be useful in connection with phonon runs (quasi-harmonic). Nothing that needs to be done immediately, of course, but maybe as a long term goal.
> 
> Yeah absolutely. These basically just need three ingredients: an initial relax maker, a static maker (optional), and a postprocessing step (electronic-structure code-agnostic as of right now). I can try to port these to the common lib

This would be awesome! Thanks!

src/atomate2/vasp/flows/eos/base.py
```diff
@@ -0,0 +1,207 @@
...
+        prev_dir : str or Path or None
+            A previous VASP calculation directory to copy output files from.
+        """
+        jobs: dict[str, list[Job]] = {"relax": [], "static": [], "postprocess": []}
+
+        relax_flow = self.relax_maker.make(structure=structure, prev_dir=prev_dir)
+        relax_flow.name = "EOS equilibrium relaxation"
+        equil_props = {
+            "relax": {
+                "E0": relax_flow.output.calcs_reversed[0].output.energy,
```
utf, line 114: If you use `equil_static.output.output.energy` this should be less VASP specific. See below for structure and other instances where you access `calcs_reversed`.

src/atomate2/vasp/flows/eos/base.py
```diff
@@ -0,0 +1,207 @@
...
+            shift = strain_delta / (nzs + 1.0) * np.linspace(-1.0, 1.0, nzs)
+            strain_l[np.abs(strain_l) < 1.0e-15] += shift
+
+        insert_equil_index = np.searchsorted(strain_l, 0, side="left")
+        deformation_l = [(np.identity(3) * (1 + eps)).tolist() for eps in strain_l]
+
+        for ideformation, deformation in enumerate(deformation_l):
+            # Doubly ensure that relaxations are done at fixed volume --> ISIF = 2
+            deform_relax_maker = update_user_incar_settings(
+                flow=self.deformation_maker, incar_updates={"ISIF": 2}
```
utf, line 150: It should be fine to just set this on the deformation maker input set. If users really want to override it then so be it.

This will also help keep the workflow generalisable.

src/atomate2/vasp/jobs/eos/base.py
```diff
@@ -0,0 +1,167 @@
...
+                    "b0 GPa": float(eos.b0_GPa),
+                }
+            except EOSError:
+                output[jobtype]["EOS"][eos_name] = {}
+
+    return output
+
+
+@dataclass
+class DeformationMaker(BaseVaspMaker):
```
utf, line 69: See previous comment about removing `DeformationMaker` in line with the elastic workflow.

src/atomate2/vasp/flows/eos/base.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+            deform_relax_job = deform_relax_maker.make(
```
utf, line 152: Can we copy the logic elastic workflow and generate the deformations in a separate job. This will mean we won't need to implement a `DeformationMaker` for each code (VASP, CP2K, etc). You can just use a regular maker. In the elastic workflow we also write the transformation information to a file in the directory so it easy to track which deformations have been applied.

E.g., see here: https://github.com/materialsproject/atomate2/blob/8746633bf1f597deefc2dd52463bcc3272213cbe/src/atomate2/common/jobs/elastic.py#L103-L163

utf (commented):
This looks awesome. Thanks @esoteric-ephemera. I really appreciate including multiple versions of the workflows for backwards compatibility.

I agree with @JaGeo, it looks like there won't have to be too many modifications to make these generalisable. I added some thoughts as comments.

Regarding the test files, there is an automated script that removes unnecessary files for you. See: https://materialsproject.github.io/atomate2/dev/vasp_tests.html#compile-the-test-data

src/atomate2/vasp/flows/eos/base.py
JaGeo: I would also like to add that ISIF=4 is a common setting for equations of state as well!

src/atomate2/vasp/flows/eos/base.py
esoteric-ephemera: I've more or less copied the elastic workflow logic here. I found during testing that using a function like `atomate2.common.jobs.elastic.run_elastic_deformations` resulted in incorrect flow logic when trying to perform statics on the deformed+relaxed structures. (The postprocessing step would happen simultaneously with the statics, despite it relying on energies/volumes from the statics.)

For that reason, the deformations are applied in an external job, `atomate2.common.jobs.eos.apply_strain_to_structure`, but the relax/static jobs are handled in the flow itself (`atomate2.common.flows.eos.CommonEosMaker`)

esoteric-ephemera:
Finished generalizing and testing the EOS workflows. In addition to the VASP workflows defined previously, I've added default MLFF workflows using the EOS template in `atomate2.common.flows.eos.CommonEosMaker`.

utf:
Hi @esoteric-ephemera, this all looks great.

I had one question about the ` _store_transformation_information` - you mention that this is due to a conflict with emmet/pydantic. Is there an issue we can track in this thread that is preventing removing this option?

esoteric-ephemera:
Thanks @utf! Right the issue persists with pydantic 2.4.2, emmemt-core 0.77.1, and monty 2024.2.2

The ML FF flows run fine with this `_store_transformation_information = True`, but if I run the VASP tests with it set to `True`, I get this stacktrace:
```
File "/Users/aaronkaplan/miniconda3/envs/eos/lib/python3.11/site-packages/jobflow/core/store.py", line 522, in get_output
    return find_and_resolve_references(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aaronkaplan/miniconda3/envs/eos/lib/python3.11/site-packages/jobflow/core/reference.py", line 487, in find_and_resolve_references
    return MontyDecoder().process_decoded(encoded_arg)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aaronkaplan/miniconda3/envs/eos/lib/python3.11/site-packages/monty/json.py", line 561, in process_decoded
    return cls_(**d)
           ^^^^^^^^^
  File "/Users/aaronkaplan/miniconda3/envs/eos/lib/python3.11/site-packages/pydantic/main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 1 validation error for TaskDoc
transformations
  Input should be a valid dictionary [type=dict_type, input_value=<pymatgen.alchemy.materia...e object at 0x2bfc96810>, input_type=TransformedStructure]
    For further information visit https://errors.pydantic.dev/2.4/v/dict_type
```

I haven't looked into it much but I think it's a simple issue with serialization on the monty side. Happy to look into it further (since this will also affect the elastic flow) and/or open an issue about this

utf:
I think we just need to set the type for the transformations field to be `Any` and have done with it! Would you be willing to make that PR?

esoteric-ephemera:
Yep that's been done in [emmet PR #937](https://github.com/materialsproject/emmet/pull/937)!

The interesting thing is that `TaskDoc.transformations` isn't serialized at load. Maybe the serialization occurs when output references are resolved?

utf:
Awesome, thanks @esoteric-ephemera. It's not clear to me what is causing this issue. I wonder whether the recent monty release could help fix this?

esoteric-ephemera:
I don't think so - I'm already on the latest monty == 2024.2.2

Also: just to keep track of things, the same pydantic validation exception is thrown in the elastic flow, as in issue #607

utf:
Thanks @esoteric-ephemera!

PR #638, opened 2023-12-08
Update doc: adding metadata to flow

naik-aakash:
# Addition
When doing a large number of calculations, it would be helpful to be able to add metadata to filter results from the database. Currently I did not find this information available in the atomate2 docs.  (Maybe I am mistaken )

Thus I have added the solution provided [here](https://matsci.org/t/adding-additional-fileds-in-fireworks-docs-with-jobflow-atomate2/45121).

I think it would be nice to have this information readily available in the main atomate2 documentation itself.


naik-aakash:
Hi @janosh , I made this small addition to the documentation which I think should be useful for new users. So kindly review it and if you feel it's fine then this PR could be merged.

docs/user/fireworks.md
```diff
@@ -29,6 +29,14 @@ mgo_structure = Structure(
...
 bandstructure_flow = RelaxBandStructureMaker().make(mgo_structure)
 
+# (Optional) add metadata to the flow task document.
+# Could be useful to filter specific results from the database.
+# For e.g., adding material project ID for the compound, use following lines
+bandstructure_flow.update_maker_kwargs(
+    {"_set": {"task_document_kwargs->additional_fields": {"mp_id": "mp-xxx"}}},
+    dict_mod=True,
+)
+
```
janosh, line 39: Maybe I can learn something here. What are the virtues of `update_maker_kwargs` over something like this?

```
for job in flow if isinstance(flow, Flow) else [flow]:
    job.metadata[key] = val
```

I find this easier to read and remember whereas the above seems easy to get wrong or forget but maybe it has other advantages.

utf: I think we could modify the above code to instead be:

```
from atomate2.vasp.jobs.base import BaseVaspMaker

bandstructure_flow.update_maker_kwargs(
    {"_set": {"task_document_kwargs->additional_fields": {"mp_id": "mp-xxx"}}},
    dict_mod=True,
    class_filter=BaseVaspMaker,
)
```

Since that way only subclasses of `BaseVaspMaker` will get modified.

That said, it probably makes sense to add a powerup to do this automatically.

naik-aakash: Okay, I will then add a power-up and showcase that in the above example.

JaGeo: 👀 Happy Weekend!

naik-aakash:
Hi @utf  and @janosh , I have added powerup as suggested.  Can you please review if and would be happy to address any changes if desired.

I have also added a powerup to disable or use custom custodian handlers for vasp jobs.

src/atomate2/utils/powerups.py
```diff
@@ -0,0 +1,68 @@
...
+    class_filter: .Maker
+        The Maker to which additional metadata needs to be added
+
+    Returns
+    -------
+    Flow
+        Flow with added metadata to the task-doc.
+    """
+    flow.update_maker_kwargs(
+        {"_set": {"task_document_kwargs->additional_fields": additional_fields}},
```
utf, line 32: Thinking about this more, this will override any additional fields already set.

I think this needs modifying slightly to 
```suggestion
        {"_set": {
            f"task_document_kwargs->additional_fields->{k}": v
            for k, v in additional_fields.items()
            }
        }
```

src/atomate2/utils/powerups.py
```diff
@@ -0,0 +1,68 @@
+"""Utilities for modifying workflow."""
```
utf, line 1: I would move these to the `src/atomate2/vasp/powerups.py` directory and set the default class filter to be `BaseVaspMaker`.

src/atomate2/utils/powerups.py
naik-aakash: Oh yes!! thanks for pointing it out. I will update this as per your suggestion 😃

src/atomate2/utils/powerups.py
naik-aakash: Hi @utf , thinking a bit more about the location of this module, maybe is it not better to keep it more general? 

Like it could also be used for adding metadata/modify custodian handlers if for example other types of calculations for example Lobster runs, or cp2k, abinit etc in future?

Just wanted to know your 2nd opinion

utf: We could have a general implementation but then we should include a specific implementation for vasp/cp2k as well. I would also suggest that `atomate2/common/powerups.py` is a better place for the common implementation to live.

Then in `atomate2/vasp/powerups.py` you would do:

```python
from atomate2.common.powerups import add_metadata_to_flow as base_add_metadata_to_flow

def add_metadata_to_flow(...)
    base_add_metadata_to_flow(..., class_filter=BaseVaspMaker)
```

Obviously, you'll need to replace the `...` and add the docstrings.

src/atomate2/utils/powerups.py
naik-aakash: Done

src/atomate2/utils/powerups.py
naik-aakash: Done

utf:
This looks awesome. Thank you so much!

naik-aakash:
> This looks awesome. Thank you so much!

You are most welcome 😃

Issue #646, opened 2023-12-11
BUG: error when supplying LDAUU/J/L through user_incar_settings


JonathanSchmidt1:
I am supplying custom user_incar_settings to a Relax workflow through update_user_incar_settings.
The LDA lines in the INCAR should look like:
> LDAU = True
LDAUJ = 1 0 0
LDAUL = 3 0 0
LDAUTYPE = 2
LDAUU = 5.75 0 0

but come out as :
> LDAU = True
LDAUJ = {'Eu': 1, 'Ti': 0, 'O': 0}
LDAUL = {'Eu': 3, 'Ti': 0, 'O': 0}
LDAUTYPE = 2
LDAUU = {'Eu': 5.75, 'Ti': 0, 'O': 0}

 resulting in a VASP warning and the LDAU/J/L values being ignored. 
 Running the relaxation example ends up with the same behavior:
 ```
from atomate2.vasp.jobs.core import RelaxMaker
from jobflow import run_locally
from pymatgen.core import Structure
print('starting')
# construct an FCC silicon structure
si_structure = Structure(
    lattice=[[0, 2.73, 2.73], [2.73, 0, 2.73], [2.73, 2.73, 0]],
    species=["Eu", "Eu"],
    coords=[[0, 0, 0], [0.25, 0.25, 0.25]],
)

# make a relax job to optimise the structure
relax_m = RelaxMaker()
incar_settings = {"LDAU":True, "LDAUU":{"Eu":5.75}, "LDAUJ":{"Eu":1},"LDAUL":{"Eu":3}}
from atomate2.vasp.powerups import update_user_incar_settings
relax_m_2 = update_user_incar_settings(relax_m, incar_settings)
relax_job = relax_m_2.make(si_structure)
print("starting")# run the job
run_locally(relax_job, create_folders=True)

```

I assume the problem appears due to the last:
```
        # Finally, re-apply `self.user_incar_settings` to make sure any accidentally
        # overwritten settings are changed back to the intended values.
        _apply_incar_updates(incar, self.user_incar_settings)
```
in vasp/sets/base.py

setting skip =["LDAU", ...] for this last update solves the issue for me. However I am not sure if this might have some unintended consequences that I am not aware of. I would also assume that the same problem will appear for every INCAR parameter that is supplied as such a dictionary.
This was tested with version 0.0.12.

utf:
[...]

PR #648, opened 2023-12-12
fix materialsproject/atomate2#646 skip final LDAU/J/L/MAGMOM updates

JonathanSchmidt1:
Skip LDAU/J/L and MAGMOM when applying user settings to final INCAR. Corrected MAGMOMS to MAGMOM

## Summary
LDAU/J/L and MAGMOM tags are skipped when applying the user_incar_settings a final time to the INCAR. This fixes an issue of them being written as a dictionary in the final INCAR. The second commit corrects as spelling mistake that would cause user_incar_settings of MAGMOM to be ignored.

  type check your code.


JonathanSchmidt1:
I assume that no new tests will be needed but test_user_incar_settings() in tests/vasp/test_sets.py would have to be adjusted?
Can someone confirm this?

utf:
Thanks very much for this. It would be good to add a new test. One that simply creates a new input set with user_incar_settings, and then calls `get_input_set()["INCAR"]` and checks that the magmom/LDAU is set correctly.

JonathanSchmidt1:
The LDAU/J/L can be added to the **test_user_incar_settings()** and the MAGMOM result could be changed. Currently, **test_user_incar_settings()** fails because it anticipates MAGMOM as a dictionary in the final INCAR, whereas to my understanding it should actually be formatted as a list in the final INCAR.

I am a bit confused about the input to the failed **test_incar_magmoms_precedence()** .
{"MAGMOM": [3.7, 0.8]} is used as **user_incar_settings** however **_get_magmom()** expects a dict[str, float] .
It appears that this list format might have been a temporary workaround for the issue we are addressing now. Previously, **_get_magmom()** would fail to locate the species in **user_incar_settings** when provided as a list, rather than a dictionary with species as keys. However the final INCAR would be overwritten by the last update, which we are now omitting for MAGMOM. This allowed the test to pass by setting MAGMOM to the list.

To get the test correct I would say the input needs to be changed to the dictionary.
Do you agree or am misunderstanding sth?

JonathanSchmidt1:
@utf any opinions on which if any tests have to be changed/ new ones introduced?
One questions  I have is whether the magmoms in user_incar_settings should be a dictionary as explained in the `VaspInputGenerator` documentation or a list as in the `test_incar_magmoms_precedence` .

utf:
@JonathanSchmidt1, I think you're right that the `test_incar_magmoms_precendence` is wrong. The magmoms should be specified as a dictionary as per the `VaspInputGenerator` docstring.

utf:
Thanks for catching this. I also wouldn't have known that immediately. Would you be able to information about this to the docstring for `user_incar_settings` on the base input set?

utf:
Hi @JonathanSchmidt1, are you able to install the pre-commit by running the following in the root jobflow directory
```
pip install pre-commit
pre-commit install
pre-commit run --all
```
This should fix the failing linting.

JonathanSchmidt1:
> Hi @JonathanSchmidt1, are you able to install the pre-commit by running the following in the root atomate2 directory
> 
> ```
> pip install pre-commit
> pre-commit install
> pre-commit run --all
> ```
> 
> This should fix the failing linting.

Sorry for that, forgot to install it. Should be fixed now.

utf:
Thanks!

PR #650, opened 2023-12-14
Fix import of ASE filters when user has installed from PyPI and not gitlab

esoteric-ephemera:
Simple fix for a recent change to `atomate2.forcefields.utils` which tries to import `FrechetCellFilter` from `ase.filters`. Neither that filter nor the module in ASE exist in the current PyPI release.

I don't know that it's reasonable to expect users to install ASE from gitlab by default. I've added a kwarg to `ase.filters.utils.Relaxer.relax` that fails when `FrechetCellFilter` cannot be imported and the user does not specify and alternate `Filter`.

Added test coverage for `atomate2.forcefields.utils` to ensure higher Codecov


esoteric-ephemera:
I think something is bugged with codecov here. Test coverage of the only file I've modified, `atomate2.forcefields.utils`, is at 82%. It's claiming that `atomate2.common.jobs.utils` has indirect test coverage decrease of -33%, which makes no sense given that these two files are unrelated.

PR #655, opened 2023-12-20
Electrode Workflow

jmmshn:
### Added a general electrode insertion workflow.

Replicated the capabilities from the following wf from atomate1:

https://github.com/hackingmaterials/atomate/blob/6327b07ea1ae85fbbe29772777a80dfe4a94dc7b/atomate/vasp/workflows/base/electrode.py#L20

### Minor change to dev script
Now allows users to specify additional files to copy `--additional_file CHGCAR --additional_file LOCPOT`

TODO:
- [x] tests


src/atomate2/common/jobs/electrode.py
```diff
@@ -0,0 +1,237 @@
...
+    """A summary of a relaxation job."""
+
+    structure: Structure
+    entry: ComputedEntry
+    dir_name: str
+    uuid: str
+
+
+@job
+def get_stable_inserted_structure(
```
utf, line 35: Does this need to be a job? Could all the contents of this function be put into `ElectrodeInsertionMaker.make`?  Since everything inside this is a job anyway I think that would work.

utf (commented):
Thanks @jmmshn, just one thought.

src/atomate2/common/jobs/electrode.py
jmmshn: Since the job calls itself recursively I thought it would have to be wrapped in a job.
And I think it has to do this recursively since it doesn't know:
- How many insertions are needed
- How many insertions are topotactic

Side note. I know if `replace` or `addition` is the right thing to use here.  They both seem OK but I don't know if there is some guide to follow.

utf: Makes sense, I missed the recursion. `replace` is the right way to go here as that way you can use the outputs of the workflow in another job (see my other comment).

src/atomate2/common/flows/electrode.py
```diff
@@ -0,0 +1,141 @@
...
+        [get_stable_inserted_structure]
+        [get_stable_inserted_structure]
+        [get_stable_inserted_structure]
+        ... until the insertion is no longer topotactic.
+
+    This workflow requires the users to provide the following functions:
+        self.get_charge_density(task_doc: TaskDoc):
+            Get the charge density of a TaskDoc output from a calculation.
+        self.update_static_maker():
+            Ensure that the static maker will store the desired data.
```
utf, line 46: Is there a paper or anything that documents this further?

jmmshn: Yeah link is here, I'll add it to the PR.

https://www.nature.com/articles/s41524-020-00422-3

src/atomate2/common/jobs/electrode.py
```diff
@@ -0,0 +1,237 @@
...
+        insertions_per_step=insertions_per_step,
+        n_steps=nn_step,
+        n_inserted=n_inserted + 1,
+    )
+
+    for job_ in [static_job, chg_job, insertion_job, min_en_job, relax_jobs, next_step]:
+        job_.append_name(f" {add_name}")
+
+    replace_flow = Flow(
+        jobs=[static_job, chg_job, insertion_job, relax_jobs, min_en_job, next_step]
```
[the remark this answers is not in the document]
jmmshn: Good point.  I hadn't thought about that since the aggregation pipeline for this stuff is already pretty powerful so most of the analysis and "thinking" was relegated to that and this WF is really just a buck-shot approach to get a bunch of structures into the DB.
But there is good value in getting a document out for some simple applications.  I can add that to the PR.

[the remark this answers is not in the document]
jmmshn: Added some useful document models to the output.
Requires this PR from emmet for serialization to work:

https://github.com/materialsproject/emmet/pull/922

jmmshn:
So it looks like the different document models required are more traditional `mp_id` in the `mp-###` or `###` format: 

https://github.com/materialsproject/emmet/blob/e1896190d909a40c291a5563251b4e58bef30525/emmet-core/emmet/core/structure_group.py#L281C1-L288C1

I don't think I can use FW_ID since it'll be manager-dependent.

utf:
The idea with jobflow/atomate2 is that we'd be moving away from using sequential mp_ids as these cause an issue when merging two databases. I'd prefer to stick with UUIDs in atomate2/jobflow, and leave the mp_id generation up to the MP ingestion pipeline.

jmmshn:
> It seems like we're running into an issue of the scope of emmet. Is it solely the build pipeline for MP or is it also a definition of general schemas for materials chemistry. Ideally, the emmet models could function without mp_ids present. E.g., could we do some other sorting of the UUID if it doesn't match the mp format?

Would it be possible to use UUID1 instead of UUID4 that way there is a time component encoded in the data.

jmmshn:
This discussion seems related:
https://github.com/materialsproject/jobflow/issues/519

So the idea that the UUIDS should be sortable in time might be something to disucss.

jmmshn:
Thanks, @utf think something like this ca be added with UUID1:


```python
import datetime
import uuid
uuid_arg = uuid.uuid1()
DATETIME_EPOC = datetime.datetime(1970, 1, 1)
timestamp = (uuid_arg.time - 0x01B21DD213814000) / 1e7
dt = DATETIME_EPOC + datetime.timedelta(seconds=timestamp)
print(dt)
```

utf:
Hi @jmmshn, I've updated the emmet version and monty version. Looks like the remaining failing tests are related to this PR.

jmmshn:
This should be fixed, the `ulid` import was missing should be fixed now.

utf:
@jmmshn just this comment to address:

> can you also add the ulid dependency to the "defects" class of optional dependencies

jmmshn:
oops. Misses a save on the last commit.

utf:
Thanks @jmmshn!

jmmshn:
@utf, ooops! I updated the pyproject.toml in the other open PR instead.  Thanks for fixing!

Issue #672, opened 2024-01-10
BUG: Error messages in atomate2 r2SCAN workflow, using the default input set generators


abhardwaj73:
While using MPMetaGGADoubleRelaxMaker in atomate2, the following 2 errors have appeared in all 3 of my runs:

Although `custodian` handles them and updates the files (and the workflow finishes), is this something that can be added directly in the default settings of this workflow? Especially reading the WAVECAR part, as `custodian` deletes the WAVECAR file to resolve the issue. As the WAVECAR is quite big, and not used in the r2SCAN relax step, can it be avoided in printing?

1. Use the atomate2 workflow: MPMetaGGADoubleRelaxMaker and default input file settings
2. Use the following structure

```
# construct an FCC silicon structure
si_structure = Structure(
    lattice=[[0, 2.73, 2.73], [2.73, 0, 2.73], [2.73, 2.73, 0]],
    species=["Si", "Si"],
    coords=[[0, -0, -0], [0.25, 0.25, 0.25]],
)
```

3. See the error in the `MP meta-GGA relax 2` step (and `custodian`'s involvement)

```
while reading WAVECAR
ERROR:custodian.custodian:VaspErrorHandler
ERROR:VaspErrorHandler:ALGO=A and IALGO=5X tend to fail
ERROR:custodian.custodian:VaspErrorHandler
2024-01-10 07:56:23,510 INFO Finished job - MP meta-GGA relax 2 (372e5f3c-884a-41a1-972c-c49c910d2c37)
```
Here is the custodian.json file for this step, which tells the action it has taken:

```
[
    {
        "job": {
            "@module": "custodian.vasp.jobs",
            "@class": "VaspJob",
            "@version": "2023.10.9",
            "vasp_cmd": [
                "srun",
                "/home/ab6989/MPScanRelaxSet/atomate2/vasp_std"
            ],
            "output_file": "vasp.out",
            "stderr_file": "std_err.txt",
            "suffix": "",
            "final": true,
            "backup": true,
            "auto_npar": false,
            "auto_gamma": true,
            "settings_override": null,
            "gamma_vasp_cmd": [
                "vasp_gam"
            ],
            "copy_magmom": false,
            "auto_continue": false
        },
        "corrections": [
            {
                "errors": [
                    "coef"
                ],
                "actions": [
                    {
                        "file": "WAVECAR",
                        "action": {
                            "_file_delete": {
                                "mode": "actual"
                            }
                        }
                    }
                ],
[...]

PR #673, opened 2024-01-10
Ensure MP VASP sets don't use auto_ismear, few other fixes

esoteric-ephemera:
In response to issue [#672](https://github.com/materialsproject/atomate2/issues/672), all MP VASP input sets now set `auto_ismear = False` by default. This change leads to the following behavior:
- GGA workflow ***always*** uses tetrahedron smearing (ISMEAR = -5, SIGMA = 0.05). This is consistent with the atomate GGA workflow and left for consistency
- meta-GGA workflow uses Gaussian smearing (ISMEAR = 0, SIGMA = 0.05) for relaxations and tetrahedron smearing (ISMEAR = -5, SIGMA = 0.05) for the final static. This is not quite the behavior from the atomate workflow, but is a minor change given that the atomate wf didn't include a static or CHG/WAVECAR initialization like the atomate2 one does
- The meta-GGA workflow behavior is consistent with the wf structure moving forwards with MP
- Updated test files to reflect these changes

Few other fixes/improvements:
- Removed dependence on `pkg_resources.resource_filename` as this has been deprecated in favor of `importlib.resources.files`
- Froze the MACE model used in the elastic forcefield workflow ***test*** - I saw that the test could fail ([link](https://github.com/materialsproject/atomate2/actions/runs/7474432469/job/20340565788)) when the model couldn't be downloaded on the fly


mkhorton:
Asking to clarify: is this issue in relation to the MP GGA (PBE) workflow, or the MP GGA (PBEsol) workflow used to bootstrap the R2SCAN calculation?

If the former, I thought the historical behaviour was to use Gaussian smearing (ISMEAR=0) as a sensible default for relaxations, and tetrahedron only for statics?

esoteric-ephemera:
It ends up being both workflows (not just the PBEsol pre-relax) because the value of ISMEAR changes according to bandgap. We don't necessarily want Methfessel-Paxton smearing in case a system is incorrectly predicted to be a metal in the GGA pre-relax. 

> If the former, I thought the historical behaviour was to use Gaussian smearing (ISMEAR=0) as a sensible default for relaxations, and tetrahedron only for statics?

I recall looking into the atomate workflows and finding that that the `OptimizeFW` defaults to `MPRelaxSet` with no further changes to ISMEAR. That ends up giving `ISMEAR = -5` in relaxations. 

I've also done some sanity checks against the MP task docs, and it looks like `ISMEAR = -5` was always used in GGA relaxations.

utf:
Thanks @esoteric-ephemera. Can I just ask about the motivation for disabling `auto_ismear` - is that for consistency with atomate1/pymatgen input sets or is there another reason?

esoteric-ephemera:
Hey @utf, for the GGA workflow, it's consistency with atomate. To be consistent with atomate, the meta-GGA workflow would have to start with Methfessel-Paxton smearing (ISMEAR = 2, SIGMA = 0.2) and then switch to tetrahedron (ISMEAR = -5, SIGMA = 0.05) if the gap in the PBEsol step is nonzero

I will say I'm in the process of running the "new" (updated pseudopotentials, parameters and hopefully +U's) workflows for both meta-/GGA workflows. So if we'd rather have these exist just as their atomate 1 equivalents, I'm happy to make that change

utf:
Understood! I think this is ok for now. Thanks very much.

mkhorton:
Thanks @esoteric-ephemera, not the first time my memory has been incorrect :)

Andrew-S-Rosen:
FYI, the MACE model downloading issue is fixed in `main` of the MACE repo.

Andrew-S-Rosen:
@esoteric-ephemera --- with this change, is ISMEAR always equal to 0 and SIGMA always equal to 0.05 in the MP meta-GGA relax and pre-relax steps (unless changed by Custodian)? I'm not opposed -- I just want to make sure I'm understanding this correctly.

esoteric-ephemera:
Yeah for the pre-relaxation , ISMEAR = 0 and SIGMA = 0.05. The r2SCAN relax continues with 0.22 (metal) <= SIGMA <= 0.44 (wide-gap insulator) and ISMEAR = 0. The final r2SCAN static uses ISMEAR = -5.

This is mostly to avoid situations where Methfessel-Paxton (ISMEAR > 0) smearing is used to relax insulators, or tetrahedron (ISMEAR = -5) method is used to relax metals. The atomate 1 r2SCAN flow had the former problem, the PBE flow has the latter problem

Andrew-S-Rosen:
Thanks!! For clarification: presumably you meant KSPACING instead of SIGMA for the 0.22 to 0.44 comment.

esoteric-ephemera:
Whoops yes thanks! Corrected above

PR #674, opened 2024-01-11
add directory of task doc generation to phonon schema

JaGeo:
## Summary

Currently, the path to the folder where the task doc and phonopy output files have been generated is not added to the list of directories of the output of the phonon run. This makes collecting all data unnecessarily hard. I have now added this directory path as well.


JaGeo:
Failures come from some MACE stuff... Just as on the main branch.

utf:
Thanks! I think the MACE issue is fixed in #673

PR #689, opened 2024-01-24
Porting qchem into atomate2

rdguha1995:
## Summary

This is the same [PR](https://github.com/materialsproject/atomate2/pull/161) that was merged and reverted - For @utf 's perusal.

Thanks in advance @utf for reviewing this. Please let me know if I can change anything from my end. In the meantime I would look for ways to increase test coverage in this PR.


tests/qchem/test_drones.py
```diff
@@ -0,0 +1,7 @@
+from atomate2.qchem.drones import QChemDrone
+
+
+def test_structure_optimization(qchem_test_dir):
+    drone = QChemDrone()
+    doc = drone.assimilate(qchem_test_dir / "water_single_point" / "outputs")
+    assert doc
```
janosh, line 7: maybe assert some expected attributes on doc?

src/atomate2/qchem/run.py
```diff
@@ -0,0 +1,154 @@
...
+    #     vasp_job_kwargs["auto_npar"] = False
+
+    # vasp_job_kwargs.update({"gamma_vasp_cmd": split_vasp_gamma_cmd})
+
+    # if job_type == JobType.DIRECT:
+    #     logger.info(f"Running command: {qchem_cmd}")
+    #     # return_code = subprocess.call(qchem_cmd, shell=True)
+    #     return_code = subprocess.call(qchem_cmd)
+    #     logger.info(f"{qchem_cmd} finished running with returncode: {return_code}")
+    #     return
```
utf, line 92: Can this be removed?

src/atomate2/qchem/run.py
```diff
@@ -0,0 +1,154 @@
...
+        jobs = [
+            QCJob(
+                split_qchem_cmd, max_cores=SETTINGS.QCHEM_MAX_CORES, **qchem_job_kwargs
+            )
+        ]
+    else:
+        raise ValueError(f"Unsupported job type: {job_type}")
+
+    # if wall_time is not None:
+    #     handlers = list(handlers) + [WalltimeHandler(wall_time=wall_time)]
```
utf, line 104: And this?

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,486 @@
...
+            "radii": "uff",
+            "theory": "cpcm",
+            "vdwscale": "1.1",
+        }
+
+        plots_defaults = {"grid_spacing": "0.05", "total_density": "0"}
+
+        opt_dict = {} if self.opt_variables is None else self.opt_variables
+
+        scan_dict = {} if self.scan_variables is None else self.scan_variables
```
utf, line 281: My advice would be to turn this into a `dataclass`. You can use `__post_init__` to do any validation, etc. As I see it the benefits are:
- It will allow you to reduce the amount of repetitive initialise code, e.g., `pcm_dict = pcm_dict or {}`.
- It will bring the input set classes inline with VASP, CP2K, FHI-aims.
- The subclasses for specific calculations will be simplified substantially.

src/atomate2/qchem/sets/base.py
```diff
@@ -0,0 +1,486 @@
...
+
+    #     Returns
+    #     -------
+    #     dict
+    #         A dictionary of updates to apply.
+
+    #     Is this something which is even necessary
+    #     in context of QChem? Will have to discuss
+    #     """
+    #     raise NotImplementedError
```
utf, line 486: This can be removed?

src/atomate2/qchem/sets/core.py
```diff
@@ -0,0 +1,62 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class SinglePointSetGenerator(QCInputGenerator):
+    """Generate QChem Single Point input sets."""
+
+    def __init__(self) -> None:
+        super().__init__(job_type="sp", scf_algorithm="diis", basis_set="def2-tzvppd")
```
utf, line 18: These are currently data classes but don't take advantage of the dataclass paradigm. E.g., they implement their own `__init__` function. Instead, I'd recommend making the `QCInputGenerator` class a dataclass first, and then subclasses can be defined as:

```suggestion
@dataclass
class SinglePointSetGenerator(QCInputGenerator):
    """Generate QChem Single Point input sets."""
    
    job_type: str = "sp"
    algorithm: str = "diis"
    basis_set: str = "def2-tzvppd"
```

Which to me is immediately more obvious to see what has been changed.

utf (commented):
Hi @rdguha1995, this is looking in great shape. 

My main comment is on the format of the input sets. I think these can be simplified as data classes with very little effort.

src/atomate2/qchem/sets/core.py
rdguha1995: Thanks for these suggestions @utf ! These were really helpful. I was not aware of the `__post_init__()` functionality  in `dataclasses`. It makes the implementation much more concise.

I refactored the `InputSetGenerator` and they are now consistent with the `InputSetGenerator` in `vasp` or `cpk`.

Adding a few more tests in `drones` and `sets` should push the codecov beyond the target. I will try to get to them soon. Let me know if anything stands out to you in the current implementation for the `InputSetGenerator`

utf:
Hi @rdguha1995, thanks for making these changes. The input sets look very nice.

Would you be happy to move all of the input sets to pymatgen? We are doing the same with the VASP, FHI-aims, etc input sets too.

janosh:
so it looks like the release failed because the `test.yml` and `release.yml` workflows use different setups. the latter install `openbabel` through `miniconda`, the former doesn't. as a result only a subset of tests are running on PRs but the full test suite when trying to make release. not sure when it became this way

rdguha1995:
@utf Of Course! Thanks for your help. 

Coming to the `pymatgen` question, I am not opposed to migrating the `qchem input sets`, but I don't think it is the right time for it. Tagging in @rkingsbury on this for his perspective. `qchem` is still primarily driven in `atomate(1)`. The `VASP` workflows in `atomate2` have been used, tested and re-tested over time while we have not done that (at all) for `qchem-atomate2`. I think we should keep the `qchem` `InputSet`s in the `atomate2` sandbox for now and once we adopt it more regularly, we should migrate them to `pymatgen`

rkingsbury:
> @utf Of Course! Thanks for your help.
> 
> Coming to the `pymatgen` question, I am not opposed to migrating the `qchem input sets`, but I don't think it is the right time for it. Tagging in @rkingsbury on this for his perspective. `qchem` is still primarily driven in `atomate(1)`. The `VASP` workflows in `atomate2` have been used, tested and re-tested over time while we have not done that (at all) for `qchem-atomate2`. I think we should keep the `qchem` `InputSet`s in the `atomate2` sandbox for now and once we adopt it more regularly, we should migrate them to `pymatgen`

I agree! Let's kick the tires a bit with the full stack in `atomate2` before porting things over.

utf:
Ok, that sounds reasonable for now.

If you can add a test for the drone, I'd be happy to merge!

shyuep:
I would suggest rather than porting over to pymatgen at a later date, to just do the dev of input sets in pymatgen from the get go. I am puzzled why a two stage thing is necessary.

rdguha1995:
> Ok, that sounds reasonable for now.
> 
> If you can add a test for the drone, I'd be happy to merge!

Hey @utf , I added in tests for all the sections of the code which were missing coverage. I think it should be ready to merge now. Let me know if I can help with anything!

Thanks for all the help.

janosh (commented):
i think user experience would benefit from having more specific maker names

src/atomate2/qchem/jobs/core.py
```diff
@@ -0,0 +1,236 @@
...
+        Keyword arguments that will get passed to :obj:`.should_stop_children`.
+    write_additional_data : dict
+        Additional data to write to the current directory. Given as a dict of
+        {filename: data}. Note that if using FireWorks, dictionary keys cannot contain
+        the "." character which is typically used to denote file extensions. To avoid
+        this, use the ":" character, which will automatically be converted to ".". E.g.
+        ``{"my_file:txt": "contents of the file"}``.
+    """
+
+    name: str = "force"
```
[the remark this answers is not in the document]
rdguha1995: Happy to change these and other simplistic names. Just wanted to mention here that the names were adapted from the names of corresponding fireworks from atomate(1).

Is name consistency something to keep in mind here?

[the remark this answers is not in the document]
janosh: ah didn't realize. i would say probably not important to stay consistent assuming the schemas are already different? i.e. you couldn't add to an existing atomate1-database with these workflows, right?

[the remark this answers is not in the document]
rdguha1995: No I don't think backwards compatibility is a thing. At all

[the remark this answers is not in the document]
rdguha1995: I will go ahead and rename them as `QCForce` and so on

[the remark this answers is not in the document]
rdguha1995: Thinking about this some more. The users would import these makers from `atomate2.qchem.jobs`. They will already have context that these are `qchem` calculations. Another round of modification seems unnecessary?

src/atomate2/qchem/run.py
```diff
@@ -0,0 +1,139 @@
...
+
+        - ``DIRECT``: Run QChem without using custodian.
+        - ``NORMAL``: Normal custodian :obj:`.QCJob`.
+    """
+
+    DIRECT = "direct"
+    NORMAL = "normal"
+
+
+def run_qchem(
```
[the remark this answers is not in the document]
rdguha1995: Thanks for catching this. Added in some tests for increasing coverage here

rdguha1995:
Hey @utf, 

I think this should be good to merge now :-) . Let me know if there's any feedback and I will incorporate it.

rdguha1995:
> Hey @utf,
> 
> I think this should be good to merge now :-) . Let me know if there's any feedback and I will incorporate it.

Hey @utf , bumping this up. Let me know if there's anything left from our end. I will try to incorporate that and then get to the writing as well.

utf:
Hi @rdguha1995, I think it is worth addressing the point by @shyuep

> I would suggest rather than porting over to pymatgen at a later date, to just do the dev of input sets in pymatgen from the get go. I am puzzled why a two stage thing is necessary.

I agree that avoiding the two stage approach would be beneficial - we made this mistake with VASP and it caused difficulties down the line. The VASP input sets in pymatgen are now compatible with both atomate1 and atomate2. What are the main blockers for achieving this for QChem?

utf:
Hi @rdguha1995, there are still a lot of people that use VASP with atomate1 - indeed many in Materials Project are still using atomate1. The new input sets in pymatgen are compatible with both atomate1 and atomate2. Part of this was ensuring that that the settings themselves stayed the same after the refactor, and instead the only thing that changed was that the implementation itself was updated.

For QChem, is the plan that the new input sets give the same settings as the old input sets? If so, is the issue to do with whether the molecule is needed on input set initialisation (this would be the same as for VASP)? In that case, it should be possible to update the new input sets so that the API is unchanged but there is a new implementation under-the-hood - I'd be happy to chat about how to achieve this. If the input settings are different between the new and old input sets, then this would be a different issue and perhaps both could live in pymatgen side-by-side?

Part of the issue of a two stage approach is deciding when to put the input sets back into pymatgen. Will this only happen after people stop using atomate1? If so, this could be a very long time. If not, then the new input sets will have to be made compatible with atomate1 anyway, so we may as well do it now before they start diverging further.

rdguha1995:
Hey @utf, I agree that it might be a long time if we don't move the sets over now. I can do some local tests now and see what needs to change before we move it over.

rkingsbury:
Hi everyone, I wanted to revive this discussion because I am quite interested in having my student test out these workflows, but they will have an easier time if they can `pip install` (they're not too familiar with git yet). They look in great shape to me.

Regarding the `InputSet`, I agree that they should eventually move into `pymatgen` and I think they can be made compatible with `atomate1` (based on a cursory look a long time ago). I think that's best left for another PR though. We could always add a warning or `deprecationwarning` to the input sets here to signal to users that this is a temporary home if we wanted. Functionally speaking though, an `atomate2` user would never directly import that `InputSet` anyway, right? They are going to use the Job / Flow, so the migration could happen behind the scenes.

PR #690, opened 2024-01-26
Prep for next `pymatgen` release

janosh:
Many `pymatgen` classes had inconsistently named `get/from_(str|string)` methods. `get/from_string` was deprecated in favor of `get/from_str` 6 months ago and removed last week in https://github.com/materialsproject/pymatgen/pull/3561 which will be released this or next week.

@utf this change will require a new `atomate2` version to be compatible with the upcoming `pymatgen`.

resulting `AttributeError` incurrent `atomate2` reported by @abhardwaj73 in https://github.com/materialsproject/pymatgen/issues/3571#issuecomment-1911066021. thanks!

### DIFF
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -94,11 +94,7 @@ class VaspInputSet(InputSet):
         if make_dir:
             os.makedirs(directory, exist_ok=True)
 
-        inputs = {
-            "INCAR": self.incar,
-            "KPOINTS": self.kpoints,
-            "POSCAR": self.poscar,
-        }
+        inputs = {"INCAR": self.incar, "KPOINTS": self.kpoints, "POSCAR": self.poscar}
         inputs.update(self.optional_files)
 
         if isinstance(self.potcar, Potcar):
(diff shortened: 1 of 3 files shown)


JaGeo:
@mkhorton we have been installing atomate2 with [strict] for quite a while now as there were many issues because of the pydantic upgrade and incompatibilties with other MP codes. I think quite a few users have been doing this since then. Thus, the effect of this incompability should not be too big.

janosh:
@JaGeo i went the other extreme and have been installing using `atomate2` from `main` these last couple of months 😄 
which incidentally is why i didn't notice this fix hasn't been released yet. happy to help with that

utf:
@mkhorton thanks for raising this - unfortunately, the latest pymatgen had some bugs which meant it was not compatible with atomate2. Once a fixed version is released I can push a new atomate2 version.

Currently, I'm the only one pushing releases but technically @janosh is able to do it also (as can anyone with push access to main). If pushing atomate2 versions is something @janosh would be interested in, then happy to chat about it over email.

utf:
Also adding this for tracking: #719

Issue #654, opened 2023-12-19
BUG:ElasticMaker breaks when conventional=True


tlsong:
**Describe the bug**
When I run the following code to calculate elastic constants with VASP:
```
s=Structure.from_file("Si_POSCAR")
s_prim = SpacegroupAnalyzer(s).get_primitive_standard_structure()
flow = ElasticMaker(name="Si_conPOSCAR_elastic").make(s_prim,conventional=True)
run_locally(flow, create_folders=True)
```

get the following error：
```
  File "/home/xxx/test_Si/run_elastic_atomate2.py", line 11, in <module>
    flow = ElasticMaker(name="Si_conPOSCAR_elastic").make(s_prim,conventional=True)
  File "/home/xxx/software/miniconda3/envs/atomate2/lib/python3.10/site-packages/atomate2/common/flows/elastic.py", line 119, in make
    sga = SpacegroupAnalyzer(structure, symprec=self.symprec)
  File "/home/xxx/software/miniconda3/envs/atomate2/lib/python3.10/site-packages/pymatgen/symmetry/analyzer.py", line 83, in __init__
    for species, group in itertools.groupby(structure, key=lambda s: s.species):
  File "/home/xxx/software/miniconda3/envs/atomate2/lib/python3.10/site-packages/jobflow/core/reference.py", line 252, in __iter__
    raise TypeError("OutputReference objects are not iterable")
TypeError: OutputReference objects are not iterable
```

Installed the latest version of atomate2 with the following command: 
`pip install atomate2[strict]`

tlsong:
The following code can run normally
```
s=Structure.from_file("Si_POSCAR")
s_prim = SpacegroupAnalyzer(s).get_primitive_standard_structure()
flow = ElasticMaker(name="Si_POSCAR_elastic").make(s_prim)
run_locally(flow, create_folders=True)
```

mjwen:
Hi @tlsong Sorry for the late response! 

I confirm that I can reproduce the bug. We will look into it and update you soon.

PR #706, opened 2024-02-08
Fix elastic conventional structure

mjwen:
## Summary

- Fix the bug in #654
- Update test data to use those generated from conventional cells
- Use the new test data to test both conventional and unconventional cells 

## TODO (if any)
- ~~The elastic workflow cannot finish no matter `conventional` is set to `True` or `False`. There seems to be another bug (see #707). Let's wait for that to be corrected first.~~ (It is a pymatgen bug, and fix in https://github.com/materialsproject/pymatgen/pull/3617)
- ~~add a test for `conventional`~~ (done)


JaGeo:
Maybe interesting: https://github.com/materialsproject/atomate2/blob/e5ab3010d3441dba2681c92c1b86edebb34fce77/src/atomate2/common/jobs/utils.py#L38

utf:
Thanks very much @mjwen

tlsong:
Thanks very much @mjwen

PR #708, opened 2024-02-09
Fix import error [ijson]

### LINKED ISSUE #702: [BUG] `ijson` is an optional import
I think `ijson` is an optional import for lobster but the lobster import is a part of the mp flow. 

https://github.com/materialsproject/atomate2/blob/8b1ee044674bb75ee9cd025d9ffc6d883f772fa5/src/atomate2/lobster/schemas.py#L10

https://github.com/materialsproject/atomate2/blob/8b1ee044674bb75ee9cd025d9ffc6d883f772fa5/src/atomate2/vasp/flows/mp.py#L18

This leads to import errors when lobster is not installed.

naik-aakash:
Closes #702

## Changes

1.  Enable MPVaspLobsterMaker only if lobsterpy and ijson are installed i.e atomate2 is installed using atomate2[lobster] dependencies
2. Add warning message for the same incase MPVaspLobsterMaker  is disabled.

Below is the snapshot of warning printed when MPVaspLobsterMaker is disabled.

image


naik-aakash:
Hi @jmmshn, I hope this addresses the problem now. 

@JaGeo , just tagging you here incase you have any more suggestions or comments on this change.

utf:
Hi @naik-aakash, thanks for this PR.

I think we could handle this differently. Ideally, the warning would be raised if you are actually trying to use the `MPVaspLobsterMaker`. Instead, this warning gets raised even if you just import the MP module.

The typical way of doing this is using the `monty` requires decorator. E.g., see here: https://github.com/materialsproject/atomate2/blob/19849ff022fe199b8104b0f03f11070ba3be8692/src/atomate2/lobster/schemas.py#L523-L525

My suggestion would be:

1. Replace the ijson import in with:
```python
try:
    import ijson
except ImportError:
    ijson = None
```

2. Add the requires decorator onto the function that needs ijson.

naik-aakash:
Thanks @utf, for the suggestion, I will make the required changes.

naik-aakash:
Hi @utf , I have now made the changes and this PR could be merged if you think this is fine.

utf:
Looks great, thank you!

PR #710, opened 2024-02-09
Fix JobStoreDocument attribute access 

mjwen:
## Summary

Fix the error:
```
  File "/home/atomate2/src/atomate2/cli/dev.py", line 58, in vasp_test_data
    if not isinstance(output["output"], TaskDoc):
                      ~~~~~~^^^^^^^^^^
TypeError: 'JobStoreDocument' object is not subscriptable
```

### DIFF
--- a/src/atomate2/cli/dev.py
+++ b/src/atomate2/cli/dev.py
@@ -47,7 +47,9 @@ def vasp_test_data(test_dir: Path) -> None:
 
     outputs = loadfn("outputs.json")
 
-    task_labels = [o["output"].task_label for o in outputs if isinstance(o, TaskDoc)]
+    task_labels = [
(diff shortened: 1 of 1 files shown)


mjwen:
I was trying to add a test, and what I was doing is:
- run a vasp calculation
- add the job directory job_2024-02-10-... to the test fold 
- run `atm dev vasp-test-data dummy` to test it
 
It won't work because the directory to copy the VASP files is from outputs.json, and it is an absolute path. Because the vasp files are relocated, it won't work. 

To make it work, we need to update `dev.py` to accept a new job dir e.g. from cli, not from outputs.json. This can be done by adding an optional switch (e.g. `--job_dirs`) to `atm dev vasp-test-data`.  

@utf If you think this is a OK, I can add it. Otherwise, this PR is ready for review.

utf:
Thanks!

PR #715, opened 2024-02-12
WIP: Vasp ml

davidwaroquiers:
## Summary

Adding Vasp ML functionality in MultiMDMaker.
Also added possibility to rename additioal vasp files within copy_vasp_outputs (needed to allow ML_ABN=>ML_AB and ML_FFN=>ML_FF)

* New MLMDMaker
* MLFF feature within MultiMDMaker (as a classmethod)
* MLMDSetGenerator
* Possibilty to rename additional vasp files after they have been copied (by providing a dict instead of a list).

None

## TODO (if any)

If this is a work-in-progress, write something about what else needs to be done.

* Tests

## Checklist

title.

  type check your code.


davidwaroquiers:
@ml-evs As we discussed about using that, could you try this implementation ?
@gpetretto Comments ?

Still more features to be added in the MultiMDMaker (train+run, refit, ...)

davidwaroquiers:
This PR needs (at least) this to be merged in pymatgen: https://github.com/materialsproject/pymatgen/pull/3625
the ML_MODE tag in vasp is not recognized when its value is capitalized. E.g. ML_MODE = Train does not work. Pymatgen currently capitalizes all strings (e.g. ALGO = Fast, even if you set incar["ALGO"] = "fast").

davidwaroquiers:
This PR also needs the following PR to be merged in emmet: https://github.com/materialsproject/emmet/pull/939

davidwaroquiers:
Pinging @utf to get your opinion on this draft, especially in terms of the input sets for which I added some option to disable band structure parsing (this should obviously be moved to pymatgen anyway to the new input sets there).

utf:
Hi @davidwaroquiers, appologies for the delay looking at this. This looks good to me, except the part about the band gap handling. These seems quite heavy handed. I wonder if the input sets can just check for the band gap and if there is an error then just set a default value. I.e., this is all handled in the input set generator entirely, without needing to make changes in other files.

davidwaroquiers:
> Hi @davidwaroquiers, appologies for the delay looking at this. This looks good to me, except the part about the band gap handling. These seems quite heavy handed. I wonder if the input sets can just check for the band gap and if there is an error then just set a default value. I.e., this is all handled in the input set generator entirely, without needing to make changes in other files.
> 
> I think this workflow will also need to be updated to use the latest input sets in pymatgen once #854 is merged.

Thanks @utf ! No worries about the delay! I've seen #854 has been merged so I'll work this out when I'm back from holidays. Indeed the band gap thing is a bit clunky... I will see how to do that in a cleaner way.

Issue #515, opened 2023-09-09
Output document for large MD output files


gpetretto:
I would like to start a discussion about how to handle the outputs in cases where the generated output files have very large size. To be more explicit, I am mainly referring to long MD simulations with VASP, either generated by pure DFT or by ML-FF generated with active learning. 
In these cases the parsing of the outputs usually performed in the `TaskDoc.from_directory()` could be rather expensive in terms of time and allocated memory. Here a few points that may be worth considering:
* Is it worth in the case of an MD parsing all the data that are parsed for a simple SCF or relaxation (e.g. all the content of the OUTCAR and vasprun.xml, including all ionic and electronic steps)? Both in terms of time required to parse and space occupied in the DB. 
* The main output of a the MD is likely the Trajectory, where only the coordinates are stored. Currently first all the structures are instantiated and then used to build the `Trajectory` using `from_structures`. Would it be worth finding a more optimized way to extract the Trajectory alone, without going through the allocation of all the Structure objects? 
* VASP optionally produces an hdf5 output file. Would it be better and possible to use that to parse the outputs, if present, instead of the other text files? It should be way faster and easier to deal with.
* The data in vasprun.xml produced by ML-FF in VASP differ from the standard one. For example no equivalent of the `IonicStep` is generated, and would likely not make much sense to have it in the final document. How to handle the output data in that case?

Given that the current parsing suffers from the above limitations, what could be an acceptable improvement? Define an ad hoc output document for MD in general? An output document specific for ML-FF calculations with vasp? Or just leave things as they are for standard MD and try improving the current `TaskDoc` to properly handle ML-FF simulations, adding some fields and leave other empty?

utf:
Copying across comments from @naik-aakash in #703.

> Relax jobs using forcefields fails as the forcefield taskdoc stores all the ionic steps leading to cross the size limit
> 
> **Steps to reproduce**
> 
> 1.Run relax jobs using mace forcefield with Fire optimizer for a cell containing more than 1000 atoms with a force requirement of 0.0001
> 
> **Possible solution**
> 
> - Modify the default interval arg of relax maker from 1 to 100 / larger number (this solution should potentially work as we tested using LBFGS optimizer of ase instead of default Fire optimzer, and same job then runs without any issue as LBFGS stores 100 steps at max)
> - arg to allow users to skip storing ionic step data in taskdoc
> - Any thoughts on this or other ideas to tackle this issue are welcome.

JaGeo:
[...]

PR #722, opened 2024-02-16
[WIP] Forcefield molecular dynamics

esoteric-ephemera:
### Completed:
- Add a forcefield maker to perform molecular dynamics with forcefields in ASE, `atomate2.forcefields.md.ForceFieldMDMaker`
  - The defaults for this class (starting and ending temperature, pressure, Langevin damping, ensemble, thermostat) are matched as closely as possible to the VASP MD maker
  - Units for params are the same between both implementations, see the docstr
  - There's some duplication of the code in `atomate2.forcefields.utils.Relaxer`, but I don't see a better way to merge those approaches for trajectory observance without a messier implementation
- Specific subclasses of `ForceFieldMDMaker` for CHGNet, M3GNet, and MACE

### To Do:
- Add tests
- Hoping to get feedback about the current structure of `ForceFieldMDMaker` from those who are more familiar with MD in ASE (or MD experts generally). This is to support multiple efforts within MP where having a canned MD maker for ML forcefields would be useful


JaGeo:
@esoteric-ephemera great that you started this. One thing that will likely be really brittle are the trajectories. We need to find a solution for this in the taskdocs before merging this PR. This is already a problem for normal optimizations (okay, our structures are large but still). 
We had some discussions on this already in https://github.com/materialsproject/atomate2/issues/515

utf:
Firstly, thanks very much for this @esoteric-ephemera. It looks fantastic.

As I just commented in #515, we have the ability to store trajectories in the datastore for VASP workflows (see https://github.com/materialsproject/emmet/pull/886 for implementation details). I think we should add a similar option for the force field workflow and probably use the datastore by default since ML MD runs are likely to be much longer than AIMD.

esoteric-ephemera:
Thanks @JaGeo and @utf! I tried to mimic the `store_trajectory` functionality: by default, only energies, forces, stresses, and magmoms are stored in the TaskDoc ionic steps for FF MD. I can try to add that as a tag to the `ForcefieldTaskDocument` for consistency.

On this point: for FF MD, one only needs the structure to reconstruct those other four quantities, so maybe storing just the structure for FF MD makes more sense (probably less data-efficient)

chiang-yuan:
@esoteric-ephemera temperature and pressure schedule tests are working now. Not sure if we need to test all the calculators. It seems redundant to me if we only want to make sure schedule feature is working

chiang-yuan:
This branch should close #738

esoteric-ephemera:
Hey @utf and @JaGeo, these workflows now store the trajectory in GridFS by default. This is defined in `atomate2.forcefields.jobs.forcefield_job`. Should @chiang-yuan  and I remove these from the ionic steps by default too?

There's a pretty big inconsistency in how the regular relax and static jobs are run (CHGNet and M3GNet have their own relaxers). Is it OK to merge these into the same approach as in the MD maker? Basically having a common `Relaxer` that works for all forcefields and just passing in an ASE calculator

JaGeo:
@esoteric-ephemera 

> Hey @utf and @JaGeo, these workflows now store the trajectory in GridFS by default. This is defined in `atomate2.forcefields.jobs.forcefield_job`. Should @chiang-yuan and I remove these from the ionic steps by default too?
> 
> There's a pretty big inconsistency in how the regular relax and static jobs are run (CHGNet and M3GNet have their own relaxers). Is it OK to merge these into the same approach as in the MD maker? Basically having a common `Relaxer` that works for all forcefields and just passing in an ASE calculator

both suggestions sound like a good idea from my perspective (@utf decides, of course). Could you add an additional check that the relax has really finished and not just hit the max number of steps as well?

esoteric-ephemera:
> Could you add an additional check that the relax has really finished and not just hit the max number of steps as well?

Like a custodian-style handler for any job where ionic positions can change? I can add that check to the task doc

JaGeo:
Yeah. A check in the taskdoc (for the forcefields) would be great already. Handler would be even better, of course, but maybe not necessary.

utf:
> Is it OK to merge these into the same approach as in the MD maker? Basically having a common Relaxer that works for all forcefields and just passing in an ASE calculator

This sounds perfect. Yes, please.

esoteric-ephemera:
OK! Two bigger updates:
- The relax and static forcefield jobs have been refactored to use the same calculator class as the MD jobs. Now one can either specify the `force_field_name` to access a predefined ASE calculator (`atomate2.forcefields.utils.ase_calculator` based on `atomate2.forcefields.MLFF`), or use a dict to load a Calculator using `MontyDecoder`
- There's now a check in `atomate2.forcefields.utils.Relaxer` that sees if a calculation is force converged. The result of this check is passed to the task doc

src/atomate2/forcefields/jobs.py
```diff
@@ -7,26 +7,69 @@
 from typing import TYPE_CHECKING
 
 from jobflow import Maker, job
+from pymatgen.core.trajectory import Trajectory
 
 from atomate2.forcefields import MLFF
 from atomate2.forcefields.schemas import ForceFieldTaskDocument
-from atomate2.forcefields.utils import Relaxer, revert_default_dtype
```
JaGeo, line 13: This has just been added by @janosh . I guess this should stay?

JaGeo:
Awesome. Just raised one point I saw by chance

src/atomate2/forcefields/jobs.py
esoteric-ephemera: Oh my bad, I intended to leave this in when I resolved the merge conflicts. Undoing this now

esoteric-ephemera: Sorry about that - added back to the relax/static jobs and also added to the MD jobs

src/atomate2/common/jobs/phonons.py
```diff
@@ -153,11 +153,13 @@ def generate_phonon_displacements(
         "of the symmetry of the structure and thus will be removed now.",
         stacklevel=1,
     )
-    cell = get_phonopy_structure(
+
+    # Note that `remove_site_property` returns None
```
janosh, line 157: this is no longer true. i changed a lot of structure methods to return self 3 weeks ago in https://github.com/materialsproject/pymatgen/pull/3623

src/atomate2/forcefields/jobs.py
```diff
@@ -42,19 +86,25 @@ class ForceFieldRelaxMaker(Maker):
...
     """
 
     name: str = "Force field relax"
-    force_field_name: str = "Force field"
+    force_field_name: str = f"{MLFF.Forcefield}"
     relax_cell: bool = True
     steps: int = 500
     relax_kwargs: dict = field(default_factory=dict)
     optimizer_kwargs: dict = field(default_factory=dict)
+    calculator_args: Sequence[Any] = field(default_factory=list)
```
janosh, line 103: given every arg can be passed as a kwarg, i would get of `calculator_args`. args are more likely to cause footguns if users get the order of args wrong or when a breaking change in upstream code changes the order. so there are significant safety benefits from forcing users to pass everything as kwargs.

esoteric-ephemera: Removed the args from the forcefield jobs/md jobs and calculator. I added it yesterday for Nequip, but I agree with you that it can cause unexpected problems

esoteric-ephemera:
@chiang-yuan there are large commented-out blocks of text in `forcefields.md`, and there's a note in the forcefield `TrajectoryObserver` about kinetic stress. Are we good to remove those comments? I would think that ase only uses the calculator to evaluate the energy, forces, and stresses due to the potential energy, since the kinetic contributions are independent of the calculator, but I might be wrong there

chiang-yuan:
@esoteric-ephemera yes the comment blocks are safe to remove. It was mainly for letting you know why and how I implement them. 

`includge_ideal_gas=True` can't be set to False for MD simulations. This is a known issue on ase gitlab repo (https://gitlab.com/ase/ase/-/merge_requests/1500).  If it has conflict with relaxation workflow, we might need to create a separate Document class to save the MD information

esoteric-ephemera:
Thanks @chiang-yuan! For the relaxations, it shouldn't matter whether it's set to T / F since the ions have no momenta in a relaxation. We could be safe and set it to true only when MD outputs are stored, but that might not be necessary

esoteric-ephemera:
I think the PR is ready at this point for review. One point worth discussing is whether it makes more sense to have quantities like the temperature and pressure enter via `MDMaker`'s `__init__`, or via its `.make`. The latter method could be more for a dynamic temperature flow like those in MPMorph.

src/atomate2/forcefields/utils.py
```diff
@@ -212,7 +370,78 @@ def relax(
...
+        calculator_name = MLFF(calculator_meta.split("MLFF.")[-1])
+
+        if calculator_name == MLFF.CHGNet:
+            from chgnet.model.dynamics import CHGNetCalculator
+
+            calculator = CHGNetCalculator(**kwargs)
+
+        elif calculator_name == MLFF.M3GNet:
+            import matgl
+            from matgl.ext.ase import PESCalculator
```
Andrew-S-Rosen, line 420: @esoteric-ephemera: One thing I'm confused about --- should we be setting `stress_weight=1.0 / 160.21766208` as [here](https://github.com/materialsvirtuallab/matgl/blob/ee9e98700dd636ee6c6ce57e4c7c9128b29af78c/src/matgl/ext/ase.py#L222)? The `Relaxer` class has this in `matgl`, but the `PESCalculator` does not. In general, I'm not really sure what to make of this `stress_weight` parameter because [CHGNet also has this](https://github.com/CederGroupHub/chgnet/blob/bae0331896c64e5b04481187d63e3e47d78a2e9f/chgnet/model/dynamics.py#L159).

Maybe @janosh knows.

janosh: [it's the conversion factor](https://github.com/CederGroupHub/chgnet/blob/bae0331896c64e5b04481187d63e3e47d78a2e9f/chgnet/model/dynamics.py#L174) from `GPa` to `eV/Angstrom^3`. IIRC, it's there to correct for a unit mismatch between MP and ASE. @BowenD-UCB probably knows better. can we expand the doc string to explain why we need to convert?

Andrew-S-Rosen: Thanks! @BowenD-UCB: If it's indeed the case that it's a way to convert MP units of GPa back to ASE units of eV/A^3, it'd probably be better to do this in a more intuitive way with a keyword argument of `pressure_in_GPa: bool = False` or something rather than a user-supplied conversion factor. The phrase "stress_weight" is vague (it's not really a weight; it's a conversion factor on the result, right?), and there is really only one right value for this.

janosh: i think @kenko911 is the right person to ping about [`PESCalculator`](https://github.com/materialsvirtuallab/matgl/blob/ee9e98700dd636ee6c6ce57e4c7c9128b29af78c/src/matgl/ext/ase.py#L122)

esoteric-ephemera: I feel that an ASE calculator should be written to correctly interface with ASE and provide physical quantities in the units ASE uses. When I refactored the relaxation workflows, I didn't have to change any of the output quantities we test against, so things are at least consistent with their previous implementation

That being said, we need to figure out what the units being used are for the various calculators and ensure the correct kwargs are being passed through

kenko911: In MatGL, the PESCalculator is primarily utilized as a single-point calculator, with stress units set to GPa. However, for Relaxer and MolecularDynamics functionalities within MatGL, we convert the stress unit from GPa to eV/Å³, aligning with the internal unit of ASE. For other applications in ASE, users are required to specify the appropriate stress unit for PESCalculator. We can also add an option to allow users to select their preferred stress units within PESCalculator.

esoteric-ephemera: Thanks @kenko911! I've updated the default `calculator_kwargs` for the M3GNet and CHGNet relax/static makers to use `stress_weight = ase.units.GPa`

src/atomate2/forcefields/jobs.py
```diff
@@ -131,21 +188,19 @@ def make(
                 "Behavior may vary..."
             )
 
-        result = self._evaluate_static(structure)
+        static_calc = Relaxer(self._calculator(), relax_cell=False)
+        result = static_calc.relax(structure, steps=1)
```
utf, line 192: I think we should actually use the values of `self.relax_cell` and `self.steps` here. If the user really wants to turn a static into a relaxation, they should be able to.
```suggestion
        static_calc = Relaxer(self._calculator(), relax_cell=self.relax_cell)
        result = static_calc.relax(structure, steps=self.steps)
```

src/atomate2/forcefields/schemas.py
```diff
@@ -104,6 +127,21 @@ class ForceFieldTaskDocument(StructureMetadata):
...
+    is_force_converged: Optional[bool] = Field(
+        None,
+        description=(
+            "Whether the calculation is converged with respect "
+            "to interatomic forces."
+        ),
+    )
+
     @classmethod
     def from_ase_compatible_result(
```
utf, line 146: Can you confirm that none of these changes impacts the compatibility between DFT and FF schemas in terms of the trajectory? E.g., hopefully we can use both DFT and FF trajectories interchangeably.

utf (commented):
Thanks very much @esoteric-ephemera, these changes are fantastic. I had a couple of minor comments.

Also, I haven't that much experience using the FF jobs/flows in atomate2, so I would appreciate if @JaGeo, @Andrew-S-Rosen, @janosh or anyone else with more experience could take a look and confirm that this can be merged.

JaGeo:
@utf not possible at the moment. Sorry.

src/atomate2/forcefields/schemas.py
mkhorton: Seconding, as a community member just adding a comment that any efforts to keep schemas stable and un-changing (backwards compatible) are hugely appreciated 🙏

src/atomate2/forcefields/jobs.py
esoteric-ephemera: Sounds good to me, I went back and forth on this. Also added a bit to the docstr to explain that defaults are set but they are not fixed

src/atomate2/forcefields/schemas.py
esoteric-ephemera: I'll go through and see if there are any incompatibilities. @QuantumChemist  seems to have found a missing `dir_name` which they patched in [#791](https://github.com/materialsproject/atomate2/pull/791), so there may be other missing attrs

esoteric-ephemera:
Just a note, there's a dgl incompatibility with torch==2.2.2, so I've pinned the dependencies to be torch<=2.2.1

src/atomate2/forcefields/utils.py
```diff
@@ -82,18 +130,42 @@ def __init__(self, atoms: Atoms) -> None:
...
+        self.magmoms: list[np.ndarray] = []
+
         self.atom_positions: list[np.ndarray] = []
         self.cells: list[np.ndarray] = []
 
+        self._store_md_outputs = store_md_outputs
+        if self._store_md_outputs:
+            self.velocities: list[np.ndarray] = []
+            self.temperatures: list[float] = []
+
```
janosh, line 144: i think i agree with this design but maybe worth discussing whether `self.velocities` and `self.temperatures` should always be initialized and just left empty if `store_md_outputs=False`?

janosh (changes requested):
@utf @esoteric-ephemera left a few minor comments re API design

src/atomate2/forcefields/md.py
```diff
@@ -0,0 +1,377 @@
...
+        {
+            "store_trajectory": "partial",
+            "ionic_step_data": ("energy",),
+        }
+        is consistent with atomate2.vasp.md.MDMaker
+    """
+
+    name: str = "Forcefield MD"
+    force_field_name: str = f"{MLFF.Forcefield}"
+    timestep: float | None = 2.0
```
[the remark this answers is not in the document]
esoteric-ephemera: Added both requested changes. If a structure contains an isotope of H, `time_step = 0.5` fs, and 2 fs otherwise. Also changed the behavior so that it isn't assumed that a user specifies `time_step` in fs

src/atomate2/forcefields/utils.py
esoteric-ephemera: Changed as well, these are always initialized but no data is added to them, and no info is saved to the trajectory if an MD run is not performed

tests/forcefields/test_md.py
```diff
@@ -0,0 +1,255 @@
...
+        step.energy == pytest.approx(-10.6, abs=0.1)
+        for step in output["from_str"].output.ionic_steps
+    )
+
+    # ensure that output is consistent if molecular dynamics object is specified
+    # as str or as MolecularDynamics object
+    assert all(
+        output["from_str"].output.__getattribute__(attr)
+        == output["from_dyn"].output.__getattribute__(attr)
+        for attr in ("energy", "forces", "stress", "structure")
```
janosh, line 197: this looks like a float comparison (except for `structure`)? if so, should this use `pytest.approx` (and `getattr`)?

janosh (changes requested):
code-wise this looks great! thanks again @esoteric-ephemera 👍 
only remaining concern is test time. tests on the `mlff_md` branch take 17 min vs 12 on the main branch. can we reduce the number of steps or system size in tests?

tests/forcefields/test_md.py
esoteric-ephemera: Changed this to use approx when appropriate. The intent was that two runs should be identical when run either explicitly specifying the dynamics object, or implicitly via str, but that is maybe too optimistic

I tried using `getattr`, but the forcefield task doc has no defined `getattr` method, hence the dunder

I can write these out explictly if that's preferable to the dunder. Or try to add a `getattr`

esoteric-ephemera:
@janosh:
>  only remaining concern is test time. tests on the mlff_md branch take 17 min vs 12 on the main branch. can we reduce the number of steps or system size in tests?

Tried to reduce the time for this by reducing the number of steps in the temperature / pressure MD tests, as well as only testing for CHGNet (MACE is 3x slow for these tests)

@chiang-yuan: the temperature and pressure scheduling didn't seem right to me. When I ran the tests, the temperature schedule gave me `[300, 3000, 3000,...,3000]`, when it should've been a smooth ramp from 300 to 3000 K, no?

I've modified the logic for the scheduling in a new function `ForceFieldMDMaker._interpolate_quantity`, can you take a look and let me know if this is what you had in mind?

esoteric-ephemera:
> Thanks @esoteric-ephemera !! Sorry I confuse the behavior of numpy interp and scipy interp1d. Thanks for the nice catch! I slightly changed a bit the syntax for schedule interpolation and make it seems (at least to me) explainable

Perfect thanks! This looks much cleaner now

utf:
Hi @esoteric-ephemera. Thanks again for all the edits. Is this good to go now?

> I'll go through and see if there are any incompatibilities. QuantumChemist seems to have found a missing dir_name which they patched in https://github.com/materialsproject/atomate2/pull/791, so there may be other missing attrs

Did you have a chance to do this?

esoteric-ephemera:
> Hi @esoteric-ephemera. Thanks again for all the edits. Is this good to go now?
> 
> > I'll go through and see if there are any incompatibilities. QuantumChemist seems to have found a missing dir_name which they patched in #791, so there may be other missing attrs
> 
> Did you have a chance to do this?

Confirmed that the following works without issue:
```

forcefield_taskdoc : ForceFieldTaskDocument = < output of some forcefield job >
vasp_style_taskdoc = TaskDoc(**forcefield_taskdoc.model_dump())
```
There are some optional fields that aren't populated (`calcs_reversed`, `dir_name`, etc.) but those aren't so important given that MD trajectories are being stored in GridFS.

For relax / static jobs, is there interest in having a `calcs_reversed` field on the forcefield task doc?

utf:
Thanks for checking - `calcs_reversed` is specific to the calculator used and won't be part of the universal task document (when we get to addressing #741). As long as the other fields match we should be good to go.

janosh (approved):
i'll go ahead and merge since everyone seems happy with this.

thanks so much @esoteric-ephemera, very excited to have MLFF MD makers in `atomate2`! 🎉

utf:
Fantastic, thanks everyone who contributed to this PR and especially @esoteric-ephemera and @chiang-yuan!

JaGeo:
Great!

Short question to the ML potential crowd: wouldn't a LAMMPS interface be even more beneficial? Especially for MD? I am aware of the atomate2-LAMMPS addon but it currently does not look ready to use (and is even archived: https://github.com/Matgenix/atomate2-lammps).

janosh:
never used LAMMPS myself, I think OpenMM might be the better fit for GPU-accelerated ML potentials based on what I heard from @orionarcher

orionarcher:
I've found OpenMM to be MUCH easier to use than LAMMPS and seen it run much faster on GPUs. Though I don't know how many popular ML potentials have been implemented, OpenMM has a [plugin for integration with Pytorch](https://github.com/openmm/openmm-torch). 

Caveat: I have only used OpenMM for classical MD and not ML potentials

JaGeo:
@orionarcher thanks! From the ML community side I always hear the LAMMPS wish. Just wanted to see if there are plans or opinions.

PR #725, opened 2024-02-17
Fix docstring on `MatPesMetaGGAStaticSetGenerator`

Andrew-S-Rosen:
It said GGA when it should have said meta-GGA. Trivial. CC @janosh.


janosh (approved):
thanks!

PR #728, opened 2024-02-18
Print invalid value and list valid ones in `PhononMaker` error messages

janosh:
dbd7e991 `PhononMaker` print invalid values and list valid ones in for `kpath_scheme`, `use_symmetrized_structure`
37750762 fix `cclib_calculate()` wrong err msg claiming to `return None` when actually raising
f20c68e1 prefer tuple for static sequence literal
bd9b02a7 assert unused ret val distances in `test_sort_pos_dist()`
0f326f52 simplify dict construction in lobster `read_saved_json()`, `LobsterTaskDocument.from_directory`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ default_language_version:
   python: python3
 repos:
 - repo: https://github.com/charliermarsh/ruff-pre-commit
-  rev: v0.2.1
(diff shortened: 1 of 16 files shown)


janosh:
CI errors are unrelated caused by MP API. how about we skip those tests on GH or mock the API call @utf?

PR #742, opened 2024-02-26
Defect: Allow bulk SC calculation to be skipped.

jmmshn:
## Defect Calcs: Allow bulk SC calculation to be skipped. 

Upstream changes now allow arbitrary UC Locpots to be used for the Freysoldt correction.
https://github.com/materialsproject/pymatgen-analysis-defects/pull/171

So you can just start with the UC `LOCPOT` and the SC defect `LOCPOT` and obtain the correction value.
For the large-scale runs that we are doing this is a lot easier for bookkeeping purposes. 

Additionally, using a relaxed bulk supercell is not directly comparable with the output of defect SC relaxation with selective dynamics.
So skipping the bulk relax calculation should also be better in that situation.


yuan-gist:
Not sure the reason to get bulk LOCPOT from unit cell. I mean a bulk static calculation costs nothing, though calling it a bulk relax calculation is indeed strange. Also noticed the get_defect_entry no longer contains "bulk_entry.structure". For me, I prefer every defect's get_defect_entry contains bulk information (LOCPOT, structure), so the calculation for every single defect can be parsed in a consistent way

jmmshn:
Agreed it's a little strange: so the rationale is this:
When we do a bunch of defects I would like to run them selective dynamics to avoid a point defects causing long-range relaxations of the bulk and finding a lower energy atomic configuration that corresponds to another structural phase (we have actually seen this in some defects).   

Originally we allowed for a ISIF=3 relaxation for the bulk because some of the single-point supercell Kpoint settings are on the verge of being sufficiently converged so having an ISIF=3 reference bulk allows for a bit of cancellation of errors.

Now if we have a very converged unit cell, we can just use the Locpot from that and always assume it's good enough and just live with the possible energy convergence error in the defect supercell. 

I also agree that the get_defect_entry operation should include the bulk supercell, this update just really targeted at large-scale database building and not automated defect analysis.   Since the idea is just to have as little provenance information as possible required to construct a `DefectEntry`.   Using UC Locpot you not longer have to keep track of which defect SC and bulk SC correspond to each other and just have to get a matched UC locpot from somewhere in your database at the same level of theory and you can arrive at the correction values.

yuan-gist:
Thanks a lot for clarifying. I am building a defect database, though it will not be large scale. I really like the "get_defect_entry", which combines different charge states of a defect. The previous one also contains more bulk info. These greatly facility the query. The only thing I feel missing is the "electrostatic potential" for Kumagai corrections. I am considering adding them to the "get_defect_entry". It is just because sometimes I do not want to do the original FNV corrections because of anisotropy in dielectric constant.

yuan-gist:
I would also like to suggest the MP_id and formula_pretty can be added somewhere to facilitate database organization and query.

yuan-gist:
Hi @jmmshn Not sure there will be adding `atomic-site electrostatic potential ` to `get_defect_entry`? I did it locally.

jmmshn:
Can you start a WIP PR for that?  I think it will be really useful.

yuan-gist:
> @yuan-gist Can you start a WIP PR for that? I think it will be really useful.

Sure. Will be done by next week

utf:
Is this still WIP?

jmmshn:
@utf  missed the message above.  I have been running this and some downstream builders at scale and wanted to get a good handle on the results to see if more changes are needed.
This should be good to go now.

utf:
Thanks @jmmshn. Appologies for the delay.

jmmshn:
> Thanks @jmmshn. Appologies for the delay.

No worries!  I'm actually able to scale this workflow to do some real science for the first time.
Doing some data cleanup before writing the manuscript, but I will send that to you once completed!

PR #752, opened 2024-02-27
Add workflow to compute Gruneisen parameters 

naik-aakash:
# Summary

This PR aims to add a Gruneisen workflow using the phonopy API

# Todo

- [x] Add job to compute Gruneisen parameter to the flow
- [x] Add initial schema for Gruneisen runs
- [x] adapt base phonon workflow to be able to change saved `phonopy.yaml` file name (add kwarg)
- [x] Add BaseGruneisenMaker
- [x] Add GruneisenMaker for forcefields 
- [x] Add more kwargs for flexibility (kpath scheme, plot, and yaml file names, phonopy gruneisen python API for mesh and bandstructure)
- [x] Add tests (preliminary for forcefield workflow and job)
- [x] Add high-accuracy reference files to test specifically the `compute_gruneisen_param` job to test derived properties field entries in the output doc
- [x] Fix the custom band structure that includes Grüneisen parameter values for color mapping. We need a single color map for the whole range of values; presently, it is for each band. The color bar on the right shows only the last band values (not correct). We need to add a label for the color bar. (Solution: Normalize the color map using the min and max values of the Gruneisen parameter with log scale)
- [x] Add a VASP-compatible flow as well (currently, ISIF=2 is implemented, maybe ISIF=4)
- [x] add names to steps in the workflow
- [x] fix https://github.com/materialsproject/jobflow/issues/639 (otherwise, `mock_vasp` will not work correctly)
- [x] add vasp tests
- [x] check symmetry after optimizations
- [x] Add workflow to documentation


naik-aakash:
Hi @JaGeo , just tagging you here for updates.

JaGeo:
Thanks!

Can we add that we might need to adapt the current optimization in the force fields, as we don't change the cell shape currently and that we need symmetry and potentially checks that the optimization has converged (especially for force fields?)

tpurcell90:
If you also want to make this more of a common workflow I can add aims support

JaGeo:
@tpurcell90 
Surely. It will take some time though. The current Implementation is the result of a first tiny hackathon that we had to introduce jobflow concepts more widely in the group

tpurcell90:
Ya no rush on that, just ping me when you want me to take a look at it

joabustamante:
Hi @JaGeo, We have updated the job to the flow that also includes computing Gruneisen parameters from the phonon runs. 

Next, we will add a schema for the flow if desired and also tests

JaGeo:
Thank you both, @joabustamante and @naik-aakash !

I left one inital comment. Might need to be checked in the schema itself.

src/atomate2/common/jobs/gruneisen.py
```diff
@@ -0,0 +1,123 @@
...
+    gruneisen_parameter = get_gruneisenparameter(
+        gruneisen_path="gruneisen_mesh.yaml", structure=structure
+    )
+    gp_plot = GruneisenPlotter(gruneisen=gruneisen_parameter)
+    gp_plot.save_plot(
+        filename="gruneisen_mesh.pdf",
+        units=plot_kwargs.get("units"),
+        img_format=plot_kwargs.get("img_format"),
+    )
+    ground.auto_band_structure()
```
JaGeo, line 87: This would mean that we just have one kpath schema available.

naik-aakash: Yeah, it's not yet finalized.  Just got the workflow running first , indent to add kwargs

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,153 @@
...
+if TYPE_CHECKING:
+    from pymatgen.core.structure import Structure
+
+    from atomate2.vasp.jobs.base import BaseVaspMaker
+
+
+@dataclass
+class BaseGruneisenMaker(Maker):
+    """
+    Maker to calculate gruneisen parameters.
```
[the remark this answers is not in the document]
naik-aakash: Thanks, will make the change

src/atomate2/common/jobs/gruneisen.py
JaGeo: Great, maybe, you can reuse code from the phonon workflow!

JaGeo:
@naik-aakash @joabustamante Great work already! :)

src/atomate2/common/jobs/gruneisen.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Compute Gruneisen parameters from phonon runs.
+
+    Requires phonopy yaml files from ground, expanded and contracted structures
+
+    phonopy_yaml_paths_dict: optimized pymatgen structure obj
+    mesh: sampling mesh
+    structure: pymatgen structure object at ground state
+    """
+    ground = phonopy.load(Path(phonopy_yaml_paths_dict["ground"]) / "phonopy.yaml")
```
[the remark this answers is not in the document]
naik-aakash: This info is added to the schema for all the 3 phonon runs

src/atomate2/common/jobs/gruneisen.py
naik-aakash: Have added it now

JaGeo: Congrats @naik-aakash . Really great already!

naik-aakash:
Hi @tpurcell90, We have now added a draft for common flow as `BaseGruneisenMaker`. It would be great if you could have a look at it and provide us feedback on it, mainly whether it would be fine to use it with FHI-aims in its current state. And also if you have suggestions for the output schema. We have added very minimal schema as of now. 

The custom bandstructure plotter in the schema will be moved to Pymatgen in the next few days (it's residing here now just for testing purposes)

We plan to add a VASP-compatible flow soon, but it could take some time.

tpurcell90:
> Hi @tpurcell90, We have now added a draft for common flow as `BaseGruneisenMaker`. It would be great if you could have a look at it and provide us feedback on it, mainly whether it would be fine to use it with FHI-aims in its current state. And also if you have suggestions for the output schema. We have added very minimal schema as of now.
> 
> The custom bandstructure plotter in the schema will be moved to Pymatgen in the next few days (it's residing here now just for testing purposes)
> 
> We plan to add a VASP-compatible flow soon, but it could take some time.

Ya I can do this, I will try to get this done next week (just flew into Berlin for a visit and have meetings all day today)

JaGeo:
I will now try to implement the VASP support.

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,186 @@
...
+        scheme to generate kpoints. Please be aware that
+        you can only use seekpath with any kind of cell
+        Otherwise, please use the standard primitive structure
+        Available schemes are:
+        "seekpath", "hinuma", "setyawan_curtarolo", "latimer_munro".
+        "seekpath" and "hinuma" are the same definition but
+        seekpath can be used with any kind of unit cell as
+        it relies on phonopy to handle the relationship
+        to the primitive cell and not pymatgen
+    mesh: tuple
```
JaGeo, line 62: this should not be a mesh .it should be a density/or a reciprocal distance similar to the phonon workflow.

naik-aakash: Do you mean the arg name?

JaGeo: No, i rwally mean the input. Will fix this. Don't worry. Just leaving comments.

tpurcell90:
> > Hi @tpurcell90, We have now added a draft for common flow as `BaseGruneisenMaker`. It would be great if you could have a look at it and provide us feedback on it, mainly whether it would be fine to use it with FHI-aims in its current state. And also if you have suggestions for the output schema. We have added very minimal schema as of now.
> > The custom bandstructure plotter in the schema will be moved to Pymatgen in the next few days (it's residing here now just for testing purposes)
> > We plan to add a VASP-compatible flow soon, but it could take some time.
> 
> Ya I can do this, I will try to get this done next week (just flew into Berlin for a visit and have meetings all day today)

There is nothing obvious that would prevent FHI-aims from also having this, but I will try to implement it now

JaGeo:
@tpurcell90 started working on the vasp integration. I have already done some small modifications. Maybe, let me push that first. We need different names for the jobs before we can implement any tests

tpurcell90:
> @tpurcell90 I started working on the vasp integration. I have already done some small modifications. Maybe, let me push that first. We need different names for the jobs before we can implement any tests

Okay I will wait for you to be done first and then port it for FHI-aims

JaGeo:
@utf should be ready for review. I have now also added tests for the VASP parts.

src/atomate2/common/flows/gruneisen.py
JaGeo: ah, just got reminded about this part here....

JaGeo: added new density option and switched to gamma centered meshes. Will need to check if this is a good idea for the grüneisen parameter

JaGeo:
@jmmshn Just to notify you: the job names of your workflows have slightly changed due to an update on jobflow. I hope this does not affect any of your analysis scripts.

JaGeo:
Tests are finally passing.

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,230 @@
...
+    name: str = "Gruneisen"
+    bulk_relax_maker: ForceFieldRelaxMaker | BaseVaspMaker | BaseAimsMaker | None = None
+    code: str = None
+    const_vol_relax_maker: ForceFieldRelaxMaker | BaseVaspMaker | BaseAimsMaker = None
+    kpath_scheme: str = "seekpath"
+    phonon_displacement_maker: ForceFieldStaticMaker | BaseVaspMaker | None = None
+    phonon_static_maker: ForceFieldStaticMaker | BaseVaspMaker | None = None
+    phonon_maker_kwargs: dict = field(default_factory=dict)
+    perc_vol: float = 0.01
+    mesh: tuple | float = 7_000
```
utf, line 85: Is this a 7000x7000x7000 q-point mesh? Seems a bit high?

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,230 @@
...
+    compute_gruneisen_param_kwargs: dict = field(default_factory=dict)
+    symprec: float = 1e-4
```
utf, line 87: Should think about standardising these across all the related phonon workflows? E.g., defining a constant somewhere?

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        Optimizes structure and runs phonon computations.
+
+        Phonon computations are run for ground state, expanded and shrunk
+        volume structures. Then, Grueneisen parameters are computed from
+        this three phonon runs.
+
+        Parameters
+        ----------
+        structure: Structure
```
utf, line 99: Missing doc string information, e.g. description of structure and also prev_dir.

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,230 @@
...
+        "seekpath", "hinuma", "setyawan_curtarolo", "latimer_munro".
+        "seekpath" and "hinuma" are the same definition but
+        seekpath can be used with any kind of unit cell as
+        it relies on phonopy to handle the relationship
+        to the primitive cell and not pymatgen
+    mesh: tuple|float
+        Mesh numbers along a, b, c axes used for Grueneisen parameter computation.
+        Or float to indicate a kpoint density.
+    phonon_displacement_maker: .ForceFieldStaticMaker | .BaseVaspMaker | None
+    phonon_maker_kwargs: dict
```
utf, line 69: Missing doc string descriptions.

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,230 @@
...
+    Calculate the harmonic phonons of a material for and compute Grueneisen parameters.
+    Initially, a tight structural relaxation is performed to obtain a structure without
+    forces on the atoms. The optimized structure (ground state) is further expanded and
+    shrunk by 1 % of its volume. Subsequently, supercells with one displaced atom are
+    generated for all the three structures (ground state, expanded and shrunk volume)
+    and accurate forces are computed for these structures. With the help of phonopy,
+    these forces are then converted into a dynamical matrix. This dynamical matrix of
+    three structures is then used as input phonopy Grueneisen api to compute Grueneisen
+    parameters are computed.
+
```
utf, line 42: Too many blank lines.

src/atomate2/common/flows/gruneisen.py
```diff
@@ -0,0 +1,230 @@
...
+        Parameters
+        ----------
+        phonon_displacement_maker: ForceFieldStaticMaker|BaseVaspMaker|None
+            Maker for displacement calculations.
+        phonon_static_maker: ForceFieldStaticMaker|BaseVaspMaker|None
+            Maker for additional static calculations.
+        bulk_relax_maker: : ForceFieldRelaxMaker|BaseVaspMaker|None
+            Maker for optimization. Here: None.
+        phonon_maker_kwargs: dict
+            Additional keyword arguments for phonon maker.
```
utf, line 225: missing symprec.

src/atomate2/common/jobs/gruneisen.py
```diff
@@ -0,0 +1,250 @@
...
+
+    from atomate2.forcefields.jobs import ForceFieldStaticMaker
+    from atomate2.vasp.jobs.base import BaseVaspMaker
+
+
+@job
+def shrink_expand_structure(structure: Structure, perc_vol: float) -> Response:
+    """
+    Create structures with expanded and reduced volumes.
+
```
utf, line 40: Docstrings not formatted according to numpy style, also missing `Parameters` title. Noticed this is the case for many functions in this file so won't mention it again.

src/atomate2/common/jobs/gruneisen.py
```diff
@@ -0,0 +1,250 @@
...
+    phonopy_yaml_paths_dict: phonopy yaml files path for ground, expanded and
+        contracted structure phonon runs
+    phonon_imaginary_modes_info: dict with bool indicating if structure
+        has imaginary modes
+    mesh: sampling mesh, or kpoint density
+    structure: pymatgen structure object at ground state
+    compute_gruneisen_param_kwargs: kwargs for phonopy Grueneisen
+        api and pymatgen plotters
+    """
+    ground = phonopy.load(
```
utf, line 140: Does the phonon_yaml get stored in the task document? If so, is it better to use that as it will support distributing calculations over multiple compute resources?

src/atomate2/common/jobs/gruneisen.py
```diff
@@ -0,0 +1,250 @@
...
+        thermal_conductivity_slack = gruneisen_parameter.thermal_conductivity_slack()
+    except ValueError:
+        thermal_conductivity_slack = None
+
+    derived_properties = {
+        "average_gruneisen": average_gruneisen,
+        "thermal_conductivity_slack": thermal_conductivity_slack,
+    }
+
+    return GruneisenParameterDocument.from_structure(
```
utf, line 242: I feel like a lot of this function could be on the task document itself. E.g. `from_phonon_yaml` or something. And then this function is only responsible for loading the files and passing the contents on. That way, if we ever want to create a builder for this workflow it will be much easier. I was also under the impression that this is how the phonon and other documents work.

src/atomate2/forcefields/flows/gruneisen.py
```diff
@@ -0,0 +1,130 @@
...
+        bulk_relax_maker: .ForceFieldRelaxMaker|None
+            Relax Maker for Phonon Run. Typically None.
+        phonon_maker_kwargs: dict
+            Dict to set additional info for phonons.
+
+        Returns
+        -------
+        .PhononMaker
+        """
+        return PhononMaker(
```
utf, line 124: I'm confused about the logic of this part of the code. Why can't the `BaseGruneisenMaker` just have a `phonon_maker` attribute directly which is overriden by each of the sub-implementations? That would also reduce the total number of attributes the superclass has.

utf (commented):
Hi @JaGeo, thanks very much for this! Very glad to have it in atomate2.

I have some general questions which don't necessarily require changes. But the one thing that could potentially be adjusted is how the `PhononMaker` works. I think this can be optimised but maybe I've misunderstood something.

src/atomate2/forcefields/flows/gruneisen.py
JaGeo: Yes, it could, but it might also lead to issues if the 'bulk_relax_maker' is accidentally added. If it's added without changing ISIF to 2 in the INCAR, the entire computation would be lost. I actually made this mistake while modifying the workflow. The current implementation prevents this. I'm also considering creating a dedicated Maker for constant volume optimizations to prevent the user from allowing a volume change...

src/atomate2/common/jobs/gruneisen.py
JaGeo: The whole yaml isn't getting stored. I will need to think about how to do it. While I agree with you about multiple computing resources, this adds data transfer costs. 

At the moment, the whole `prev_dir` concept also doesn't really agree with multiple computing resources, right?

src/atomate2/common/jobs/gruneisen.py
JaGeo: Thanks, will fix. Overlooked those when I started from the pull request ;).

src/atomate2/common/flows/gruneisen.py
JaGeo: Ah, no. This is only the kpoint density. Maybe, I should make this clearer that you can either give a mesh or define a density.

jmmshn:
> @jmmshn Just to notify you: the job names of your workflows have slightly changed due to an update on jobflow. I hope this does not affect any of your analysis scripts.

Can you point me to the jobflow PR for this?

JaGeo:
> > @jmmshn Just to notify you: the job names of your workflows have slightly changed due to an update on jobflow. I hope this does not affect any of your analysis scripts.
> 
> Can you point me to the jobflow PR for this?

Yes, sure. I am also responsible for this one as it was a requirement for testing any more complex workflows. https://github.com/materialsproject/jobflow/pull/644

I am happy to help to fix the naming for both of your workflows but it is likely faster if you point me to the parts where a name might be appended to a workflow I a response object.

src/atomate2/common/jobs/gruneisen.py
JaGeo: I think these files are too large to pass through the jobstore. Especially with remote submission, this will become really slow and nearly impossible to run on a larger scale. In the Lobster workflows, we also don't put the WAVECAR into the jobstore.

We should think about a safe way to handle file copying between different computing resources in the future instead.

src/atomate2/common/jobs/gruneisen.py
JaGeo: Thanks. Yes, I agree. I have adapted this.

src/atomate2/common/jobs/gruneisen.py
JaGeo: Done

src/atomate2/common/jobs/gruneisen.py
utf: That's fine, thanks for checking.  @gpetretto has an open PR on this which I just need to have one final look over.

src/atomate2/forcefields/flows/gruneisen.py
utf: If we made a ConstantVolumeRelaxMaker I think that would be used in a few workflows. So I'm in favour of doing that and then simplying this  part of the code along the lines of my first comment.

JaGeo: @utf  i will work on this constant volume optimizert.

I am not sure I am getting the part on the phononmaker. I would prefer not to write a PhononMaker without bulk relaxation or is this what you are suggesting?

JaGeo: Just to clarify: the phonon maker currently is used without any optimization. The constant volume relaxation is performed before the phonon run with a separate maker. I would find it very confusing to add this to the phonon maker rather than having a separate maker for the optimization. This is why I added this initialization.

JaGeo: I added a constat volume RelaxMaker and TightRelaxMaker. 

(Side note: can we think about how to make writing these Makers, Sets easier in the future? I feel this is still quite complex to write)

utf: I'd be happy to chat about making those easier. Would be useful to know which parts are the most bothersome at the moment.

Thanks for explaining more about the phonon maker. I think I misunderstood before. I think the best fix in this case is just to add some code in the `__post_init__` method that checks if the phonon maker has bulk_relax_maker set. It should just print a warning in that case (someone may have a reason for setting it in the future).

E.g., 

```python
@dataclass
class BaseGruneisenMaker(Maker, ABC):
        
        def __post_init__(self):
            if self.phonon_maker.bulk_relax_maker is None:
                 warnings.warn("...")
```

I think this approach will fix the issue you're worried about but also clean up the code of the generalised maker and make doing specific sub implementations easier.

JaGeo: > I'd be happy to chat about making those easier. Would be useful to know which parts are the most bothersome at the moment.

There are still a lot of layers in the implementation. Especially also writing this for a new "DFT" code seems to be a lot of effort. Can one maybe unify this a bit more or maybe start from a function to write inputs? You remember that I tried to reimplement this for Quantum Espresso (just for showing it to new users): https://github.com/materialdigital/ADIS2023/blob/main/jobflow.ipynb 

> Thanks for explaining more about the phonon maker. I think I misunderstood before. I think the best fix in this case is just to add some code in the `__post_init__` method that checks if the phonon maker has bulk_relax_maker set. It should just print a warning in that case (someone may have a reason for setting it in the future).
> 
> E.g.,
> 
> ```python
> @dataclass
> class BaseGruneisenMaker(Maker, ABC):
>         
>     def __post_init__(self):
>         if self.phonon_maker.bulk_relax_maker is None:
>             warnings.warn("...")
> ```
> 
> I think this approach will fix the issue you're worried about but also clean up the code of the generalised maker and make doing specific sub implementations easier.

I will use this then instead.

JaGeo: @utf Yeah, I tried this. However, there are at least three points that can go wrong. Not sure, the updated implementation now is better.

JaGeo:
@utf I think I implemented all suggestions. Let me know what you think.

src/atomate2/common/flows/gruneisen.py
utf: Bump

src/atomate2/common/utils.py
```diff
@@ -11,6 +11,8 @@
 if TYPE_CHECKING:
     from pathlib import Path
 
+PHONON_SYM_PREC = 1e-4
```
utf, line 14: Do you think this should be defined in the `SETTINGS` object?

utf (commented):
Hi @JaGeo, this looks great to me. Just two very minor questions.

src/atomate2/common/utils.py
JaGeo: As an additional symprec or as a standard symprec? I also think we might have to test this setting on a larger number of structures.

JaGeo:
@utf thank you for the comments. I will correct the broken sentences in the comments and then wait for your opinion on the settings

src/atomate2/common/utils.py
utf: I would imagine this should be an additional symprec.

JaGeo:
@utf should hopefully now be ready to be merged. I had to adapt the code after the new input sets got merged.

naik-aakash:
Hi @JaGeo , It would be better to now replace the method used for generating Gruneisen bandstrucuture plot in the schema to use the one we implemented in pymatgen. Will push the change if it is fine with you in sometime today 😃

naik-aakash:
> Hi @JaGeo , It would be better to now replace the method used for generating Gruneisen bandstrucuture plot in the schema to use the one we implemented in pymatgen, before merging this PR. I will push the change if it is fine with you in sometime today 😃

Nevermind, seems there are other workflows breaking on updating pymatgen version. So I leave it as is for now and change it later.

JaGeo:
okay, @naik-aakash ! 

@utf this would be ready to merge!

utf:
This looks great. Thanks @JaGeo @naik-aakash and all other contributors.

PR #764, opened 2024-03-05
Add nequip force field relax- and staticmaker and corresponding tests

JonathanSchmidt1:
## Summary

* Add a nequip force field relax- and staticmaker and the corresponding tests
Nequip is a quite popular force field code and I have been using it with atomate2 for a while so I thought it might be useful to the community.

* nequip (and its dependencies) required to run nequip force fields. Nequip is still one of the most popular equivariant force fields with 420 citations last year. There is conflict with the torch versions nequip (PyTorch >= 1.8, !=1.9, <=1.11.*.) and other force field packages demand however as far as I can see running nequip potentials with torch>2.0  seems to work fine.

## TODO (if any)

~~The force field I use for testing needs to be added to github. As it's 2.3 MB I was wondering if I just add it or do you have a procedure for the larger files?~~
I am not sure what's the best way to add nequip as a dependency as we do not want to install the torch version nequip demands.

## Checklist

title.

  type check your code.
pre-commit run --all ran succesfully. I only tested the new tests I added.


JaGeo:
I will just mention this PR here which has plans for the Relaxmaker as well #722

janosh:
very happy to see this! thanks for this PR @JonathanSchmidt1 👍 

> The force field I use for testing needs to be added to github. As it's 2.3 MB I was wondering if I just add it or do you have a procedure for the larger files?

that's larger than expected. how many parameters does your model have? if it's fitted to a single chemical system i would expect it to be smaller. maybe the checkpoint stores other things we don't need beside the model params?

> I am not sure what's the best way to add nequip as a dependency as we do not want to install the torch version nequip demands.

that's a good question! a `constraints` file might be one possible solution at least for `atomate2`'s own CI. given it's an optional dependency maybe that's all we need to solve. we could leave it to the user to create a working local environment. many might only want to run Nequip, not other FFs and so won't encounter any version conflicts anyway.

JaGeo:
@janosh I think it might be good to come up with a general approach for the forcefields. I feel there will soon be dependency conflicts. Do you have any suggestion?

JonathanSchmidt1:
> in any case, we could just run the tests with a randomly initialized model

@janosh Makes sense I added a very small model for testing.

> that's a good question! a constraints file might be one possible solution at least for atomate2's own CI. given it's an optional dependency maybe that's all we need to solve. we could leave it to the user to create a working local environment. many might only want to run nequip, not other FFs and so won't encounter any version conflicts anyway.

I noticed that the other force field packages do not have pinned python versions in their pyproject.toml anyway so I guess it might still work to just add nequip as a dependency else this should work  for the CI (just not sure whether it needs to be added to both the test.yml or also the deploy.yml for the CI): 
```
          pip install --no-dependencies nequip
          pip install --no-dependencies torch-runstats>=0.2.0
```

Personally, as someone who always has some very specific torch/cuda/dgl version in an environment it would be nicer anyway if atomate does not install torch or any other ml packages.

JonathanSchmidt1:
@janosh 
just adding nequip does not work:

```
The conflict is caused by:
    chgnet 0.3.4 depends on torch>=1.11.0
    matgl 1.0.0 depends on torch
    mace-torch 0.3.3 depends on torch>=1.12
    e3nn 0.4.4 depends on torch>=1.8.0
    nequip 0.5.6 depends on torch!=1.9.0, <1.13 and >=1.10.0
    chgnet 0.3.4 depends on torch>=1.11.0
    matgl 1.0.0 depends on torch
    mace-torch 0.3.3 depends on torch>=1.12
    e3nn 0.4.4 depends on torch>=1.8.0
    nequip 0.5.5 depends on torch!=1.9.0, <=1.12 and >=1.8
    chgnet 0.3.4 depends on torch>=1.11.0
    matgl 1.0.0 depends on torch
    mace-torch 0.3.3 depends on torch>=1.12
    e3nn 0.4.4 depends on torch>=1.8.0
    nequip 0.5.4 depends on torch!=1.9.0, <=1.12 and >=1.8
```

but adding this to the atomate[strict] installation in a clean conda environment worked and all the force field tests passed.
```
          pip install torch-runstats>=0.2.0
          pip install --no-dependencies nequip
```

janosh:
> Personally, as someone who always has some very specific torch/cuda/dgl version in an environment it would be nicer anyway if atomate does not install torch or any other ml packages.

i feel the same. i prefer for `pip install` not to mess with the versions of existing packages. it's fine to auto-add new packages that aren't installed yet but i think `atomate2` should try to leave versions of existing force field dependencies unaffected. which at present suggests leaving dep installations to the user and just worrying about our own CI

> ```sh
> pip install torch-runstats>=0.2.0
> pip install --no-dependencies nequip
> ```

didn't know `torch-runstats` is required but was also going to suggest trying `pip install --no-deps nequip`. seems like the right way to go for now

JonathanSchmidt1:
Seems fine now, I am just not sure if the addition of nequip and run-stats to the deploy.yml is necessary or not?

janosh:
> if the addition of nequip and run-stats to the deploy.yml is necessary or not?

don't think so

esoteric-ephemera:
This looks great @JonathanSchmidt1! Since, as @JaGeo mentioned, we've been refactoring the MLFF jobs in #722, it might make more sense to merge this PR first, and then I'll refactor the Nequip jobs

PR #768, opened 2024-03-07
limit python to 3.11

naik-aakash:
## Summary

Currently, on python 3.12, atomate2 installation fails due to some of the dependencies of pymatgen not supporting 3.12 yet. 
It seems better fix for now to strictly limit the python version in atomate2.

# Open question

There is still issue with ase with pip install. Would It be okay to strictly revert to `pip install git+https://gitlab.com/ase/ase.git` ?

Currently even with strict install ,  using forcefields based workflows is not possible without installing ase main branch.


janosh:
> some of the dependencies of pymatgen not supporting 3.12 yet.

could you specific which ones? i'm not aware of this. unless there are some real blockers, i think now would be a good time to add official 3.12 support to `pymatgen`

> Would It be okay to strictly revert to pip install git+https://gitlab.com/ase/ase.git

unfortunately, you can't release to pypi with urls in your package deps

naik-aakash:
Hi @janosh , seems the issue is with spglib 

image


# Steps to reproduce

1. Create conda env with python 3.12
2. `pip install atomate2`

janosh:
[`spglib` officially supports 3.12](https://github.com/spglib/spglib/blob/a66056afbe8434cecd7a127aa3ee7290d4a42b0b/pyproject.toml#L41). yours looks like a compilation error that you can maybe circumvent by installing a pre-compiled wheel rather than building yourself. you could also try installing with `uv` instead of `pip`. i'd be curious what happens there

naik-aakash:
Hi @janosh ,yes it supports but I just tried with `uv`, I still get same error

janosh:
just tried on macOS and worked:

```py
. ~/.venv/py312/bin/activate
uv pip install -U spglib
python3.12 -c "import spglib; print(spglib.__version__)"
>>> 2.3.1
```

so looks like its linux specific

JaGeo:
For testing, you could switch on the workflows on pymatgen for Python 3.12. We now had this issue in different contexts: atomate2 but also LobsterPy

naik-aakash:
Closing this as the new pymatgen version with 3.12 support  is released now. 
[#3685](https://github.com/materialsproject/pymatgen/pull/3685)

PR #780, opened 2024-03-19
Input set tutorial

JonathanSchmidt1:
## Summary

This is a first draft of additional documentation for the InputSets/Generators in atomate2.

@utf It would be great if you can give me some feedback whether this is the direction you had in mind.

Also there is of course documentation of input sets already in various places that I link to. The question is to what extent we want to "double"  the documentation. This mostly concerns the first half of what I wrote e.g. the basic introduction of InputFiles/Sets/Generators already mostly exists in pymatgen and a few paragraphs exist in [the workflow tutorial](https://github.com/materialsproject/atomate2/blob/33d747adfda16089d6794019828e3988c6a27a6e/docs/dev/workflow_tutorial.md). One sentence I just copied over for now. (`While generally the input classes are supposed to be part of `pymatgen` during development it is recommended to include them in `atomate2` at first to facilitate rapid iteration. Once mature, they can be moved to `pymatgen` or to a `pymatgen` [addon package](https://.org/addons).)

For now I am also mixing information for developers and users in one document. It might be possible to separate this but of course this would mean doubling some more information (e.g. I could separate out developer information and add it to [the workflow tutorial](https://github.com/materialsproject/atomate2/blob/33d747adfda16089d6794019828e3988c6a27a6e/docs/dev/workflow_tutorial.md). 

Two additional questions that popped up when I was looking at the different InputSets/Generators:
Vasp input files do not seem to be inheriting from InputFile. e.g. `class Poscar(MSONable)`, `class Incar(dict, MSONable)` etc... Any specific reason for this that I should discuss or is this just for historic reasons?

A bit random: What is the `previous_input` doing in `_get_input` in the cp2k input  sets/base.py. I don't see that it is used anywhere? https://github.com/materialsproject/atomate2/blob/33d747adfda16089d6794019828e3988c6a27a6e/src/atomate2/cp2k/sets/base.py#L326C1-L334C46


utf:
Hi @JonathanSchmidt1, thanks very much for this. This looks perfect to me. They tutorial will need to be updated based on #854 (the input sets have moved there and some of the syntax has changed very slightly). But once that's done I would be happy to merge.

JonathanSchmidt1:
Thank you for taking a look @utf . I will take care of it next week.

JonathanSchmidt1:
the error should be unrelated.

JaGeo:
If you fix the torch data version similar to #940 , it should work

JonathanSchmidt1:
@utf reminder

utf:
Thanks @JonathanSchmidt1

PR #717, opened 2024-02-13
[WIP] Atomate2 support in OpenMM

orionarcher:
## Summary

This PR ports over functionality from [atomate2-openmm](https://github.com/xperrylinn/atomate2-openmm) developed by @xperrylinn and myself. It includes:

* Support for setting up OpenMM `InputSets` using `pymatgen.io.openmm`
* Standard classical MD workflows operating on OpenMM inputsets
* Optionally recording trajectory and simulation state information to AWS
* Rich metadata on simulation setup for downstream analysis with Emmett

## Additional dependencies introduced

* openmm
* openff-toolkit
* mdanalysis
* pymatgen.io.openmm
* packmol
* openbabel

These are all necessary for the classical MD setup and execution workflow.

## TODO (if any)

* Tests are not fully passing yet in atomate2
* OpenMM input sets, which are used as the input and output of jobs, can be very large and a more compact solution is needed

## Checklist

Before a pull request can be merged, the following items must be checked:

* [ ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  The easiest way to handle this is to run the following in the **correct sequence** on
  your local machine. Start with running [`ruff`](https://docs.astral.sh/ruff) and `ruff format` on your new code. This will
  automatically reformat your code to PEP8 conventions and fix many linting issues.
* [ ] Doc strings have been added in the [Numpy docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.
* [ ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org) to
  type check your code.
* [x] Tests have been added for any new functionality or bug fixes.
* [ ] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more
efficient if you already fix most errors prior to submitting the PR. It is highly
recommended that you use the pre-commit hook provided in the repository. Simply run
`pre-commit install` and a check will be run prior to allowing commits.

## Questions:

Best way to deal with new requirements?

PR #782, opened 2024-03-20
Atomate2 OpenMM integration & broader classical MD framework

orionarcher:
Developed by @xperrylinn and @orionarcher

## Summary

This PR builds out support for OpenMM with a framework that could be extended to support other MD codes. Namely LAMMPS, Amber, and Gromacs. Rough visual example [here](https://www.figma.com/file/aIwwYaPOeRPytAVNwLewOS/MD-Workflow-Diagram?type=whiteboard&node-id=0%3A1&t=rXTgEcKkXIyKnzes-1)

### Core ideas:
 - Use the  `openff.interchange.Interchange` object as a core engine-agnostic representation of an MD simulation.
 - Define a generic `ClassicalMDTaskDocument` and generic `Interchange` creation functions, then hand off responsibility for evolving the simulation to workflows for each MD engine. Currently only OpenMM support is implemented.
 - Support low-throughput workflows by making workflows that can easily output to local directories.
 - Use `prev_task` to pass meta-data between jobs. Since OpenMM is programmatic, the core information is not stored in the local directory.
 - No `InputSets`. Since OpenMM is programmatic, `InputSets` and `InputGenerators` don't make conceptual sense. 

### What's implemented:
- Integration of Atomate2 and OpenFF for classical MD simulations
- Support for energy minimization, NPT, NVT, and temperature change jobs
- Implementation of annealing and production workflows
- Serialization and deserialization of OpenFF objects (Molecule, Topology, Interchange, Quantity) with `monty`
- Utility functions for generating OpenFF Interchange objects and creating molecule specifications

### To do / open questions:
- [x] Need to add create conditional imports that fail nicely, where are some examples of this?

### Future PR
- [x] Store trajectories in `additional_stores`, if available
- [x] The core `openff.interchange.Interchange` object can be larger than 16MB MongoDB doc limit. Need to implement a workaround by zipping the object and/or putting it on S3.
- [x] migrate schema dependencies to emmet
- [x] migrate utility dependencies to pymatgen

### Related PRs:
- PR to `pymatgen` would move most of `atomate2.classical_md.utils` upstream. [PR #3729](https://github.com/materialsproject/pymatgen/pull/3729)
- PR to `emmet` would move `atomate2.classical_md.schemas` upstream. [PR #975](https://github.com/materialsproject/emmet/pull/975)

### Example usage

```python
mol_specs_dicts = [
    {"smile": "CCO", "count": 10, "name": "ethanol"},
    {"smile": "O", "count": 200, "name": "water"},
]
inter_job = generate_interchange(mol_specs_dicts, mass_density=1)

production_maker = ProductionMaker(
    energy_maker=EnergyMinimizationMaker(),
    npt_maker=NPTMaker(steps=300000, pressure=1),
    anneal_maker=AnnealMaker.from_temps_and_steps(
        steps=1000000, anneal_temp=400, final_temp=300
    ),
    nvt_maker=NVTMaker(steps=5000000),
)

production_flow = production_maker.make(
    inter_job.output.interchange, 
    output_dir="my/directory/"
    prev_task=inter_job.output
)

wf = Flow([inter_job, production_flow])

run_locally(wf)

# will put outputs for the whole flow in "my/directory/"
```

* openmm
* openff-toolkit
* openff-interchange
* packmol
* openbabel

These are all necessary for the classical MD setup and execution workflow.

## Checklist

  type check your code.


orionarcher:
Hey @utf, this PR supercedes #717 and is now ready for feedback. Testing is passing locally but I need to fix some issues with the CI. I also need to flesh out the documentation/tutorials.

@rkingsbury might be of interest to you as well!

orionarcher:
The OpenForceField toolkit can only be installed via conda, making the current testing framework unsuitable. 

I initially tried migrating the whole test suite to use `micromamba` instead of pip. I got almost all of the tests passing but could not debug the tests for `forcefields`. Since I wasn't sure you'd want to migrate even if it worked, I decided to drop that line of work. Instead, I decided to exclude the `classical_md` tests from the `test` action and create a new action called `test_classical_md` that uses `micromamba` to install dependencies and then only executes the `classical_md` tests. This seems to be working.

orionarcher:
I am currently working on a builder for the classical_md code. I plan to stage the builder in `atomate2` for now. Long term, I plan to migrate most of the code in `utils.py` to `pymatgen` and the builder code and TaskDoc code to `emmet`, mirroring the organization for the rest of the MP codes.

orionarcher:
As far as I can tell the failing tests are upstream and not due to this PR.

rkingsbury:
Looks like a great contribution @orionarcher . I'm especially glad you included a tutorial as part of the PR. Is there any aspect you or @utf would find helpful for me to review specifically?

orionarcher:
From @rkingsbury, I'd love it if you could take a look at the pydantic schemas (now moved to emmet, [here](https://github.com/materialsproject/emmet/pull/975/files)). Your MD experience would add a useful perspective.

@utf, it'd be very helpful to get your feedback on the changes to `testing.yml`. I added a new action and want to make sure it looks good to you.

Overall, I think this is pretty much done and could be merged. The outstanding features aren't critical and could be added in a later PR.

rkingsbury:
> From @rkingsbury, I'd love it if you could take a look at the pydantic schemas (now moved to emmet, [here](https://github.com/materialsproject/emmet/pull/975/files)). Your MD experience would add a useful perspective.

Done! I left a review on the already merged PR here - https://github.com/materialsproject/emmet/pull/975#pullrequestreview-1994562429

orionarcher:
@utf would you mind taking a look at this?

I broke compatibility between emmet and atomate2 in the most recent commits but I will resolve it soon. If you'd prefer to look at something that is passing, please consider [commit d0683c9](https://github.com/materialsproject/atomate2/pull/782/commits/d0689c36a9c047e61d906b47ee8db0bc6c092d16) which has all of the core functionality.

docs/user/codes/openmm.md
```diff
@@ -0,0 +1,639 @@
...
+From here, we can do all sorts of very cool analysis, but that's beyond the
+scope of this tutorial. Consult the tutorials in SolvationAnalysis and MDAnalysis
+for more information.
+
+```python
+from atomate2.classical_md.core import ClassicalMDTaskDocument
+from emmet.builders.classical_md.utils import create_universe, create_solute
+from openff.interchange import Interchange
+
+ec_emc_taskdoc = ClassicalMDTaskDocument.parse_file("tutorial_system/taskdoc.json")
```
utf, line 488: Better to just get the task doc from the database rather than relying on folders.

pyproject.toml
```diff
@@ -57,6 +57,7 @@ forcefields = [
     "quippy-ase>=0.9.14",
     "torch<=2.2.1",       # incompatibility with dgl if newer versions are used
 ]
+classical_md = ["openmm-mdanalysis-reporter>=0.1.0"]
```
utf, line 60: Put all other requirements here.

src/atomate2/classical_md/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+def openff_mol_as_monty_dict(self: Molecule) -> dict:
+    """Convert a Molecule to a monty dictionary."""
+    mol_dict = self.to_dict()
+    mol_dict["@module"] = "openff.toolkit.topology"
+    mol_dict["@class"] = "Molecule"
+    return mol_dict
+
+
+Molecule.as_dict = openff_mol_as_monty_dict
```
utf, line 17: I'm not sure how I feel about this. Seems like it could break easily? The other options are: 
1. Add as_dict methods to openff classes for monty support.
2. Add support for openff to monty (doesn't seem likely given we tried to do something similar for ase and it was turned down).
3. Some kind of plugin system for monty where we can define some common serialisation routes for materials science objects.

For now we can leave this as is but I'd like to explore option 3 further (cc'ing @Andrew-S-Rosen @janosh @mkhorton)

src/atomate2/classical_md/__init__.py
```diff
@@ -0,0 +1,70 @@
+"""Module for classical md workflows."""
```
utf, line 1: To me this folder should not be called `classical_md` but instead `openff`. And the `openmm` subdirectory should be its own folder in the base dir. E.g., `atomate2/openmm`

pyproject.toml
utf, line 60: Add other requirements here.

src/atomate2/classical_md/openmm/flows/core.py
```diff
@@ -0,0 +1,159 @@
...
+    from pathlib import Path
+
+    from emmet.core.classical_md import ClassicalMDTaskDocument
+    from openff.interchange import Interchange
+
+    from atomate2.classical_md.openmm.jobs.base import BaseOpenMMMaker
+
+
+@dataclass
+class OpenMMFlowMaker:
```
utf, line 23: This is an interesting approach to generate a flow but makes sense in the context!

src/atomate2/classical_md/openmm/flows/core.py
```diff
@@ -0,0 +1,159 @@
...
+
+    name: str = "flexible"
+    tags: list[str] = field(default_factory=list)
+    makers: list[BaseOpenMMMaker | OpenMMFlowMaker] = field(default_factory=list)
+
+    def make(
+        self,
+        interchange: Interchange | bytes,
+        prev_task: ClassicalMDTaskDocument | None = None,
+        output_dir: str | Path | None = None,
```
utf, line 46: Can we remove output_dir from this. We should handle this another way.

src/atomate2/classical_md/openmm/flows/core.py
```diff
@@ -0,0 +1,159 @@
...
+            prev_task = job.output
+            jobs.append(job)
+
+        return Flow(
+            jobs,
+            output=jobs[-1].output,
+        )
+
+    @classmethod
+    def anneal_flow(
```
utf, line 90: Would it make sense to turn this into's its own class?

src/atomate2/classical_md/openmm/jobs/base.py
```diff
@@ -0,0 +1,494 @@
...
+from openmm.app import StateDataReporter
+from openmm.unit import kelvin, picoseconds
+
+from atomate2.classical_md.core import openff_job
+from atomate2.classical_md.utils import increment_name, task_reports
+
+if TYPE_CHECKING:
+    from openmm.app.simulation import Simulation
+
+OPENMM_MAKER_DEFAULTS = {
```
utf, line 32: What is the benefit of having these here rather than settings them in the `BaseOpenMMMaker` directly? It seems like this would be cleaner?

src/atomate2/classical_md/openmm/jobs/base.py
```diff
@@ -0,0 +1,494 @@
...
+        Whether to report velocities in the trajectory file.
+    traj_file_name : Optional[str]
+        The name of the trajectory file to save.
+    traj_file_type : Optional[str]
+        The type of trajectory file to save. Supports any output format
+        supported by MDAnalysis.
+    """
+
+    name: str = "base openmm job"
+    tags: list[str] | None = field(default=None)
```
utf, line 103: I.e., the `field` could be replaced with the default.

src/atomate2/classical_md/openmm/jobs/base.py
```diff
@@ -0,0 +1,494 @@
...
+        """
+        # this is needed because interchange is currently using pydantic.v1
+        if isinstance(interchange, bytes):
+            interchange = interchange.decode("utf-8")
+        if isinstance(interchange, str):
+            interchange = Interchange.parse_raw(interchange)
+        else:
+            interchange = copy.deepcopy(interchange)
+
+        dir_name = Path(
```
utf, line 156: I wouldn't expect to create a new directory? This will already be handled by the workflow manager (e.g., jobflow, jobflow-remote, or fireworks). I think we should remove this to be consistent with the other workflows.

src/atomate2/classical_md/openmm/jobs/base.py
```diff
@@ -0,0 +1,494 @@
...
+        elapsed_time = time.time() - start
+
+        self._update_interchange(interchange, sim, prev_task)
+
+        del sim
+
+        task_doc = self._create_task_doc(interchange, elapsed_time, dir_name, prev_task)
+
+        # write out task_doc json to output dir
+        with open(dir_name / "taskdoc.json", "w") as file:
```
utf, line 177: Why do we write the task doc? I'm not entirely opposed to this but it will get stored in the database anyway.

src/atomate2/classical_md/openmm/jobs/base.py
```diff
@@ -0,0 +1,494 @@
...
+            "`run_openmm` should be implemented by each child class."
+        )
+
+    def _resolve_attr(
+        self,
+        attr: str,
+        prev_task: OpenMMTaskDocument | None = None,
+        add_defaults: dict | None = None,
+    ) -> Any:
+        """Resolve an attribute and set its value.
```
utf, line 296: I think this would be even simpler if the defaults are on the base class, as then there are only two places to check.

tutorials/openmm_tutorial.ipynb
```diff
@@ -0,0 +1,605 @@
+{
```
utf, line 1: This is awesome. But I think this should live in the docs folder somewhere. E.g., for jobflow we have the `jobflow/docs/tutorials` folder that contains python notebooks, and these are automatically built into the docs.

utf (commented):
Hi @orionarcher, appologies it has taken quite some time to get to this PR.

Firstly, it looks fantastic. This is a serious amount of work, so thank you for contributing it to atomate2. I particularly like all the documentation you have added that covers the setup, running and even builders.

My comments are mainly related to bringing some quirks of these workflows more inline with the rest of atomate2. These are primarily:
- Updating the requirements/tests so that there isn't a separate mode of installing/testing the md workflows.
- Removing the folder control from the workflows. This is something that should be handled by the workflow manager. I'd strongly encourage you to use the database to get the outputs in your examples, rather than using the file system.
- Renaming classical_md to openff, and making openmm it's own folder in the base directory.

I'm happy to chat futher about any of these issues if you want to call. Once these changes are done (and all tests pass) I'd be happy to merge.

src/atomate2/classical_md/openmm/jobs/base.py
orionarcher: I agonized over this but it was ultimately the only solution I could think of. I'll respond here but the answer would be the same for this comment too:

> I think this would be even simpler if the defaults are on the base class, as then there are only two places to check.

When users are defining workflows I wanted to support easy inheritance between jobs. Priority is: settings from current maker > settings from previous task input >  a default value. If I set the defaults in BaseOpenMMMaker directly, there is no way of differentiating between settings from the current maker and the defaults. Maybe there is another solution, but the only one I could think of was external defaults. 

You can see the inheritance logic below.

```python
        if getattr(self, attr, None) is not None:
            attr_value = getattr(self, attr)  # if the defaults were set directly this would always return them.
        elif getattr(prev_input, attr, None) is not None:
            attr_value = getattr(prev_input, attr)
        else:
            attr_value = defaults.get(attr)
```

.github/classical_md_requirements.txt
```diff
@@ -0,0 +1,8 @@
+# for consistency, should align with pyproject.toml
+openmm>=8.1.0
```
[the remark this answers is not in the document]
orionarcher: The problem is that all of the dependencies in this file are not pip installable. They are only available on conda.

.github/workflows/testing.yml
```diff
@@ -74,7 +74,39 @@ jobs:
...
-        run: pytest --cov=atomate2 --cov-report=xml
+        run: pytest --cov=atomate2 --cov-report=xml --ignore=tests/classical_md
+
+      - uses: codecov/codecov-action@v1
+        if: matrix.python-version == '3.10' && github.repository == 'materialsproject/atomate2'
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: ./coverage.xml
+
+  test_md:
```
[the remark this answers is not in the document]
orionarcher: Many of the classical_md dependencies are not pip installable. We could consolidate the tests, but I'd need to modify the testing action to use micromamba instead of pip. I'll say more about this in another comment below.

docs/user/codes/openmm.md
```diff
@@ -0,0 +1,639 @@
+# Installing Atomate2 from source with OpenMM
+
+```bash
+# setting up our conda environment
+>>> conda create -n atomate2 python=3.11
+>>> conda activate atomate2
+
+# installing atomate2
+>>> pip install git+https://github.com/orionarcher/atomate2.git
```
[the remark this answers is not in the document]
orionarcher: Will update.

docs/user/codes/openmm.md
```diff
@@ -0,0 +1,639 @@
...
+The first job we need to create generates the `Interchange` object.
+To specify the system of interest, we use give it the SMILES strings,
+counts, and names (optional) of the molecules we want to include.
+
+
+```python
+from atomate2.classical_md.core import generate_interchange
+
+mol_specs_dicts = [
+    {"smile": "O", "count": 200, "name": "water"},
```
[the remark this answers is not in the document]
orionarcher: Good catch. Will update.

src/atomate2/classical_md/__init__.py
orionarcher: Agreed that it's not a perfect solution.

I think I'd be turned down for 1 & 2 both. I'd be happy to support option 3 in the future.

src/atomate2/classical_md/openmm/flows/core.py
orionarcher: Let me think about this. I agree that a monolithic flow maker isn't great design.

src/atomate2/classical_md/openmm/jobs/base.py
orionarcher: See my response above RE:

> What is the benefit of having these here rather than settings them in the BaseOpenMMMaker directly? It seems like this would be cleaner?

tutorials/openmm_tutorial.ipynb
orionarcher: Do you have a particular preference? I chose the location based on @jmmshn's tutorial PR but can definitely put it elsewhere. 

I would love to see automatically built jupyter notebooks for tutorials and would happily rewrite the docs into a notebook if it was available.

docs/user/codes/openmm.md
```diff
@@ -0,0 +1,639 @@
...
+        NPTMaker(n_steps=100),
+        OpenMMFlowMaker.anneal_flow(n_steps=150),
+        NVTMaker(n_steps=100),
+    ],
+)
+
+production_flow = production_maker.make(
+    elyte_interchange_job.output.interchange,
+    prev_task=elyte_interchange_job.output,
+    output_dir="./tutorial_system",
```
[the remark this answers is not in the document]
orionarcher: Might it still make sense to store this information?

I am considering removing the argument and then interally replacing:

```python
dir_name = Path(
    output_dir or getattr(prev_task, "dir_name", None) or Path.cwd()
)
```

with

```python
dir_name = Path.cwd()
```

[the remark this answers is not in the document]
utf: It still makes sense to put this in the task document but not to have it controllable by the job. You can see how we did this in the FHI-aims task document: https://github.com/materialsproject/atomate2/blob/e210b58b93d33b13af800dd001c9689f96a6eb0c/src/atomate2/aims/schemas/task.py#L326

orionarcher:
Hi @utf, I've addressed several of the points you mentioned and am still mulling over the best solution for others. In the meantime, I'd appreciate your help with two issues.

1. I migrated the testing to use micromamba and uv. This cuts the "Install dependencies" step from ~150 seconds to ~50 seconds. Further, the migration only broke a single [cclib-related-test](https://github.com/materialsproject/atomate2/actions/runs/10117212365/job/27981713046). If you could fix it or tag the appropriate code owner, that'd be much appreciated. 

For now, I am ignoring the `classical_md` tests because they are dependent on upstream changes in `emmet` that have not been merged in.

2. Currently, `calcs_reversed` is being used to store all calculations that preceded the task, which is not the intended behavior. I want to change that while preserving the job's history, ideally by storing all previous `job_uuid`. Is that possible? Can I access the `job_uuid` from inside the maker and store it in the `TaskDoc`? If not, how should I link the job to previous jobs?

src/atomate2/classical_md/openmm/jobs/base.py
orionarcher: Mainly to enable easier local analysis.

orionarcher:
I've now addressed all the points you brought up in your review. Most significantly:
- splitting `classical_md` into `openmm` and `openff`.
- replacing history tracking using calcs_reversed with a final `collect_jobs` job that gathers uuids and calcs
- reworking the CI and dependencies

Once emmet is adjusted and tests are passing, I think this will be ready to merge.

src/atomate2/openmm/jobs/base.py
```diff
@@ -0,0 +1,528 @@
...
+        OpenMMTaskDocument
+            The created task document.
+        """
+        maker_attrs = copy.deepcopy(dict(vars(self)))
+        job_name = maker_attrs.pop("name")
+        tags = maker_attrs.pop("tags")
+        state_file_name = self._resolve_attr("state_file_name", prev_task)
+        traj_file_name = self._resolve_attr("traj_file_name", prev_task)
+        traj_file_type = self._resolve_attr("traj_file_type", prev_task)
+        calc = Calculation(
```
utf, line 496: Ideally the sub task documents should be created in a function in `OpenMMTaskDocument`. E.g., `OpenMMTaskDocument.from_directories()`. Can you move this logic there? The role of the `_create_task_doc` would then be to just resolve the previous directory names etc.

src/atomate2/openmm/jobs/opls.py
```diff
@@ -0,0 +1,256 @@
+"""Utilities for working with the OPLS forcefield in OpenMM."""
```
utf, line 1: Currently this file is in the `jobs` folder but nothing here is a `job`. Perhaps this should be one directory up and called `utils.py`?

utf (commented):
Thanks @orionarcher, I'll get back to you about the failing test separately.

.github/workflows/testing.yml
```diff
@@ -31,58 +31,97 @@ jobs:
...
-          pip install --no-deps nequip==0.5.6
+          uv pip install .[strict,tests,abinit,classical_md]
+          uv pip install torch-runstats
+          uv pip install --no-deps nequip==0.5.6
 
       - name: Install pymatgen from master if triggered by pymatgen repo dispatch
         if: github.event_name == 'repository_dispatch' && github.event.action == 'pymatgen-ci-trigger'
-        run: pip install --upgrade 'git+https://github.com/materialsproject/pymatgen@${{ github.event.client_payload.pymatgen_ref }}'
+        run: |
+          micromamba activate a2
```
utf, line 72: Do we actually need all of these `micromamba active` commands. I think the first one should be enough.

.github/workflows/testing.yml
```diff
@@ -31,58 +31,97 @@ jobs:
...
       - name: Install dependencies
         run: |
+          micromamba activate a2
           python -m pip install --upgrade pip
           mkdir -p ~/.abinit/pseudos
           cp -r tests/test_data/abinit/pseudos/ONCVPSP-PBE-SR-PDv0.4 ~/.abinit/pseudos
-          pip install .[strict,tests,abinit]
-          pip install torch-runstats
-          pip install --no-deps nequip==0.5.6
+          uv pip install .[strict,tests,abinit,classical_md]
```
utf, line 65: No need to put `classical_md` here if your dependencies are already included in the `strict` set of requirements (I should look at copying the abinit dependencies into strict too).

pyproject.toml
```diff
@@ -59,6 +59,11 @@ forcefields = [
     "sevenn>=0.9.3",
     "torch<=2.2.1",       # incompatibility with dgl if newer versions are used
 ]
+classical_md = [
+    "mdanalysis>=2.7.0",
```
utf, line 63: Do these workflows actually require the latest version of mdanalysis etc? If so you can remove these pins from the `classical_md` section. You should keep them for the `strict` section however.

pyproject.toml
```diff
@@ -108,6 +113,9 @@ strict = [
     "sevenn==0.9.3",
     "torch==2.2.1",
     "typing-extensions==4.12.2",
+    "mdanalysis==2.7.0",
+    "openmm>=8.1.0",
```
utf, line 117: This should be pinned with `==`. Also there is a new version available (`8.1.1`)

utf (commented):
The failing test is very unusual - I wonder if @Andrew-S-Rosen has any idea's as it's related to cclib.

Have you tested installing the dependencies with `pip` rather than `uv`?

src/atomate2/openmm/jobs/base.py
orionarcher: I considered doing this originally but I think it'd be a headache. We could only reconstitute the information purely from the directories if we already had the current `task_doc` which is what we need to write. I'm sure there is some hack around but ultimately I feel this is where the logic belongs.

OpenMM has no input sets so the input logic is never fully encoded as a string. To get the task doc from a directory, we'd almost need to already have the taskdoc in the directory.

src/atomate2/openmm/jobs/base.py
```diff
@@ -0,0 +1,528 @@
...
+    report_velocities: bool | None = field(default=None)
+    traj_file_name: str | None = field(default=None)
+    traj_file_type: str | None = field(default=None)
+    embed_traj: bool | None = field(default=None)
+
+    @openff_job
+    def make(
+        self,
+        interchange: Interchange | OpenMMInterchange | bytes,
+        prev_task: OpenMMTaskDocument | None = None,
```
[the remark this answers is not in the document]
orionarcher: Yeah can do. I somewhat dislike having `prev_dir` because you have no idea what needs to be in the directory. But for the sake of minimizing database calls and consistency with other codes, I do think it makes sense.

src/atomate2/openmm/jobs/opls.py
orionarcher: `generate_openmm_interchange` is a job but I do agree some of this could be moved upstream to utils.

pyproject.toml
orionarcher: I believe so. They actually require `2.8.0` for some features to work but that hasn't been released yet.

> If so you can remove these pins

A little confused by this, shouldn't I keep the pins if the dependency is required? Did you mean "If not" or am I missing something?

pyproject.toml
Andrew-S-Rosen: What in the world..

Andrew-S-Rosen:
> The failing test is very unusual - I wonder if @Andrew-S-Rosen has any ideas as it's related to cclib.
> 
> Have you tested installing the dependencies with `pip` rather than `uv`?

Uh... weird! Both `main` and this PR are using the same version of cclib, which hasn't been updated since March. I'm trying to find the source of the issue but am not successful thus far. [I recognize that the cclib interface isn't really used by anything, so it would be a shame for that to hold up a PR...].

pyproject.toml
janosh: @peastman very curious what was the motivation behind removing old versions from pypyi?

Andrew-S-Rosen: Are we sure there were prior PyPI versions to begin with? :)

peastman: There weren't any.  The old package was an ancient front end for running simulations.  It hadn't been supported in many years.

The new packages are still experimental and shouldn't be relied on.  We're still exploring ways of creating PyPI packages.

orionarcher:
@utf is the best way forward to just disable the test here?

orionarcher:
And thank you for looking into it @Andrew-S-Rosen

.github/workflows/testing.yml
orionarcher: I tried removing them and ran into errors

.github/workflows/testing.yml
orionarcher: Removed

orionarcher:
@utf I believe I've addressed all comments and this is ready to merge.

orionarcher:
I've pinned the emmet pre-release to get tests passing. @utf are you good to merge this once the new emmet version is released?

.github/workflows/testing.yml
janosh: confirm you have to activate envs in every step, there's no propagation of shell state between job steps in GH Actions

src/atomate2/openff/core.py
```diff
@@ -0,0 +1,165 @@
...
+        force_field=ff_object,
+        topology=topology,
+        charge_from_molecules=[
+            tk.Molecule.from_json(spec.openff_mol) for spec in mol_specs
+        ],
+        allow_nonintegral_charges=True,
+    )
+
+    # currently not needed because ForceField isn't correctly supporting iterables
+
```
janosh, line 142: extra blank line between code and comment?

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    mol_specs = []
+
+    for spec in input_mol_specs:
+        if isinstance(spec, dict):
+            mol_specs.append(create_mol_spec(**spec))
+        elif isinstance(spec, MoleculeSpec):
+            mol_specs.append(copy.deepcopy(spec))
+        else:
+            raise TypeError("mol_specs must be a list of dicts or MoleculeSpec")
```
janosh, line 104: improves UX to print received type here

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+    mol_specs : List[MoleculeSpec]
+        A list of MoleculeSpecs to merge.
+
+    Returns
+    -------
+    List[MoleculeSpec]
+        A list of merged MoleculeSpecs with unique name and SMILES combinations.
+    """
+    mol_specs = copy.deepcopy(mol_specs)
+    merged_spec_dict: dict[tuple[str, str], MoleculeSpec] = {}
```
janosh, line 130: could use `merged_spec_dict=defaultdict(int)` to avoid the `if key in merged_spec_dict`

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Returns
+    -------
+    dict
+        A dictionary containing the normalized mass ratios of molecules in
+        the electrolyte solution.
+    """
+    # Check if all solvents have corresponding densities
+    solvent_densities = solvent_densities or {}
+    if set(solvents.keys()) > set(solvent_densities.keys()):
```
janosh, line 167:
```suggestion
    if set(solvents) > set(solvent_densities):
```

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+        else:
+            merged_spec_dict[key] = spec
+    return list(merged_spec_dict.values())
+
+
+def calculate_elyte_composition(
+    solvents: dict[str, float],
+    salts: dict[str, float],
+    solvent_densities: dict = None,
+    solvent_ratio_dimension: str = "mass",
```
janosh, line 144:
```suggestion
    solvent_ratio_dimension: Literal["mass", "volume"] = "mass",
```

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+    mws = []
+    for smile in species:
+        mol = tk.Molecule.from_smiles(smile, allow_undefined_stereo=True)
+        mws.append(sum([masses[atom.atomic_number] for atom in mol.atoms]))
+
+    mol_ratio = np.array(list(species.values())) / np.array(mws)
+    mol_ratio /= sum(mol_ratio)
+    return {
+        smile: int(np.round(ratio * n_mol))
+        for smile, ratio in zip(species.keys(), mol_ratio)
```
janosh, line 234:
```suggestion
        for smile, ratio in zip(species, mol_ratio)
```

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+    n_mol = (total_mass / mean_mw) * na
+
+    # Calculate the number of moles needed for each species
+    mol_ratio = np.array(list(species.values())) / np.array(mws)
+    mol_ratio /= sum(mol_ratio)
+
+    # Convert moles to number of molecules
+    return {
+        smile: int(np.round(ratio * n_mol))
+        for smile, ratio in zip(species.keys(), mol_ratio)
```
janosh, line 278:
```suggestion
        for smile, ratio in zip(species, mol_ratio)
```

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+        Number of each species needed to fill the box with the given density.
+    """
+    masses = {el.Z: el.atomic_mass for el in Element}
+
+    na = 6.02214076e23
+    volume = (side_length * 1e-7) ** 3  # Convert from nm3 to cm^3
+    total_mass = volume * density  # grams
+
+    # Calculate molecular weights
+    mws = []
```
janosh, line 264: would use a longer more descriptive name like `mol_weights: list[float] = []`

src/atomate2/openff/utils.py
```diff
@@ -0,0 +1,304 @@
...
+) -> list[dict]:
+    """Create lists of mol specs from just counts. Still rudimentary."""
+    spec_dicts = []
+    for smile, count in counts.items():
+        spec_dict = {
+            "smile": smile,
+            "count": count,
+            "name": name_lookup.get(smile, smile),
+        }
+        if bool(re.search(r"[+-]", smile)):
```
janosh, line 296:
```suggestion
        if re.search(r"[+-]", smile):
```

src/atomate2/openmm/jobs/core.py
```diff
@@ -0,0 +1,221 @@
...
+    pressure : float
+        The pressure of the simulation in atmospheres.
+        Default is 1 atm.
+    pressure_update_frequency : int
+        The number of steps between pressure
+        update attempts.
+    """
+
+    name: str = "npt simulation"
+    n_steps: int = 1000000
```
janosh, line 87:
```suggestion
    n_steps: int = 1_000_000
```

src/atomate2/openmm/jobs/core.py
```diff
@@ -0,0 +1,221 @@
...
+    """A maker class for performing NPT (isothermal-isobaric) simulations using OpenMM.
+
+    This class inherits from BaseOpenMMMaker, only new attributes are documented.
+
+    Attributes
+    ----------
+    name : str
+        The name of the NPT simulation job. Default is "npt simulation".
+    n_steps : int
+        The number of simulation steps. Default is 1000000.
```
janosh, line 77:
```suggestion
        The number of simulation steps. Default is 1,000,000.
```

src/atomate2/openmm/jobs/core.py
```diff
@@ -0,0 +1,221 @@
...
+    pressure : float
+        The pressure of the simulation in atmospheres.
+        Default is 1 atm.
+    pressure_update_frequency : int
+        The number of steps between pressure
+        update attempts.
```
janosh, line 83: accidental line break?

src/atomate2/openmm/jobs/core.py
```diff
@@ -0,0 +1,221 @@
...
+    temp_steps : Optional[int]
+        The number of steps over which the temperature is raised, by
+        default will be set to steps / 10000.
+    starting_temperature : Optional[float]
+        The starting temperature of the simulation.
+        If not provided it will inherit from the previous task.
+    """
+
+    name: str = "temperature change"
+    n_steps: int = 1000000
```
janosh, line 177: same here, more readable with thousands seps imo

src/atomate2/openmm/jobs/generate.py
```diff
@@ -0,0 +1,323 @@
...
+    def assign_partial_charges(self, mol_or_method: tk.Molecule | str) -> None:
+        """Assign partial charges to the XMLMoleculeFF object.
+
+        Parameters
+        ----------
+        mol_or_method : Union[tk.Molecule, str]
+            If a molecule is provided, it must have partial charges assigned.
+            If a string is provided, openff_toolkit.Molecule.assign_partial_charges
+            will be used to generate the partial charges.
+
```
janosh, line 135: blank line

tests/openff_md/conftest.py
```diff
@@ -0,0 +1,102 @@
...
+from openff.toolkit import ForceField
+from openff.units import unit
+
+from atomate2.openff.utils import create_mol_spec, merge_specs_by_name_and_smiles
+
+
+@pytest.fixture
+def temp_dir():
+    with tempfile.TemporaryDirectory() as temp_dir:
+        yield Path(temp_dir)
```
janosh, line 19: redundant fixture, [`pytest` ships its own `tmp_path`](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture)

janosh (changes requested):
huge amount of work here! 👍

src/atomate2/openff/utils.py
orionarcher: My preference is to leave as is, the values are tuples with no sensible default

orionarcher:
This is now pinned to the most recent version of emmet and ready to merge @utf. 

Your previous suggestion to split out OpenMM and OpenFF was very good. They are now independent and the OpenMM workflows support MLFFs and can return structures to better interoperate with the rest of Atomate2.

.github/workflows/testing.yml
```diff
@@ -31,51 +31,58 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Check out repo
+        uses: actions/checkout@v4
 
-      - uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
+      - name: Set up micromamba
```
janosh, line 45: @utf i think it could help to have the `micromamba`-dependent MD CI be its own job or even its own `test-md.yml` workflow. that would enable only running the MD tests when any MD source files change and would also uncouple the remaining CI from any install/env issues `micromamba` might encounter

janosh (approved):

.github/workflows/testing.yml
utf: I'd be happy with that solution!

utf:
This really is fantastic. Thank you very much @orionarcher. @janosh, once you're happy I think we can merge.

.github/workflows/testing.yml
orionarcher: Using micromamba saves us from manually building enumlib so I favor keeping it in the main CI.

rkingsbury:
Amazing work, thanks for sticking with this @orionarcher !

Issue #781, opened 2024-03-19
BUG: custom CHGNET model in PhononFlow throws `jsanitize` error


badw:
**Describe the bug**

Including a custom CHGNet model when doing PhononFlow (locally) with CHGNETRelaxMaker and CHGNetStaticMaker results in an error 

```
(from monty.json) 

       689 return jsanitize(
--> 690     obj.as_dict(),
       691     strict=strict,
       692     allow_bson=allow_bson,
       693     enum_values=enum_values,
       694     recursive_msonable=recursive_msonable,
       695 )

AttributeError: 'Tensor' object has no attribute 'as_dict'
```

_see full error output below_

**To Reproduce**

This was following the jupyter notebook: https://jageo.github.io/TutorialAtomate2Forcefields/phonon.html
```
from atomate2.forcefields.jobs import CHGNetRelaxMaker,CHGNetStaticMaker
from pymatgen.core import Structure
from chgnet.model import CHGNet

custom_model = CHGNet.from_file('custom_chgnet_model.pth.tar')

struct = Structure.from_file('teststructure.vasp')

phonon_flow = PhononMaker(
                          min_length=15.0, 
                          store_force_constants=False,
                          bulk_relax_maker=CHGNetRelaxMaker(optimizer_kwargs={'model':custom_model},
                                                            relax_kwargs={'fmax':0.00001}), 
                          static_energy_maker=CHGNetStaticMaker(optimizer_kwargs={'model':custom_model)
                          ).make(structure=struct)
```
and then:

``` 
from jobflow import run_locally

run_locally(phonon_flow, create_folders=False)

from pymatgen.phonon.bandstructure import PhononBandStructureSymmLine
from pymatgen.phonon.dos import PhononDos
from pymatgen.phonon.plotter import PhononBSPlotter, PhononDosPlotter
from jobflow import SETTINGS

store = SETTINGS.JOB_STORE
store.connect()
``` 

**Expected behavior**

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[10], line 3
      1 from jobflow import run_locally
----> 3 run_locally(phonon_flow, create_folders=False)
      5 from pymatgen.phonon.bandstructure import PhononBandStructureSymmLine
      6 from pymatgen.phonon.dos import PhononDos

[...]

PR #783, opened 2024-03-22
 Simplify phonon `get_supercell_size()` and test clean up

janosh:
- include method (`vasp`|`force_field`) in `PhononMaker` test names
- simplify phonon `get_supercell_size()` job
- better use `pytest` structure fixtures to reduce duplication


janosh:
waiting on `uv` release https://github.com/astral-sh/uv/issues/313#issuecomment-2014268571

janosh:
looks like we can successfully migrate CI deps installation to `uv`. @utf are you ok with this? should give a decent speed up as in https://github.com/materialsproject/pymatgen/pull/3675#issuecomment-1983033505

janosh:
`uv` install time is ~40s vs [`pip`'s ~150s](https://github.com/materialsproject/atomate2/actions/runs/8397746376/job/23001572012), so a 3.7x speed up

janosh:
looks like the test failures could be related to recent `jsanitize` changes in https://github.com/materialsvirtuallab/monty/pull/638. @Andrew-S-Rosen do you happen to know a good fix here?

Andrew-S-Rosen:
@janosh: I must admit, I can't see how that PR would cause the issue here. The command being called here is `jsanitize(obj, strict=True, allow_bson=True)`, but the logic I added only kicks in if `recursive_msonable=True` (which is not the case by default). I also don't think replacing an `isinstance()` check with `try`/`except` would trigger this error in particular. That said, the timing is suspicious. 😅  There was also https://github.com/materialsproject/maggma/pull/930 in maggma, but that doesn't seem to be called here. 

Might help if we knew which dependencies changed. Is there not a `requirements.txt` file with pinned versions used in CI?

JaGeo:
I think we already talked about a similar issue here ( #781 , not the CHGNet issue but the one of the band structure workflow). I think emmet was at fault.

JaGeo:
See https://github.com/materialsproject/atomate2/issues/781#issuecomment-2007490186

janosh:
@Andrew-S-Rosen my bad, should have looked more closely at the stack trace. was going mostly by temporal correlation 😅 

@JaGeo thanks for the pointer! i missed that discussion 🙏

PR #789, opened 2024-03-25
Add `fix_symmetry: bool = False` option to forcefield relax makers

JonathanSchmidt1:
## Summary

* Adds a fix_symmetry option to the Relaxer class for force fields, and the corresponding RelaxMakers that use it (Mace, Nequip, Gap). It's not available for m3gnet and chgnet that use their own internal relaxer. This is very useful when comparing force field relaxations to dft relaxations as the latter usually are fixing the symmetry.
* added a new test to force_fields/test_utils.py that confirms that fix_symmetry keeps the symmetry during an example relaxation while the fix_symmetry=False leads to a change in space group.
* added fix_symmetry as an option to one of the relaxer test jobs

As far as I can see https://github.com/materialsproject/atomate2/pull/722 also does not have any changes conflicting with this one.
Maybe @esoteric-ephemera  could confirm this or do you foresee any issues. 
None

## TODO (if any)
Right now the symprec is fixed to 0.01.  One could add  an additional symprec parameter or fix_sym_kwargs to allow users to change it. I would suggest adding one of the two options symprec/fix_sym_kwargs but I have no opinion on which one

## Checklist

  type check your code.
checked that the force field tests pass


esoteric-ephemera:
Thanks @JonathanSchmidt1! This looks fine to me, might just need to add docstr's and such when the MD PR is merged

tests/forcefields/test_jobs.py
```diff
@@ -236,7 +242,6 @@ def test_nequip_relax_maker(
             "model_path": test_dir / "forcefields" / "nequip" / "nequip_ff_sr_ti_o3.pth"
         },
     ).make(sr_ti_o3_structure)
-
     # run the flow or job and ensure that it finished running successfully
```
janosh, line 245: i think we're missing a new assert here for the spacegroup of the relaxed structure?

janosh (changes requested):
thanks @JonathanSchmidt1 just resolved the merge conflicts here now that #722 is merged

tests/forcefields/test_jobs.py
JonathanSchmidt1: I added the fix symmetry options to the test_nequip_relaxmaker to confirm the functionality of providing the option to the RelaxMaker. We can add the assert symmetry but the problem is this STO force field just goes to cubic anyway so the symmetry was staying the same (or higher symmetry) with and without the fix symmetry option.
So I used this test case instead in the test_utils that I think you removed when merging 
```
@pytest.mark.parametrize(("fix_symmetry"), [(True), (False)])
def test_fix_symmetry(fix_symmetry):
    # adapted from the example at https://wiki.fysik.dtu.dk/ase/ase/constraints.html#the-fixsymmetry-class
    relaxer = Relaxer(
        calculator=LennardJones(), relax_cell=True, fix_symmetry=fix_symmetry
    )
    atoms_al = bulk("Al", "bcc", a=2 / 3**0.5, cubic=True)
    atoms_al = atoms_al * [2, 2, 2]
    atoms_al.positions[0, 0] += 1.0e-7
    symmetry_init = check_symmetry(atoms_al, 1.0e-6, verbose=True)
    final_struct = relaxer.relax(atoms=atoms_al)["final_structure"]
    symmetry_final = check_symmetry(
        AseAtomsAdaptor.get_atoms(final_struct), 1.0e-6, verbose=True
    )
    if fix_symmetry:
        assert symmetry_init["number"] == symmetry_final["number"]
    else:
        assert symmetry_init["number"] != symmetry_final["number"]
```
We can add the assert statement concerning the space group to the nequip test but we also need  a test case like the one above where the symmetry actually changes.

janosh: oops, was a big hasty here and just ran `git merge -X theirs main`. didn't notice this swallowed the test 😅. now restored

janosh:
do we foresee other sym kwargs besides `symprec`? given `pymatgen`/`spglib` has just the 1 parameter, i'm guessing no. so leaning towards adding `symprec` as a top-level keyword.

janosh:
> @janosh: There's also angle_tolerance for spglib

good point! i think 2 top-level keywords is still okay but i'm easy either way

JonathanSchmidt1:
It's a good point but as I am using the FixSymmetry Constraint from ase angle tolerance is not an option at the moment. We would have to write our own constraint class or suggest ase to change theirs.
For now  I just added a symprec option.
I also noticed that the RelaxMakers were not working with fix symmetry anymore after the merge so I fixed and added additional tests for the chgnet and macerelaxmakers to hopefully make sure it stays working.
I also added fix_symmetry and symprec to the task_doc

JonathanSchmidt1:
Ps: my latest commit does not seem to show up here let's see if it changes today or if there is some issue

tests/forcefields/test_jobs.py
```diff
@@ -138,36 +138,73 @@ def test_mace_static_maker(si_structure: Structure, test_dir: Path, model):
...
+        responses = run_locally(job, ensure_success=True)
+        output1 = responses[job.uuid][1].output
+        assert output1.is_force_converged
+        final_space_group = SpacegroupAnalyzer(
+            output1.output.structure
+        ).get_space_group_number()
+        if fix_symmetry and symprec == 1e-2:
+            assert initial_space_group == final_space_group
+        else:
+            assert initial_space_group != final_space_group
```
janosh, line 182: would it make sense to split the large `if/else` in this test into 2 tests? one for `ba_ti_o3_structure` testing `fix_symmetry` and one for `si_structure`? i think it might help readability

JonathanSchmidt1:
sry forgot to run the md tests as well again. Now all tests pass again on my machine. I guess the question is still whether it was correct to change the ForceFieldTaskDocument or if there is some other way to save the fix_symmetry/symprec

esoteric-ephemera:
Regarding the ForceFieldTaskDocument: changes to the schema should be compatible across task doc schemas. Running:
```
from atomate2.forcefields.schemas import ForceFieldTaskDocument

taskdoc = TaskDoc(**<ForceFieldTaskDocument>.model_dump())
ff_taskdoc = ForceFieldTaskDocument(**taskdoc.model_dump())
```
should produce no errors. Since both the pydantic models for these docs permit extra top-level fields, I imagine this is OK but good to check

JonathanSchmidt1:
> Regarding the ForceFieldTaskDocument: changes to the schema should be compatible across task doc schemas. Running:
> 
> ```
> from emmet.core.tasks import TaskDoc
> from atomate2.forcefields.schemas import ForceFieldTaskDocument
> 
> taskdoc = TaskDoc(**<ForceFieldTaskDocument>.model_dump())
> ff_taskdoc = ForceFieldTaskDocument(**taskdoc.model_dump())
> ```
> 
> should produce no errors. Since both the pydantic models for these docs permit extra top-level fields, I imagine this is OK but good to check

thank you, that worked without issues so I will assume it's fine.

tests/forcefields/test_jobs.py
JonathanSchmidt1: Right now we have a test for chgnet, mace and the relaxer itself so I consider this is resolved as well.

tests/forcefields/test_jobs.py
```diff
@@ -232,6 +248,13 @@ def test_mace_relax_maker(
         assert output1.output.energy == approx(-0.051912, rel=1e-4)
         assert output1.output.n_steps == 4
 
+    # get space group number of input structure
+    _, final_spg_num = output1.output.structure.get_space_group_info()
+    if fix_symmetry:
+        assert init_spg_num == final_spg_num == 12
```
janosh, line 254: @JonathanSchmidt1 this test is failing which seems to suggest `fix_symmetry` doesn't work for MACE? maybe i'm missing something?

JonathanSchmidt1: Just checked on my phone so I am sorry if I overlooked sth but it looks to me like the energy checks are failing before right now. Two ideas concerning this:
When Fixing the symmetry the structure is first refined and second you changed the fmax both might lead to a different energy.
If the symmetry check is failing later on which I cannot see on the phone, it might be because we are not determining the initial symmetry with the same symprec  in the si-test as the one used for fix_symmetry. I also think the angle tolerance in pymatgen might not be the default one used in ase ( 5 vs -1 I believe).

 Gonna take a look later in the evening to see if I can find the problem.

> =========================== short test summary info ============================
> FAILED tests/forcefields/test_jobs.py::test_mace_relax_makera[/home/runner/work/atomate2/atomate2/tests/test_data/forcefields/mace/MACE.model-True-True-0.01] - assert -0.07111746072769165 == -0.0526856 ± 5.3e-03
>   
>   comparison failed
>   Obtained: -0.07111746072769165
>   Expected: -0.0526856 ± 5.3e-03
> FAILED tests/forcefields/test_jobs.py::test_mace_relax_makera[/home/runner/work/atomate2/atomate2/tests/test_data/forcefields/mace/MACE.model-True-False-0.01] - assert -0.0711173564195633 == -0.0526856 ± 5.3e-03
>   
>   comparison failed
>   Obtained: -0.0711173564195633
>   Expected: -0.0526856 ± 5.3e-03
> FAILED tests/forcefields/test_jobs.py::test_mace_relax_makera[/home/runner/work/atomate2/atomate2/tests/test_data/forcefields/mace/MACE.model-True-True-0.1] - assert -0.07111746072769165 == -0.0526856 ± 5.3e-03
>   
>   comparison failed
>   Obtained: -0.07111746072769165
>   Expected: -0.0526856 ± 5.3e-03
> FAILED tests/forcefields/test_jobs.py::test_mace_relax_makera[/home/runner/work/atomate2/atomate2/tests/test_data/forcefields/mace/MACE.model-False-True-0.01] - assert -0.06772968173027039 == -0.051912 ± 5.2e-06
>   
>   comparison failed
>   Obtained: -0.06772968173027039
>   Expected: -0.051912 ± 5.2e-06
> FAILED tests/forcefields/test_jobs.py::test_mace_relax_makera[/home/runner/work/atomate2/atomate2/tests/test_data/forcefields/mace/MACE.model-False-False-0.01] - assert -0.06772962212562[561](https://github.com/materialsproject/atomate2/actions/runs/8586508423/job/23529051664?pr=789#step:6:562) == -0.051912 ± 5.2e-06
>   
>   comparison failed
>   Obtained: -0.06772962212[562](https://github.com/materialsproject/atomate2/actions/runs/8586508423/job/23529051664?pr=789#step:6:563)561
>   Expected: -0.051912 ± 5.2e-06
> FAILED tests/forcefields/test_jobs.py::test_mace_relax_makera[/home/runner/work/atomate2/atomate2/tests/test_data/forcefields/mace/MACE.model-False-True-0.1] - assert -0.06772968173027039 == -0.051912 ± 5.2e-06
>   
>   comparison failed
>   Obtained: -0.06772968173027039
>   Expected: -0.051912 ± 5.2e-06

janosh: yeah, sorry didn't bother updating the energy in case the changes will be reverted. just commented out the energy assert locally to reach the symmetry check which failed.

pymatgen defaults to [`angle_tolerance=5`](https://github.com/materialsproject/pymatgen/blob/a2ab944bf8e1b90e51ca2b65f853972392177286/pymatgen/symmetry/analyzer.py#L56) like you said while [all the `ase` call sites](https://gitlab.com/ase/ase/-/blob/0ca67f36ff4df1be8790742d005c8337021164f4/ase/spacegroup/symmetrize.py#L93) for `spglib.get_symmetry_dataset` i could find no `angle_tolerance` is passed so the [default and recommended -1](https://spglib.readthedocs.io/en/stable/api/python-api/spglib/spglib.html#spglib.get_symmetry) is used

> Symmetry search tolerance in the unit of angle deg. Normally it is not recommended to use this argument. See a bit more detail at [angle_tolerance](https://spglib.readthedocs.io/en/stable/variable.html#variables-angle-tolerance). If the value is negative, an internally optimized routine is used to judge symmetry.

doesn't seem ideal that `pymatgen` defaults to 5. but would break stuff to change it now...
probably best to just pass `si_structure.get_space_group_info(angle_tolerance=-1)` here and leave pmg as is

JonathanSchmidt1: I think the problem might be the following While the space group staying the same means the symmetries are conserved we can of course arrive at a higher symmetry during optimization which changes the space group but is still the correct results for the fix_symmetry setting. Consequently we need to check if the previous symmetry operations are a subset of the new symmetry operations (This is also how the test of Fix_symmetry in ase is done). 
```
    from pymatgen.io.ase import AseAtomsAdaptor
    from ase.spacegroup.symmetrize import check_symmetry, is_subgroup, FixSymmetry
    si_atoms = AseAtomsAdaptor.get_atoms(si_structure)
    symmetry_ops_init = check_symmetry(si_atoms, symprec=1.0e-3)
    si_atoms_final = AseAtomsAdaptor.get_atoms(output1.output.structure)
    symmetry_ops_final = check_symmetry(si_atoms_final, symprec=1.0e-3)
    if fix_symmetry:
        assert is_subgroup(symmetry_ops_init, symmetry_ops_final)
    else:
        assert not is_subgroup(symmetry_ops_init, symmetry_ops_final)
```

JonathanSchmidt1: I pushed a commit with this but left the rest of the test unchanged so it will still be failing.
If you think it's correct now we can fix the rest.

janosh: sorry for letting this slip! that makes sense and thanks for pushing a fix. just updated the test assertions to match our changes

janosh (approved):

PR #790, opened 2024-03-25
Update LOBSTER instruction

JaGeo:
## Summary

Add more documentation, to remind users about the option to pass the number of threads to the lobster command within one line.


JaGeo:
Close as this currently does not work together with custodian.

PR #791, opened 2024-03-28
Add dir_name to ForceFieldMaker and **task_document_kwargs to from_ase_compatible_result

QuantumChemist:
I need to write the trajectory of a GAPrelax run into a file and then later access the directories of the files. That's why I added dir_name to ForceFieldRelaxMaker and **task_document_kwargs to ForceFieldTaskDocument.from_ase_compatible_result because as far as I can tell, that hasn't been implemented like that yet.

PR #792, opened 2024-03-29
Add flexible filename and extension to phonon new_plotter

QuantumChemist:
Finally had some time to do this.

I added a way to set a flexible filename and extension for the phonon band structure and DOS new_plotter object.

This PR also makes this PR https://github.com/materialsproject/atomate2/pull/624 obsolete.


QuantumChemist:
according to @janosh's suggestion here https://github.com/materialsproject/atomate2/pull/624#discussion_r1406637380 :)

QuantumChemist:
@janosh I hope it's ok if I also ask whether this PR is ok like that?

src/atomate2/common/schemas/phonons.py
```diff
@@ -359,9 +359,9 @@ def from_forces_born(
             filename_band_yaml, labels_dict=kpath_dict, has_nac=born is not None
         )
         new_plotter = PhononBSPlotter(bs=bs_symm_line)
-
         new_plotter.save_plot(
-            "phonon_band_structure.eps", units=kwargs.get("units", "THz")
+            filename=kwargs.get("filename_BS", "phonon_band_structure.eps"),
```
janosh, line 363: let's stick with snake_case here: `filename_bs` and `filename_dos`

tests/common/schemas/test_phonons.py
```diff
@@ -58,3 +59,31 @@ def test_phonon_bs_dos_doc():
...
+    phojob = PhononMaker(
+        bulk_relax_maker=CHGNetRelaxMaker(steps=5),
+        phonon_displacement_maker=CHGNetStaticMaker(),
+        static_energy_maker=CHGNetStaticMaker(),
+        store_force_constants=False,
+        generate_frequencies_eigenvectors_kwargs={
+            "filename_BS": "phonon_BS_test.png",
+            "filename_DOS": "phonon_DOS_test.pdf",
+        },
+    ).make(structure=si_structure)
```
janosh, line 79: can we avoid running the workflow given we're only testing the `from_forces_born` method?

janosh (changes requested):
no problem. similar comment re keeping test time low as in #791

tests/common/schemas/test_phonons.py
QuantumChemist: I have moved the tests now to `tests/forcefields/flows/test_phonon.py`.  It's not a schemas test anymore, then, but that way, we can avoid unnecessary run time.

QuantumChemist:
@janosh I implemented the requested changes and hope it's ok that way!

janosh (approved):
thanks @QuantumChemist, looks great!

PR #794, opened 2024-03-30
Warn on non-default `ATOMATE2_CONFIG_FILE` that's not found

janosh:
just a small QoL improvement to warn users when they set a custom `ATOMATE2_CONFIG_FILE` but the file doesn't exist

@utf i added tests to assert
- no warning if config path is default and file does not exist
- warning if config path is non-default and file does not exist

also  used `monkeypatch.setenv("ATOMATE2_CONFIG_FILE", ...)` which is auto-reverted on test-exit to avoid polluting other tests

### DIFF
--- a/src/atomate2/settings.py
+++ b/src/atomate2/settings.py
@@ -10,6 +10,7 @@ from pydantic import Field, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 _DEFAULT_CONFIG_FILE_PATH = "~/.atomate2.yaml"
+_ENV_PREFIX = "atomate2_"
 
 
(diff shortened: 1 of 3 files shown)


src/atomate2/settings.py
```diff
@@ -10,6 +10,7 @@
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 _DEFAULT_CONFIG_FILE_PATH = "~/.atomate2.yaml"
+_ENV_PREFIX = "atomate2_"
```
janosh, line 13: should this be upper-cased?

janosh:
CI failure caused by `pytorch==2.2.2` / `dgl` incompat which will be fixed in #722

Issue #800, opened 2024-04-04
FEATURE: GW workflow with VASP


yanghan234:
Hi all,

After discussion with @mkhorton and @utf, we are planning to add a GW workflow based on VASP. Please let us know if you think this is a good thing to do.

Also, we want to ensure our efforts are in harmony with the broader development plans. So, please let us know if other contributors are already working on this, to avoid any potential overlap.

Thank you!

Han.

utf:
Hi @yanghan234 and @mkhorton. Thanks again for planning to contribute this feature. I don't think anyone else is working on it, so please go ahead and submit a PR when you are ready.

It might be best to submit a WIP PR as soon as you're ready so any issues can be flagged earlier.

yanghan234:
Hi Gian-Marco and Tathagata, currently I'm working on implementing a GW workflow with VASP (almost done). It sounds like we're in a great place if you're focusing on GW in abinit at the moment. Also, I'm totally open to teaming up on the VASP GW workflow if you're interested. Let me know what you think!

tatha0003:
Hi Han, It would be great to team up for the development of these workflows.

JaGeo:
I will close this issue as it seems to be completed.

PR #808, opened 2024-04-17
Feat: GW workflow with VASP

yanghan234:
## Summary

This is a work-in-progress PR to implement a GW workflow with VASP, as proposed in #800.

## TODO 

- [x] Debug current code
- [x] Add tests
- [ ] Benchmark the settings


yanghan234:
Tagging @mkhorton and @utf for awareness.

yanghan234:
This PR is ready for review now. In this PR, I have implemented a simple workflow to run GW calculations in atomate2 with the setup defined in the [MVLGWSet](https://github.com/materialsproject/pymatgen/blob/1ac6c6863a0546454194ba73d548a374866bf062/src/pymatgen/io/vasp/sets.py#L1992) in pymatgen. 

Here I summarize what has been done in more detail:
- [x] Implement the static, diag and GW types of calculations with the setup compatible with MVLGWSet.
- [x] Added a test case along with the reference test data.
- [x] Modified runtypes in emmet to accomodate for GW calculations, and the PR is in [Feat: Add run types for GW calculations](https://github.com/materialsproject/emmet/pull/1099). The PR in emmet must be merged first, before this PR can be merged.

Please note that benchmarks of the setups are *not* done yet. One can only get the level of accuracy at the MVLGWSet.

Tagging @mkhorton and @utf for awareness.

src/atomate2/vasp/flows/core.py
```diff
@@ -171,6 +174,62 @@ def make(self, structure: Structure, prev_dir: str | Path | None = None) -> Flow
         return Flow(jobs, outputs, name=self.name)
 
 
+@dataclass
+class MVLGWBandStructureMaker(Maker):
```
utf, line 178: I think these files should go in `vasp/flows/mvl.py` since they are using the MVL input sets, rather than the atomate2 ones (which have different settings e.g., for static calculations).

src/atomate2/vasp/sets/core.py
```diff
@@ -543,6 +547,135 @@ def incar_updates(self) -> dict:
...
+                    # In pymatgen/io/vasp/MVLGWSet.yaml,
+                    # the default value of NELM is 100;
+                    # but in the pymatgen.io.vasp.sets.MVLGWSet,
+                    # the default value of NELM for DIAG calculation
+                    # is manually set to be 1.
+                    # If 1 were used, the NSCF calculation would not converge,
+                    # and Custodian would not proceed with later calculations.
+                    # Here, I decide to keep the default value of NELM
+                    # to be 100 as defined in pymatgen.io.vasp.sets MVLGWSet.yaml.
+                    "NELM": 100,
```
utf, line 639: FYI NELM should be 1 as per the VASP manual - exact diagonalisation already converges after 1 step

src/atomate2/vasp/sets/core.py
```diff
@@ -543,6 +547,135 @@ def incar_updates(self) -> dict:
         return updates
 
 
+@dataclass
+class MVLGWSetGenerator(VaspInputGenerator):
```
utf, line 551: In the latest version of atomate2, we have moved to using the input sets defined in pymatgen. Can you remove this definition and import the pymatgen versions instead (hopefully they should just be a drop in replacement). If you found any bugs etc in the MVL input sets then can you submit a PR to pymatgen.

src/atomate2/vasp/jobs/core.py
```diff
@@ -490,6 +491,171 @@ def make(
         return super().make.original(self, structure, prev_dir)
 
 
+@dataclass
+class MVLStaticMaker(BaseVaspMaker):
```
utf, line 495: Can you also move these input sets to `jobs/mvl.py` or `jobs/gw.py`

utf (commented):
Thanks for the PR. I've put some comments for you.

One thing to note is that in VASP v6.3 there is the option to do a single step GW calculation (i.e., no need for separate static+gw) which simplifies everything. E.g., see the manual here: https://www.vasp.at/wiki/index.php/Practical_guide_to_GW_calculations

As the MVL workflow is quite old, it could be nice to support this newer approach (potentially in a future PR).

src/atomate2/vasp/sets/core.py
yanghan234: Thanks for the information. I understand. However, this is due to the fact that custodian will mistakenly treat the results of NEML = 1 as an unverged calculation. Thus, currently I set NELM = 100 and enforced ICHARG=11 (constant charge) to make it converge, so that I can pass custodian checks and proceed to G0W0 calculation.

yanghan234: By the way, I also notice that `NonSCFMaker` in `atomate2.vasp.jobs.core` has the same issue. Although it finishes gracefully, custodian will still complain about being not converged.

src/atomate2/vasp/sets/mvl.py
```diff
@@ -0,0 +1,56 @@
...
+            A dictionary of updates to apply.
+        """
+        updates = super().incar_updates
+        if self.mode == "DIAG":
+            # TODO: Update the values later
+            # Currently, custodian will raise VaspErrorHandler
+            # because it mistakenly thinks the calculation is not converged if NELM = 1
+            # So, I make NELM = 100 here but keep the density
+            # to be constant (ICHARG = 11).
+            updates["NELM"] = 100
```
utf, line 52: I don't think this is an ideal solution. I think we should submit a PR to custodian that doesn't apply the error handler if NELM = 1 and ALGO = Exact. That way we can just use the input set directly from pymatgen and we can also follow the correct VASP procedure.

utf (commented):
Thanks @yanghan234, this is almost there now. Please see the final comment (I believe we will need a PR to custodian to keep the calculation procedure optimal).

src/atomate2/vasp/sets/mvl.py
yanghan234: It turns out this is actually due to the issue in the `Vasprun` class in pymatgen. It mistakenly tells `custodian` that the non-self-consistent calculation is not converged. I opened a PR in which I modified two lines to jump from this condition, you may review the change here: [Fix: fixed electronic step check with algo = exact and nelm = 1](https://github.com/materialsproject/pymatgen/pull/4071)

yanghan234:
I have updated the codes and testcases accordingly. Now, the NELM issue is solved. Once the PRs in emmet and pymatgen repos are merged, this PR should be good to go. Thanks for the comments. @utf

utf:
Awesome, thanks @yanghan234.

src/atomate2/vasp/sets/core.py
computron: This should be fixed by updating the handlers for custodian

computron: (nevermind, just saw this was handled otherwise)

utf:
Hi @yanghan234 are you ok to update the emmet/pymatgen versions in `pyproject.toml`?

yanghan234:
Thanks @utf, I have updated the version of emmet-core to the latest release candidate [0.84.3rc3](https://github.com/materialsproject/emmet/releases/tag/0.84.3rc3). The pymatgen version has already been updated, so I don't need to change.

yanghan234:
Thanks @utf. I updated the branch again, and made some modifications as you suggested. Please test it again and see if it is ready. Thank you!

yanghan234:
Hi @utf, when you have a moment, could you kindly review the latest changes? I believe we’re close to finalizing everything, and I’d appreciate your insights before merging. Thank you very much!

yanghan234:
Hi @utf, could you take a moment to review this PR? It’s nearly finalized. Thank you!

utf:
Hi @yanghan234, appologies for the delay.

This looks good to go.

PR #723, opened 2024-02-16
Update LobsterTaskDoc 

naik-aakash:
# Summary

The changes in the are first step toward LOBSTER dataset available via MP

# Todo

- [x] decouple file pymatgen file parsing and post-processing (lobsterpy)
- [x] Set `save_computational_data_jsons `and `add_coxxcar_to_task_document ` default to false 
- [x] Adapt tests to accommodate the update in defaults
- [x] Move`Bandoverlaps ,Charge, Grosspop, Icohplist, MadelungEnergies, SitePotential` fields to data (Objects could be larger than 16MB for large structure)
- [x] Option to manipulate lobsterpy kwargs
- [x] redefine fields to have minimum redundant information
- [x] Move `Charge`, `MadelungEnergies `back to mongostore (size we suspect will not be larger than 16 MB , as these are simple text files with minimal data in them)
- [x] Nest strongest_bonds fields

PR #811, opened 2024-04-19
Document the option to run LOBSTER with one jobscript only

JaGeo:
@utf, I have recently made some modifications to custodian so that we can run VASP and LOBSTER now with a minimal setup on 1 node. Users can use this new option once the new custodian version is required and https://github.com/materialsproject/atomate2/pull/723 is merged. 

I would love to advertise this also in the new atomate2 paper, as this now enables us to use the full capabilities of jobflow with the LOBSTER workflow as well. Of course, specialists might want to run VASP on more than one node. However, in most use cases, this version should be sufficient. And, we might attract a few new atomate2 users in this way.


JaGeo:
This is only additional documentation. This is why I will merge.

PR #830, opened 2024-05-03
Add missing `calculator_kwargs` and remove outdated `model`/`model_kwargs` in `ForceFieldRelaxMaker` doc strings

### LINKED ISSUE #829: Feature: Easy switch between GPU/CPU for forcefields
For workstations, switching between GPUs and CPUs should be possible for the force field workflows. Do we currently have an easy option within the workflow to do so? 

We typically run these jobs on either GPUs or CPUs and rarely have an issue. Did anyone of you run into this? @esoteric-ephemera @janosh ? Especially in the context of CHGNet?

janosh:
closes #829


janosh:
noting in case it comes up again that the [`codedov` action seems to have gotten stuck](https://github.com/materialsproject/atomate2/actions/runs/8941618714/job/24562362585)

Screenshot 2024-05-03 at 12 41 09

PR #850, opened 2024-05-14
[WIP] Documentation update

AntObi:
## Summary

In the API reference of the current documentation for atomate2, not all of the subpackages and modules are displayed, which means that a new user to atomate2 will not be able to see all the new workflows added (e.g. forcefields, CP2K etc).
image

* Fix 1: Add new subpackages to index.rst to update the documentation

## TODO (if any)

image
* Updating index.rst adds the subpackages to the API reference, but not all subpackages have been automatically documented correctly (see image above). I need to investigate the cause of this.
* Need to document that when developers create a new subpackage, they should also add it to the index.rst file and also check that the documentation renders properly locally.

## Checklist

title.

  type check your code.


src/atomate2/cp2k/schemas/calc_types/enums.py
```diff
@@ -1,14 +1,13 @@
...
+Autogenerated Enums for CP2K RunType, TaskType, and CalcType
+Do not edit this by hand. Edit generate.py or run_types.yaml instead
 """
 
 from emmet.core.utils import ValueEnum
 
 
 class RunType(ValueEnum):
-    """CP2K calculation run types."""
+    """CP2K calculation run types"""
```
utf, line 10: I think these full stops should actually be there?

utf:
Thanks very much @AntObi for the detailed PR. I'm happy to merge once the CI passes. I think the main issue is the punctuation that was removed.

To make things easier, you can install the pre-commit which will catch most things before committing.

```bash
pip install pre-commit
pre-commit install
```

src/atomate2/cp2k/schemas/calc_types/enums.py
AntObi: I didn't manually edit those, so they were changed automatically when I built the docs using sphinx. Will fix those!

utf: Ah, I remember this. Yes, sphinx causes them to be regenerated. I think this problem could be fixed in the future by adding the correct punctuation in the script that auto generates this file.

AntObi:
@utf Thanks for the fast response! You can hold off on merging this PR until I fix the issue with the `forcefields` subpackage not showing all of the available modules!

JaGeo:
Great work, @AntObi ! Wondering if we should also update the readme?

src/atomate2/cp2k/schemas/calc_types/enums.py
AntObi: I've now fixed the script which autogenerates that file, so building the docs and running pre-commit should leave that file unaffected,

AntObi:
Hi @utf @JaGeo , I think this PR will take longer than I expected 😅 .
 
> ## TODO (if any)
> image
> 
> * Updating index.rst adds the subpackages to the API reference, but not all subpackages have been automatically documented correctly (see image above). I need to investigate the cause of this.
> * Need to document that when developers create a new subpackage, they should also add it to the index.rst file and also check that the documentation renders properly locally.

As I mentioned in the TODO for this PR, some of the subpackages/modules don't have their classes and functions appear in the docs. So far, I haven't been identify the cause of this (likely something I have to dig into Sphinx to find out).

@JaGeo What did you have in mind for updates to the README?

utf:
Hi @AntObi. All you should need to do is add the top level links. Sphinx-autodoc should do the rest. Can you try removing the build folder and rebuilding from scratch. Does that help at all?

JaGeo:
> @JaGeo What did you have in mind for updates to the README?

Similar to the publication draft, we list all available calculators and highlight the most important workflows.

utf:
Thanks @AntObi

Issue #817, opened 2024-04-24
Improve failed perturbations handling in elastic workflow


gpetretto:
I have seen a few cases where some pertubations of the elastic workflow fail and the elastic tensor is still calculated since `on_missing_references` is set to `NONE`. Due to the fact that `sym_reduce` is True by default, the number of perturbations is already relatively small for highly symmetric systems. So, the result, although still somewhat reasonable, was wrong.
I think there are a few points that could be addressed to improve the current situation:
* there should probably be an option to disable this behaviour, since it may lead to wrong results. Especially if symmetry is used to reduce the number of perturbations.
* maybe a check should be performed in the `fit_elastic_tensor` and raise an error if too many perturbations have failed or, in a more advanced fashion, if all the perturbations concering one kind of strain failed.
* If I am not mistaken, in case of failure of some of the perturbations, there is no obvious way of knowing that it happened just from the output database. It may be good to have a warning flag in the document or a list of the deformations that were supposed to be in the list of perturbations, but were not calculated. While this information could be retrieved from fireworks/jobflow_remote, it seems quite inconvenient and is not a given that the information will still be easily available at a later time. 

What do you think? Did anybody else have these kind of issues?

JaGeo:
Even though, I have build a force-field based elastic maker, I haven't tested it yet in so much detail but I think we should make it as safe as possible for users. Warnings and additional information on the fit should be available to the user.

PR #861, opened 2024-05-28
Handling failures in elastic flow

gpetretto:
Following what already mentioned in #817, I have implemented a few changes in the BaseElasticFlow.

* It is now possible to specify a `max_failed_deformations` in the Flow, that will not complete the fit job if the number of failed deformation jobs exceeds that number. The default is no limit for backward compatibility.
* The uuids of the failed perturbations are included in the `ElasticDocument`.
* A list of warnings has also been added to the `ElasticDocument`. If the list of failed uuids is not empty, a warning is added there.


src/atomate2/common/jobs/elastic.py
```diff
@@ -200,21 +201,42 @@ def fit_elastic_tensor(
...
 
         stresses.append(Stress(data["stress"]))
         deformations.append(Deformation(data["deformation"]))
         uuids.append(data["uuid"])
         job_dirs.append(data["job_dir"])
 
+    if max_failed_deformations is not None:
+        if 0 < max_failed_deformations < 1:
+            fraction_failed = len(failed_uuids) / len(deformation_data)
+            if fraction_failed > max_failed_deformations:
```
JaGeo, line 229: This is a comparison of a fraction with something that looks like an int, right?

src/atomate2/forcefields/flows/elastic.py
```diff
@@ -45,6 +45,11 @@ class ElasticMaker(BaseElasticMaker):
     bulk_relax_maker : .ForceFieldRelaxMaker or None
         A maker to perform a tight relaxation on the bulk. Set to ``None`` to skip the
         bulk relaxation.
+    max_failed_deformations: int or float
```
JaGeo, line 48: Doesn't show up below in the list of default parameters

src/atomate2/vasp/flows/elastic.py
```diff
@@ -49,6 +49,11 @@ class ElasticMaker(BaseElasticMaker):
         bulk relaxation.
     elastic_relax_maker : .BaseVaspMaker
         Maker used to generate elastic relaxations.
+    max_failed_deformations: int or float
```
JaGeo, line 52: Same here

tests/common/jobs/test_elastic.py
```diff
@@ -54,3 +57,137 @@ def _get_strains(structure, sym_reduce):
     deformations = response[job.uuid][1].output
 
     return [d.green_lagrange_strain for d in deformations]
+
```
JaGeo, line 60: Might be good to add a test for the whole flow or at least nake sure that the parameter won't be removed.

JaGeo:
@gpetretto Thank you! I left a comment on the overall flow. I would expect that we can also really change this parameter at the flow level. I assume nearly no one will use the fitting part alone.

src/atomate2/forcefields/flows/elastic.py
JaGeo: (Just to add: Typically, all parameters are repeated from the BaseMaker.)

gpetretto: I can add it to the subclasses. But why redefining them, if the default value and definition are the same? I was actually confused by the fact that all the attributes are overwritten, while it is not strictly necessary.
This does not seem to be the case for all the Makers. For example the subclasses of `BaseVaspMaker` do not redefine the parameters of the base class.

JaGeo: It was intitially intended to make the inheritance more transparent to users that directly read the code.

JaGeo:
My initial review comment was unclear (sorry about this, i should not write on my phone): what I meant to say was that I would also recommend a test at the flow level. Of course, a part of the functionality is also already tested with the default values but not everything.

tests/common/jobs/test_elastic.py
gpetretto: Adding a test for the whole workflow is a bit problematic because it is difficult to consistently trigger the failure of one of the pertrubations. I can manually generate a test by explicitly killing vasp for one of the perturbations, but if the tests will need to be regenerated in the future it may be annoying.
It would be easier if I could modify the INCAR of only one of the perturbations, but it seems to not be possible because `update_kwargs` is not applied to dynamically generated jobs (see https://github.com/materialsproject/jobflow/issues/588).
I thought these tests will be enough, since all the logic stays in the fit job, but if you think it is needed I can try to generate the workflow test by killing the job. Or maybe there is a better way?

JaGeo: Ah, I see. This sounds like too much work.

I think the fitting part is well tested. I was rather thinking about testing that a parameter change really arrives in the fit function

JaGeo: If it was part of the output schema, it would be easy. 😅

src/atomate2/forcefields/flows/elastic.py
gpetretto: I added it. I thought that it does not make sense to have some parameters overridden and some not, in this specific flow. We can remove it at a later stage if really needed.

tests/common/jobs/test_elastic.py
JaGeo: I think this is the only missing point. As this is a rather small but still important pull request, I would in this case merge it.
Does it make sense to add the max_* to the output schema and just test if it arrives there if you set it in the wf?

gpetretto: Sorry, I think I had partially misunderstood your previous comment. I could add `max_failed_deformations` to the output document, but I am not sure if it is really worth to have that information just to make the test. I will try if I can quickly make a test by explicitly killing one of the deformation jobs, if it gets too complicated I will add `max_failed_deformations`.

gpetretto: I managed to add a test with the flow failure triggered by the `max_failed_deformations` being set to 0.

JaGeo: Awesome. Thank you!

JaGeo:
@gpetretto Thank you! I will merge.


@utf I am merging this now as this is a rather small change!

PR #872, opened 2024-06-04
Bug in supercell matrix determination in phonon workflow

JaGeo:
## Summary

Currently, the supercells that are generated with the phonon workflow are sometimes not as cubic as possible (e.g., for diamond-type Si). This is due to a bug in the supercell matrix handling. The supercell matrix as determined by pymatgen needs to be transposed before it is inserted into phonopy. 

- [ ] add tests


utf:
Good catch. Thanks @JaGeo

JaGeo:
I will check this in more detail. I actually thought I had checked this before and I am a bit sad that it did not work as intended so far ....

PR #892, opened 2024-06-19
Fix the path of `.pmgrc.yaml` used in install.md.

hongyi-zhao:
The path should be `~/.pmgrc.yaml` instead of `~/.config/.pmgrc.yaml`.

### DIFF
--- a/docs/user/install.md
+++ b/docs/user/install.md
@@ -329,7 +329,7 @@ and then return to this tutorial.
 
 You can get an API key from the [Materials Project] by logging in and going to your
 [Dashboard](materials project). Add this also to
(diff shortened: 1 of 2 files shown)


janosh:
thanks for taking the time to submit a PR. 👍 

have a look at this [line in pymatgen](https://github.com/kaueltzen/pymatgen/blob/81f76f4e0df361a47b56533e858d5353b17075df/pymatgen/core/__init__.py#L31) though. `.pmgrc.yaml` is actually preferentially read from `~/.config/`.

hongyi-zhao:
See https://github.com/materialsproject/pymatgen/pull/3891 for my improvement on loading and expanding YAML variables.

Issue #899, opened 2024-06-24
BUG: cannot resolve CP2K reference


lory-w:
**Describe the bug**
When running the CP2K workflow with RelaxBandStructureMaker, a ValidationError occurs in the TaskDocument schema due to improper handling of the `cp2k_input` and `locpot` fields. The error message is as follows:

> Traceback (most recent call last):
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/jobflow/managers/local.py", line 114, in _run_job
    response = job.run(store=store)
               ^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/jobflow/core/job.py", line 572, in run
    self.resolve_args(store=store)
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/jobflow/core/job.py", line 678, in resolve_args
    resolved_args = find_and_resolve_references(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/jobflow/core/reference.py", line 473, in find_and_resolve_references
    resolved_references = resolve_references(
                          ^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/jobflow/core/reference.py", line 361, in resolve_references
    resolved_references[ref] = ref.resolve(
                               ^^^^^^^^^^^^
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/jobflow/core/reference.py", line 179, in resolve
    data = MontyDecoder().process_decoded(data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/monty/json.py", line 777, in process_decoded
    return cls_(**d)
           ^^^^^^^^^
  File "/opt/conda/envs/ai-dopant/lib/python3.12/site-packages/pydantic/main.py", line 176, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 2 validation errors for TaskDocument
calcs_reversed.0.input.cp2k_input
  Input should be a valid dictionary [type=dict_type, input_value=<pymatgen.io.cp2k.inputs....bject at 0x7f4e4eb21640>, input_type=Cp2kInput]
[...]

PR #900, opened 2024-06-24
fix CP2K TaskDocument

lory-w:
## Summary

This PR addresses the ValidationError occurring in the TaskDocument schema due to improper handling of the cp2k_input and locpot fields. This issue is documented in issue #899 

* Changed `cp2k_input` in CalculationInput to be a union of `CP2KInput` object and dict to accommodate different possible types for cp2k_input
* Changed `locpot` in CalculationOutput to be an optional field

### DIFF
--- a/src/atomate2/cp2k/schemas/calc_types/utils.py
+++ b/src/atomate2/cp2k/schemas/calc_types/utils.py
@@ -71,7 +71,7 @@ def task_type(inputs: dict) -> TaskType:
     """
     calc_type = []
     cp2k_run_type = inputs.get("cp2k_global", {}).get("Run_type", "")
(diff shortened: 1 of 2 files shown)


JaGeo:
@lory-w thank you! Is there a chance that you could add a test as well?

lory-w:
Hi, 
Thank you for reviewing the PR. I attempted to write a pytest but the error only occurs when the calculation results are queried by Jobflow from Store (in my case mongodb). However, the current test suite in the repo does not seem to have integration with mongodb, so I cannot reproduce the bug within the unit tests.

Would you suggest any specific approach to handle this? Thanks!

JaGeo:
@utf maybe you could have a look once you have time!

utf:
I will look into the jobflow issue but will merge this for the time being.

PR #901, opened 2024-06-25
Anharmonicity Quantification workflow

4kevinbeck5:
## Summary

Creates a workflow for anharmonicity quantification as defined in  [Knoop 2020](https://journals.aps.org/prmaterials/abstract/10.1103/PhysRevMaterials.4.083809)

* Workflow and schema for anharmonicity quantification
* Implementation in FHI-aims

* None

## TODO (if any)

Larger scale testing in progress

## Checklist

title.

  type check your code.


JaGeo:
@4kevinbeck5 oh, that's cool. Would love to make it available for Vasp once we have some time

tpurcell90:
ready for review

utf:
Hi @JaGeo would you be happy to lead the review on this WF since it is phonon related?

JaGeo:
@utf Yes, I will try to take a look latest next week!

src/atomate2/aims/flows/anharmonicity.py
```diff
@@ -0,0 +1,44 @@
+"""Defines the phonon workflows for FHI-aims."""
```
JaGeo, line 1:
```suggestion
"""Defines the anharmonicity quantification workflows for FHI-aims."""
```

src/atomate2/common/flows/anharmonicity.py
```diff
@@ -0,0 +1,270 @@
+"""A workflow to evaluate the anharmonicity of a material with sigma^A.
+
+For details see: doi.org/10.1103/PhysRevMaterials.4.083809
+"""
```
JaGeo, line 4: Could be similar in the fhi aims part.

src/atomate2/common/jobs/anharmonicity.py
```diff
@@ -0,0 +1,611 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Constants
+EV = 1.60217733e-19  # [J]
+AA = 1e-10  # [m]
+AMU = 1.66053904e-27  # [kg]
+THZ = 1e12  # [1/s]
+PI = np.pi
+OMEGA_TO_THZ = (EV / AA**2 / AMU) ** 0.5 / THZ / 2 / PI  # 15.633302 THz
```
JaGeo, line 37: can't it be imported from pymatgen or elsewhere? Especially eV to J or amu to kg?

src/atomate2/common/jobs/anharmonicity.py
```diff
@@ -0,0 +1,611 @@
...
+    -------
+    Structure
+        The phonopy structure
+    """
+    cell = get_phonopy_structure(phonon_doc.structure)
+    phonon = Phonopy(
+        cell,
+        phonon_doc.supercell_matrix,
+    )
+    return get_pmg_structure(phonon.supercell)
```
JaGeo, line 59: could be implemented without the doc and only supercell matrix and structure but maybe not necessary to change it.

src/atomate2/common/jobs/anharmonicity.py
```diff
@@ -0,0 +1,611 @@
...
+    if len(np.shape(forces_dft)) == 2:
+        forces_dft = np.expand_dims(forces_dft, axis=0)
+        forces_harmonic = np.expand_dims(forces_harmonic, axis=0)
+
+    atom_numbers = np.array([site.specie.number for site in structure.sites])
+    if np.shape(forces_dft)[1] == 3 * structure.num_sites:
+        atom_numbers = atom_numbers.repeat(3)
+
+    symbols = np.array([site.specie.name for site in structure.sites])
+    unique_atoms = np.unique(atom_numbers)
```
JaGeo, line 98: It's more about elemental anharmonicities or am I wrong? Might be good to emphasize this. I would have expected an anharmonicty score for each unique site in the structure.

src/atomate2/common/jobs/anharmonicity.py
```diff
@@ -0,0 +1,611 @@
...
+    eig_val, eig_vec = np.linalg.eigh(dynamical_matrix)
+    eig_val = np.sqrt(eig_val[3:])
+    x_acs = eig_vec[:, 3:].reshape((-1, 3, len(eig_val)))
+
+    # gauge eigenvectors: largest value always positive
+    for ii in range(x_acs.shape[-1]):
+        vec = x_acs[:, :, ii]
+        max_arg = np.argmax(abs(vec))
+        x_acs[:, :, ii] *= np.sign(vec.flat[max_arg])
+
```
JaGeo, line 219: does it make sense to have this without the workflow as a function somewhere? (not necessary but maybe useful for reuse in the future)

tests/aims/test_flows/test_anharmonic_quantification.py
```diff
@@ -0,0 +1,329 @@
...
+
+from atomate2.aims.flows.anharmonicity import AnharmonicityMaker
+from atomate2.aims.flows.phonons import PhononMaker
+from atomate2.aims.jobs.core import RelaxMaker, StaticMaker
+from atomate2.aims.jobs.phonons import (
+    PhononDisplacementMaker,
+    PhononDisplacementMakerSocket,
+)
+
+cwd = os.getcwd()
```
JaGeo, line 16: isn't there another option to get the current directory in test environment?

JaGeo (commented):
I have added some smaller comments for now. Will think more carefully about it next week as well.

src/atomate2/common/jobs/anharmonicity.py
JaGeo: https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/core/units.py (see here)

tpurcell90: I suggested this initially because I was not sure which CODATA was used for different codes. The impact should be minor, but I know VASP and aims have slightly different definitions of the electronvolt because of it.

src/atomate2/common/jobs/anharmonicity.py
tpurcell90: Potentially it could be useful, so we should probably move it into a separate function

tests/aims/test_flows/test_anharmonic_quantification.py
tpurcell90: There might be, I think this was probably copied from some other FHI-aims tests and I don't know if Andrei and I thought to check when we first wrote it

src/atomate2/common/jobs/anharmonicity.py
4kevinbeck5:
```
from pymatgen.core.units import Energy, Mass, Length, FloatWithUnit
import numpy as np 

inner = Energy(1, "eV") / (Length(1, "ang"))**2 / Mass(1, "amu")
inner_sqrt = np.sqrt(inner.unit.as_base_units[1])
omega_to_THz = inner_sqrt / FloatWithUnit(1, "THz").to("Hz") / 2 / np.pi
```
I did some messing around with the pymatgen units and found a way that works and gets to the same OMEGA_TO_THZ conversion factor as is currently in the code

JaGeo: Ah, i see. If this is required for aims, this is absolutely fine, of course.

JaGeo: Or maybe you define the aims units once and then import them? I think there have been unit conversions elsewhere in the code, right?

JaGeo: Ah, here they are: https://github.com/materialsproject/atomate2/blob/main/src/atomate2/aims/utils/units.py

I think you might want to reuse these definitions

tests/aims/test_flows/test_anharmonic_quantification.py
JaGeo: I thin the `clean_dir` fixture takes care of this: https://github.com/materialsproject/atomate2/blob/a9e5edecc267760c84562efe8101acc7390b61cc/tests/conftest.py#L41
I think everywhere else in the code, this is used for the workflow tests

src/atomate2/common/jobs/anharmonicity.py
tpurcell90: You are correct these are elemental values, in principle we could also add site values as well.

src/atomate2/common/jobs/anharmonicity.py
tpurcell90: Ya we should use these. I forgot to check that.

src/atomate2/common/jobs/anharmonicity.py
JaGeo: I would recommend making this clearer in the documentation/naming of the methods.

I could imagine that such site-wise values should be very insightful for typical structures with rattling atoms.

src/atomate2/common/jobs/anharmonicity.py
4kevinbeck5: I just changed the code to import the conversion factor from atomate2 rather than re-adding it to the anharmonicity workflow

src/atomate2/common/jobs/anharmonicity.py
4kevinbeck5: There is already a separate function build_dynmat on line 241. One possibility could be to move lines 210-211 into this function and have it return the eigenvals and eigenvecs. Then this part could just call the updated build_dynmat and gauge the eigenvectors. In the step for mode-resolving sigma^A, build_dynmat is also called but we don't gauge the eigenvectors so I'm not sure if it would be better to have the gauging step separate from the new function

tpurcell90: I would have the dynamic matrix and the eigenvalue/vector solver because some applications may just need the matrix

4kevinbeck5: I separated out the code so that in the jobs/anharmonicity.py file there is one helper function to calculate the dynamical matrix and then a second to diagonalize the dynamical matrix and square root the eigenvalue so that we have the eigenmodes and eigenfrequencies.

JaGeo:
Tests are failing here. I think you have to fix the torchdata version similar to #940 .

JaGeo:
Let me know if I should do another review round. Currently, there are some conflicts

4kevinbeck5:
@JaGeo I just pushed a version that implemented the mode resolved sigma^A and fixed all the things you brought up previously

src/atomate2/common/jobs/anharmonicity.py
```diff
@@ -0,0 +1,733 @@
...
+                    list(site_to_wyckoff.keys())[i]
+                ]
+            },
+            sigma_sites[i],
+        )
+        for i in range(len(sigma_sites))
+    ]
+
+
+def get_sigma_per_element(
```
JaGeo, line 149: @4kevinbeck5 this looks very good. 

The only comment that I have is that some of these are functions and are not really jobs. @utf should we move those somewhere else? `common/utils` would be a possibility but it might get messy.

4kevinbeck5: The rationale for making those functions rather than jobs was just to reduce the number of jobs per flow on the HPC

JaGeo: This is completely fine and desired. I am just wondering if functions should be in the `jobs` folder/file.

4kevinbeck5:
I just committed a version that passes the linting this time. There was a small mistake I made before in the conftest.py file

JaGeo:
@4kevinbeck5 I will check! Thanks!

Could you also add yourself to the list of contributors 😃?

4kevinbeck5:
@JaGeo I just pushed the contributors.md file with my info added!

src/atomate2/common/flows/anharmonicity.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # Get DFT and harmonic forces
+        force_calcs = get_forces(
+            phonon_doc.force_constants,
+            phonon_supercell,
+            displacement_calcs.output,
+        )
+        jobs.append(force_calcs)
+
+        sigma_a_vals = {}
```
JaGeo, line 255: This can be removed.

JaGeo:
@4kevinbeck5 I removed one dict initialization and checked if the torchdata pin is still needed. I am now happy with the changes. I am enabling auto merge! 

If I have some time, I might implement a VASP interface. It might, however, be that this will be next year ;).

PR #902, opened 2024-06-28
[WIP] Workflow for  Quasi-harmonic approximation (forcefields and VASP)

JaGeo:
## Summary

ToDo:

- [x] Reduce number of steps in the workflow
- [x] Finalize Force field tests
- [ ] Check all units and add field descriptions
- [ ] Document the workflows
- [ ] Fix all linting issues
- [ ] write workflows for the different force fields
- [ ] add a VASP Workflow (with minimal settings)

PR #889, opened 2024-06-18
Equation of State workflow for FHI-aims

ansobolev:
## Summary

 * This PR adapts the common Equation of State workflow to FHI-aims. 

## Additional dependencies introduced (if any)

None

## TODO (if any)

None

## Checklist

Before a pull request can be merged, the following items must be checked:

* [x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  The easiest way to handle this is to run the following in the **correct sequence** on
  your local machine. Start with running [`ruff`](https://docs.astral.sh/ruff) and `ruff format` on your new code. This will
  automatically reformat your code to PEP8 conventions and fix many linting issues.
* [x] Doc strings have been added in the [Numpy docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.
* [x] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org) to
  type check your code.
* [x] Tests have been added for any new functionality or bug fixes.
* [x] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more
efficient if you already fix most errors prior to submitting the PR. It is highly
recommended that you use the pre-commit hook provided in the repository. Simply run
`pre-commit install` and a check will be run prior to allowing commits.

PR #903, opened 2024-06-28
Workflow for Quasi-harmonic approximation (forcefields and VASP)

JaGeo:
Continues #902 .

ToDo:

- [x] Reduce number of steps in the workflow
- [x] Finalize Force field tests
- [x] Check all units and add field descriptions
- [x] Extend the tests
- [x] Fix all linting issues
- [x] add options for different eos fits
- [x] add a VASP Workflow (with minimal settings)
- [x] Job names are updated
- [x] Check execution order (resolved)
- [x] Add tests for VASP workflow (due to issues with the naming convention, I reduced the VASP test workflow to 1 optimization, 1 eos structures, 1 phonon run and skipped the analysis step)
- [x] check born_maker initialisation
- [x] Add large amounts of data to data store
- [x] Figure out good vasp settings for the optimization and how to deal with prev_dir
- [x] Document the new workflows


JaGeo:
I tracked it down. As soon as I initialize `PhonopyQHA`, I cannot jsanitanize the document. Even in cases when none of the outputs are used for the document. This is an extremely weird error and I might have to go through the phonopy source to understand where this might be coming from.

JaGeo:
I think it is this line here: https://github.com/phonopy/phonopy/blob/04f912a77e3efbac00a9987b545fc4f5e8927eb9/phonopy/qha/eos.py#L134

tests/common/schemas/test_qha.py
```diff
@@ -0,0 +1,247 @@
...
+        76.29,
+    ]
+    assert_allclose(qha_doc.temperatures[5], 10.0)
+    assert_allclose(qha_doc.heat_capacities[10][9], 2.5144627)
+
+    assert_allclose(qha_doc.entropies[10][2], 0.1449301)
+    assert_allclose(qha_doc.bulk_modulus_temperature[30], 75.0604563982221, atol=1e-8)
+    assert_allclose(qha_doc.bulk_modulus, 0.48559238394681514, atol=1e-8)
+    assert_allclose(qha_doc.gibbs_temperature[10], -14.814345401430508, atol=1e-8)
+    assert_allclose(qha_doc.gruneisen_temperature[5], 3.001253318775402, atol=1e-3)
```
JaGeo, line 242:
```suggestion
    assert_allclose(qha_doc.gruneisen_temperature[5], 3.001253318775402, atol=1e-2)
```

src/atomate2/common/schemas/qha.py
```diff
@@ -0,0 +1,235 @@
...
+                t_max=t_max,
+                pressure=pressure,
+                eos=eos_type,
+            )
+
+        # create some plots here
+        # add kwargs to change the names and file types
+        qha.plot_helmholtz_volume().savefig(
+            f"{kwargs.get('helmholtz_volume_filename','helmholtz_volume')}"
+            f".{kwargs.get('plot_type','eps')}"
```
QuantumChemist, line 151: I would use pdf 🤔

JaGeo:
@utf this would be ready for a review (when there is some time). `prev_dir` is not perfectly handled yet. Happy to hear your thoughts on this.

JaGeo:
Thermal expansion from DFT (PBEsol) with 4 frames an 15 Angstrom unit cells (fit only in h +-5% region):

image

(I still need to set better defaults)

JaGeo:
@tpurcell90 this one would be nearly ready. If you have time, you could have a look here.

tpurcell90:
Yes do you want me to do directly in this PR or in a separate branch that gets merged in (not sure if there are concurrent changes going on)?

JaGeo:
@tpurcell90 i have stopped working on it for now. Tests have been passing and I am waiting for further feedback from @utf . 

But you might want to use your own branch starting from this one.

tpurcell90:
Okay I will start my own branch and then merge things in once done just in case

tpurcell90:
It might make more sense for #889 to be fully merged in before I adapt this one. Since that one seems to just need a final merge. Since this also uses eos workflows I don't want to conflict with what @ansobolev did

JaGeo:
Ah, I see! @tpurcell90

JaGeo:
@tpurcell90 to add, i don't rely on the full eos workflow as it would be too hard to modify. Therefore, I would not expect many conflicts.

src/atomate2/vasp/flows/qha.py
```diff
@@ -0,0 +1,120 @@
...
+    name: str = "VASP QHA Maker"
+    initial_relax_maker: TightRelaxMaker | None = field(default_factory=TightRelaxMaker)
+    eos_relax_maker: TightRelaxMaker | None = field(
+        default_factory=lambda: EosRelaxMaker(
+            input_set_generator=TightRelaxSetGenerator(
+                user_incar_settings={"ISIF": 2},
+            )
+        )
+    )
+    phonon_displacement_maker: StaticMaker | None = field(default_factory=StaticMaker)
```
JaGeo, line 65: Fix this here

src/atomate2/forcefields/flows/qha.py
```diff
@@ -0,0 +1,127 @@
...
+        bulk_relax_maker: .ForceFieldRelaxMaker|None
+            Relax Maker for Phonon Run. Typically None.
+        phonon_maker_kwargs: dict
+            Dict to set additional info for phonons.
+
+        Returns
+        -------
+        .PhononMaker
+        """
+        logging.log(level=0, msg="test")
```
JaGeo, line 109: remove

src/atomate2/vasp/flows/qha.py
JaGeo: Changed

src/atomate2/common/flows/qha.py
```diff
@@ -0,0 +1,189 @@
...
+        qha_jobs = []
+
+        # In this way, one can easily exchange makers and enforce postprocessor None
+        self.eos = CommonEosMaker(
+            initial_relax_maker=self.initial_relax_maker,
+            eos_relax_maker=self.eos_relax_maker,
+            static_maker=None,
+            postprocessor=None,
+            number_of_frames=self.number_of_frames,
+        )
```
JaGeo, line 121: @utf I am currently wondering about how to initalize and reuse the EosMaker in the best way. Unfortunately, starting  solely from the EosMaker for the qha workflow would take too much time.

JaGeo: @utf ? any opinions?

utf: I'm a little confused - what do you mean by:
> starting solely from the EosMaker for the qha workflow would take too much time

Either way, I'm open to both approaches:
- Adding the EosMaker as an argument would reduce the number of options in the `CommonQhaMaker` class but make it more difficult to set properties such as the number of frames. I don't think this is too much of an issue if we could include an example of how to configure the settings in the docstring/documentation page.
- The current internal use of CommonEosMaker makes it easier to configure settings but is slightly less clean code. By the way, I noticed the `linear_strain` option is not currently used.

I'll let you decide!

src/atomate2/vasp/flows/qha.py
```diff
@@ -0,0 +1,87 @@
...
+    def prev_calc_dir_argname(self) -> str:
+        """Name of argument informing static maker of previous calculation directory.
+
+        As this differs between different DFT codes (e.g., VASP, CP2K), it
+        has been left as a property to be implemented by the inheriting class.
+
+        Note: this is only applicable if a relax_maker is specified; i.e., two
+        calculations are performed for each ordering (relax -> static)
+        """
+        return "prev_dir"
```
utf, line 87: Is this actually needed any more? VASP, CP2K etc all have the same dir name now `prev_dir`. I guess we coud remove this from any other workflows that have it?

src/atomate2/common/jobs/qha.py
```diff
@@ -0,0 +1,126 @@
...
+    phonon_maker: .BasePhononMaker
+        Maker to start harmonic phonon runs.
+    eos_output: dict
+        Output from EOSMaker
+
+    """
+    phonon_jobs = []
+    outputs = []
+    for istructure, structure in enumerate(eos_output["relax"]["structure"]):
+        phonon_job = phonon_maker.make(structure)
```
utf, line 39: If possible, it would be nice to pass the prev_dir from the relaxation? That way we can make use of `auto_ispin` and make the calculations cheaper.

utf (commented):
Thanks for this @JaGeo, it looks in a very nice state. Just some minor points from me.

src/atomate2/common/jobs/qha.py
JaGeo: Did this now by extending the EOS dictionaries.

src/atomate2/vasp/flows/qha.py
JaGeo: I think yes because abinit has different names.

JaGeo: See https://github.com/materialsproject/atomate2/blob/1ebc9a77a531362ebfbb8adbd874f9c29bfa520b/src/atomate2/abinit/jobs/base.py#L140

JaGeo:
@utf I would be really happy if this could be merged. I would like to show this in an upcoming presentation (~2 weeks).

PR #910, opened 2024-07-02
add additional fields as kwargs to PhononBSDOSDoc

### LINKED ISSUE #904: BUG: add_metadata_to_flow poweup does not work on PhononBSDOSDoc
# Summary

Ideally, the `atomate2.vasp.powerup.add_metadata_to_flow` should add the information supplied by the user as  "additional_fields" of the taskdoc. But as per the `PhononBSDOSDoc `schema, this field does not exists and powerup does not have indented effect

# Solution

Add the missing `additional_fields` field to the `PhononBSDOSDoc`. (Will be addressed in a PR shortly)

naik-aakash:
Closes #904

# Changes

1. Add `additional_fields` as an acceptable kwargs to `generate_frequencies_eigenvectors_kwargs` (users can add metadata to the output taskdoc from here)

# Open question

1.  Should we rename the `generate_frequencies_eigenvectors_kwargs` to `task_document_kwargs`? This will help us use the powerups to add the metadata, but it would be a big change to the current workflow implementation. What do you think?

### DIFF
--- a/src/atomate2/common/schemas/phonons.py
+++ b/src/atomate2/common/schemas/phonons.py
@@ -269,6 +269,7 @@ class PhononBSDOSDoc(StructureMetadata, extra="allow"):  # type: ignore[call-arg
         **kwargs:
             additional arguments
         """
+        additional_fields = kwargs.get("additional_fields", {})
         factor = get_factor(code)
         # This opens the opportunity to add support for other codes
         # that are supported by phonopy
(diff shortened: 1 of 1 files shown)


src/atomate2/common/schemas/phonons.py
```diff
@@ -514,6 +515,8 @@ def from_forces_born(
             },
         )
 
+        return doc.model_copy(update=additional_fields)
```
JaGeo, line 518: Just wondering if that isn't very slow. Is there no other way to update the document?

naik-aakash: I will check again. I did not find any other way to do it so far . I also had looked in emmet for vasp doc 

https://github.com/materialsproject/emmet/blob/fdbcae26e0a00c22186bfdb2899c4cc929ba7125/emmet-core/emmet/core/tasks.py#L604.

JaGeo: I see

utf:
Thanks!

PR #923, opened 2024-07-19
Implementation to adjust get_supercell_size to also generate orthorhombic supercells

QuantumChemist:
## Summary

I'm adjusting `get_supercell_size` to also generate orthorhombic supercells.
This implementation is depending on https://github.com/materialsproject/pymatgen/pull/3938


QuantumChemist:
I don't exactly understand how I caused the `TypeError: get_supercell_size() missing 1 required positional argument: 'prefer_90_degrees'` and the other problems with the phonon unit tests with my changes, yet 😅

pyproject.toml
```diff
@@ -101,7 +101,7 @@ strict = [
     "pydantic-settings==2.3.4",
     "pydantic==2.8.2",
     "pymatgen-analysis-defects==2024.7.19",
-    "pymatgen==2024.6.10",
+    "pymatgen@git+https://github.com/materialsproject/pymatgen.git@8e392947348c75220cfb79309fe8fc9745edc580",  # has to be adjusted with new pymatgen release
```
QuantumChemist, line 104: I have fixed the pymatgen version to the merge that contains the changes in `CubicSupercellTransformation` in pymatgen.

QuantumChemist:
The failing AIMS unit tests have been addressed here https://github.com/materialsproject/atomate2/pull/928

pyproject.toml
utf: Thanks @QuantumChemist but we can't merge until a new pymatgen version has been released and this has been updated.

QuantumChemist: Hi @utf , Ok when the new pymatgen release is out, I will also adjust it here :)

QuantumChemist:
Hey @utf , I updated the pymatgen version as they released a new version. The failing unit tests still stem from aims. Thank you 😃

QuantumChemist:
I also saw that the pymatgen version gets updated anyway :) https://github.com/materialsproject/atomate2/pull/953

QuantumChemist:
Hey @utf , I just wanted to ask you if my PR is now ok to be merged? Thank you 😃

src/atomate2/common/jobs/phonons.py
```diff
@@ -57,32 +57,46 @@ def get_total_energy_per_cell(
...
+        min_atoms=kwargs.get("min_atoms"),
+        force_diagonal=kwargs["force_diagonal"],
+        step_size=kwargs["step_size"],
+    )
 
     if not prefer_90_degrees:
         transformation = CubicSupercellTransformation(
-            **common_kwds, max_atoms=kwargs.get("max_atoms"), force_90_degrees=False
+            **common_kwds,
+            max_atoms=kwargs.get("max_atoms"),
```
JaGeo, line 97: why is this not part of theother dict?

QuantumChemist: It was like this in the original implementation, so I left it like that.

JaGeo: Okay, fine

QuantumChemist: If you would like me to, I could move it to common_kwds as I don't think it will cause any backwards incompatibility issues and I would find it more systematic to have all kwargs in common_kwds.

JaGeo: Sounds good

QuantumChemist: I forgot to push those changes, I notice now, but it doesn't affect the functionalities.

QuantumChemist: I finally thought of pushing this change 😅

JaGeo:
@QuantumChemist Thanks! I am happy to merge if the tests pass 😀

QuantumChemist:
> @QuantumChemist Thanks! I am happy to merge if the tests pass 😀

thank you! I'm going to fix them over the next few days as it's quite a few that fail.

QuantumChemist:
So... `openff-toolkit` is missing in the pyproject.toml, but actually, I have problems of getting it installed at all.
I encounter the problem:
`ERROR: Could not find a version that satisfies the requirement openff-toolkit (from atomate2) (from versions: none)
ERROR: No matching distribution found for openff-toolkit
`
Seems there is no version. I'm using Python 3.10. Shall I switch to another Python version?

utf:
@janosh @orionarcher are you able to advise. Presumably this should not need to be installed to use the rest of the code?

janosh:
> Presumably this should not need to be installed to use the rest of the code?

yes, i think the intent is to only raise an `ImportError` for `openff.toolkit` if a user tries to create any of the OpenMM makers. @QuantumChemist can you post the full stack trace?

the reason it's not in `pyproject.toml` is because the package is not on PyPI (only installable with `conda`/`mamba`)

orionarcher:
As @janosh said, `openff-toolkit` isn't on PyPI which would explain the original error. `openff-toolkit` *shouldn't* need to be installed to use any code outside the `openff` module (though certainly possible I made a mistake). A little more context for the error would be helpful!

QuantumChemist:
> As @janosh said, `openff-toolkit` isn't on PyPI which would explain the original error. `openff.toolkit` _shouldn't_ need to be installed to use any code outside the `openff` module and `openmm.generate` (though certainly possible I made a mistake). A little more context for the error would be helpful!

oh, I see. The issue came up because I tried `pytest test/*` to see which unit tests are currently failing. Then I will simply exclude the openff tests. Thank you!

QuantumChemist:
> > Presumably this should not need to be installed to use the rest of the code?
> 
> yes, i think the intent is to only raise an `ImportError` for `openff.toolkit` if a user tries to create any of the OpenMM makers. @QuantumChemist can you post the full stack trace?
> 
> the reason it's not in `pyproject.toml` is because the package is not on PyPI (only installable with `conda`/`mamba`)

installing with conda somehow messed up my conda env. I will simply exclude thes openff tests. Thank you 😃

QuantumChemist:
> > Presumably this should not need to be installed to use the rest of the code?
> 
> yes, i think the intent is to only raise an `ImportError` for `openff.toolkit` if a user tries to create any of the OpenMM makers. @QuantumChemist can you post the full stack trace?
> 
> the reason it's not in `pyproject.toml` is because the package is not on PyPI (only installable with `conda`/`mamba`)

ah and I forgot to post the whole error message:

```
ImportError while loading conftest '/home/certural/git/chris/atomate2/tests/openff_md/conftest.py'.
tests/openff_md/conftest.py:2: in <module>
    import openff.toolkit as tk
E   ModuleNotFoundError: No module named 'openff'
```

so this happens when running all unit tests at once.

janosh:
it's probably best to pass the file/directory path to only the tests you currently care about to `pytest` and leave running the complete test suite to CI. that said, @orionarcher will submit a PR to sprinkle in some `pytest.importorskip("openmm")` in the OpenMM test modules to avoid future issues.

a bunch of the FF tests also fail unless you have all the FF packages installed. so the test suite is not currently designed to be out-of-the-box runnable by contributors who only care about part of the code and don't have all optional deps installed

JaGeo:
@janosh is the issue with the defects part also still existing? I usually had to install the defects add-on for the tests.

janosh:
i'm not sure, haven't run the complete test suite locally in forever 😅

QuantumChemist:
> it's probably best to pass the file/directory path to only the tests you currently care about to `pytest` and leave running the complete test suite to CI. that said, @orionarcher will submit a PR to sprinkle in some `pytest.importorskip("openmm")` in the OpenMM test modules to avoid future issues.
> 
> a bunch of the FF tests also fail unless you have all the FF packages installed. so the test suite is not currently designed to be out-of-the-box runnable by contributors who only care about part of the code and don't have all optional deps installed

I usually do but I also run an extra pytest test/* because, I don't know in what submodules the code I changed it called. As you said, I will then let that be handled by the CI :)

QuantumChemist:
@utf could you please resolve the merge conflicts? I can't do it, it seems.  The relevant unit tests are passing locally for me.

orionarcher:
@QuantumChemist to resolve the merge conflict you can:
1. update your local main branch
2. merge main into your local adjust_get_supercell_size
3. push the merge commit to the remote adjust_get_supercell_size

That should do it!

QuantumChemist:
> @QuantumChemist to resolve the merge conflict you can:
> 
>     1. update your local main branch
> 
>     2. merge main into your local adjust_get_supercell_size
> 
>     3. push the merge commit to the remote adjust_get_supercell_size (this branch)
> 
> 
> That should do it!

Right, I'm so used to do it on the GitHub website that I didn't think of this possibility! Thank you 😄

QuantumChemist:
Regarding the failing unit tests:
cp2k fails because `self.data.get("total_energy")` yields three energies [-193.39152045030943, -193.39152037984607, -193.39152037984607], from which the first one corresponds to the last total energy entry in the given test file.

For the failing aims and lobster tests I have no idea 😅

QuantumChemist:
I don't understand what is going on with the unit tests... I will try with a fresh MP/atomate2/main branch and open another PR.

JaGeo:
@QuantumChemist there are a few problems currently. @janosh was working on some of them already.

QuantumChemist:
> @QuantumChemist there are a few problems currently. @janosh was working on some of them already.

ah, I see. I was just wondering why the unit tests don't fail for the MP repo or other PRs. I thought, I caused it. Then I will just wait :) Thank you!

janosh:
`matgl` is brittle as hell. had to skip 2 tests in https://github.com/materialsproject/atomate2/pull/990. once that merges, CI should be stable

QuantumChemist:
Thank you for merging this! :)

PR #938, opened 2024-07-30
MPMorph Flows

BryantLi-BLI:
This PR seeks to port the original [MPMorph](https://github.com/materialsproject/mpmorph) workflows from atomate to atomate2. These workflows are used to generate equilibrated amorphous/crystalline/glassy structures with MD runs via AIMD.

Tagging in @Tinaatucsd, @mcgalcode, @mattmcdermott, @esoteric-ephemera, @vir-k01, @sivonxay because they have used it recently or are apart of the original development team.

Completed

- Volume equilibration workflow meant to equilibrate structure without NPT via a series of NVT runs (Equilibration + production volume)
- Code independent equilibration and production workflow, with optional quench methods
- VASP workflows used in MP papers below recreated in this PR:
  - [Hui Zheng et al.](https://arxiv.org/abs/2402.00177)
  - [Aykol and Persson 2018](https://pubs.acs.org/doi/full/10.1021/acsami.7b14868)
  - [Murat et al.](https://www.science.org/doi/full/10.1126/sciadv.aaq0148)
- There is an additional feature with this PR to incorporate MLFF MD runs with the same flow.
- Multi quench options available:
- Fast quench (double relax and static)
- Slow quench (series of MD runs to simulate cooling with different temperature profiles)
- Started to add tests for VASP
- MLFF tests are completed
- Pytests 
- Packmol conversion to pymatgen interface (used to generated initial amorphous structure)
- Slow quench temperature profiles

MISC

- Seeking feedback for code, missing features, usability

NOTE: Original PR closed and reopened due to some git commit history issues that caused large commit changes. This PR should have resolved the issue. Many thanks to @mcgalcode.


src/atomate2/vasp/flows/mpmorph.py
```diff
@@ -0,0 +1,539 @@
...
+    from typing import Any
+
+    from jobflow import Flow
+    from pymatgen.core import Structure
+
+    from atomate2.vasp.jobs.md import MDMaker
+
+
+@dataclass
+class BaseMPMorphVaspMDMaker(MPMorphMDMaker):
```
orionarcher, line 47: The differences between `BaseMPMorphVaspMDMaker`, `MPMorphVaspMDMaker`, and `MPMorphVaspMultiMDMaker` are not clear to me. Are all three needed? 

Also, is there a reason we have a `Base...` class here but no `Base...` class in the MLFF versions? I think minimizing the number of similar classes and standardizing the naming for the VASP and MLFF implementations would be helpful.

src/atomate2/vasp/flows/mpmorph.py
```diff
@@ -0,0 +1,539 @@
...
+        Use only for lowering temperarture for production run
+    md_maker : BaseMPMorphMDMaker
+        MDMaker to generate the molecular dynamics jobs specifically for MPMorph AIMD;
+        inherits from MDMaker (VASP)
+    steps_convergence: int | None = None
+        Defaults to 5000 steps unless specified
+    steps_single_production_run: int | None = 5000
+        Number of steps for a single production run; default 5000 steps.
+        If set and steps_total_production > steps_single_production_run,
+        multiple production runs (MultiMDMaker) will be generated.
```
orionarcher, line 80: My comments on the force field maker re `quench_maker_kwargs`, setting `Maker` parameters in the `args`, and the need for a `_post_init_kwargs` argument also apply here.

orionarcher (commented):
I am looking forward to using this once it's merged in! Thanks for all of this.

I left a number of comments and suggestions, mostly focused on parts of the interface I found confusing. Happy to say more if any of my comments are confusing or off-base. Y'all have much more context on the workflows than I do.

src/atomate2/common/jobs/mpmorph.py
```diff
@@ -0,0 +1,372 @@
...
+    return sum(vols) / counts
+
+
+def get_random_packed_structure(
+    composition: Composition | str,
+    target_atoms: int = 100,
+    vol_exp: float = 1.0,
+    tol: float = 2.0,
+    return_as_job: bool = False,
+    vol_per_atom_source: float | str = "mp",
```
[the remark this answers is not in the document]
esoteric-ephemera: Will change the default to MP + cached (no API key needed). We're aiming to have the base workflow (no user kw/args) reproduce the MPMorph workflow used [here](https://arxiv.org/abs/2402.00177).

src/atomate2/forcefields/flows/mpmorph.py
```diff
@@ -0,0 +1,760 @@
...
+        Name of the flows produced by this maker.
+    temperature : float = 300
+        Temperature of the equilibrium volume search and production run in Kelvin,
+        default 300K
+    steps_convergence: int | None = None
+        Defaults to 5000 steps unless specified
+    md_maker : ForceFieldMDMaker
+        MDMaker to generate the molecular dynamics jobs specifically for MLFF MDs
+    steps_total_production: int = 10000
+        Total number of steps for the production run(s); default 10000 steps
```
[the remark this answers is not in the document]
esoteric-ephemera: The idea was to ignore the temperature specified by the Maker / input set and use a common temperature in the initial equilibration and production runs. That gets around user error of setting different ensemble conditions in the various parts of the workflow

Then `_post_init_update` is needed on the VASP side of things because the temperature schedule and other relevant MD variables are fixed in the VASP input set. Similarly, for forcefields, the temperature is a Maker attr

Maybe there's a different paradigm that's clearer / makes more sense? Open to suggestions

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,546 @@
...
+        """Call MD maker for slow quench.
+
+        TODO write docstr
+        """
+        if not isinstance(self.md_maker, (MDMaker, ForceFieldMDMaker)):
+            raise TypeError(
+                "***WORK IN PROGRESS*** md_maker must be "
+                "an MDMaker or ForceFieldMDMaker."
+            )
+        return self.md_maker.make(structure, prev_dir)
```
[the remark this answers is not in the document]
BryantLi-BLI: So currently, MPMorph is only adaptable for VASP and MLFFMDMakers. The biggest issue is that MLFFMD and VASPMD takes in the kwargs for temperature different. The goal of the call_md_maker function is for it to be empty, and only when you call a specific version of SlowQuench (e.g SlowQuenchVaspMaker), it rebuilds the call_md_maker function properly. 

I envisioned the commons function to be empty, and when it needs to be adapted to other code, a new call_md_maker will be written to overwrite it, like SlowQuenchVaspMaker. I am open to suggestions on how to better handle this issue

src/atomate2/vasp/flows/mpmorph.py
BryantLi-BLI: `BaseMPMorphVaspMDMaker` is specifically for users who wants to customize their MPMorph settings. This enables them to adapt to specific MD makers, temperatures, n_steps, quench makers, convergence makers, etc. 

I wrote the other two as a clean wrapper over `BaseMPMorphVaspMDMaker`. `MPMorphVaspMDMaker` is meant for the generic, simplest case MPMorph, and `MPMorphVaspMultiMDMaker` are for systems where a long runtime is expected. I was hesitant putting them into the final repo. My only case for keeping them is user convenience. 

Also 100% agree with you the use of `Base...` is inconsistent. I will workshop this and come up with a better name for all the main, user friendly makers

src/atomate2/forcefields/flows/mpmorph.py
```diff
@@ -0,0 +1,760 @@
...
+    quench_maker_kwargs : dict or None (default)
+        If a dict, options to pass to `quench_maker`.
+        Check atomate2.common.flows.mpmorph for SlowQuenchMaker docstring
+        Example for MLFFs: quench_maker_kwargs = {
+            "md_maker": LJMDMaker(name="LJ MD Maker"),
+            "quench_n_steps": 1000,
+            "quench_temperature_step": 500,
+            "quench_end_temperature": 500,
+            "quench_start_temperature": 3000,
+        }
```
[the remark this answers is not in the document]
BryantLi-BLI: I'm interested. I do believe it's possible to simplify `quench_maker_kwargs`. Currently I've written it this way to specifically separate FastQuenchMaker and SlowQuenchMaker. If I consolidate it into a giant quench maker, and within the kwargs it specifies how to switch between fast quench and slow quench that would be ideal. 

I can workshop this.

src/atomate2/forcefields/flows/mpmorph.py
```diff
@@ -0,0 +1,760 @@
...
+    def _post_init_update(self) -> None:
+        """Ensure that forcefield makers correctly set temperature."""
+        self.md_maker = self.md_maker.update_kwargs(
+            update={
+                "temperature": self.temperature,
+                "n_steps": self.steps_convergence,
+            },
+            class_filter=ForceFieldMDMaker,
+        )
+        self.convergence_md_maker = EquilibriumVolumeMaker(
```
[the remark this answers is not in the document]
BryantLi-BLI: Ah...excellent observation. Will fix immediately. Thanks!

src/atomate2/common/flows/mpmorph.py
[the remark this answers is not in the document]
orionarcher: Ahh got it, that makes sense! I'd suggest raising a `NotImplementedError` here if the intent is for this to be reimplemented by child classes.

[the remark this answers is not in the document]
BryantLi-BLI: Yep! @esoteric-ephemera already fixed this!

src/atomate2/forcefields/flows/mpmorph.py
```diff
@@ -0,0 +1,760 @@
...
+            postprocessor=MPMorphEVPostProcess(),
+        )  # TODO: check EV versus PV
+
+        self.production_md_maker = self.md_maker.update_kwargs(
+            update=dict(
+                name="Production Run MLFF MD Maker",
+                temperature=self.temperature,
+                n_steps=self.steps_total_production,
+            )
+        )
```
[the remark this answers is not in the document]
BryantLi-BLI: So as a general response to the layout of `MPMorphMakers` across the forcefield and vasp flows, we wrote it with the post init function to make it user friendly. Typically, if a user needs to use MPMorph either on VASP or MLFF, they would have an expected temperature and number of steps they would run or their specific system (it varies a lot depending on the type of amorphous structure). Moreover, MLFF and VASP will be expected to have different ranges of n_steps because of the different scales it can be run at. So we wrote `MPMorphVaspMaker` and `MPMorphMLFFMaker` with the intention for it to be user friendly in adjusting these specific parameters. 

For the savvy `atomate2` user, I would recommend directly using `MPMorphMDMaker` from `atomate2.common.flows.mpmorph`, which is just the barebones logic for the MPMorph workflow. Using `MPMorphMDMaker` gives the user full flexibility on how to customize their calculations, adaptable to any VASP/MLFF input set requirements, and even beyond VASP and MLFF if required.

I do believe the overall question: "should we include such packaged function?" is important to address. I do worry if it causes too much codes to be built up in atomate2 over time. If that is true, then we can ax the pre-written user friendly makers, and we need to build it into the docs for the MPMorph flows instead. However, I do see the benefit for an average user to just import one maker instead of 4 different ones, each with their sets, just to use MPMorph. I am open to suggestions and feedback on this

src/atomate2/forcefields/flows/mpmorph.py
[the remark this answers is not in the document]
orionarcher: Do you think that differentiating in the `quench_maker` itself would work? Something like:

```python
quench_maker = FastQuenchMaker(**quench_maker_kwargs)
MPMorphMLFFMaker(
    ...,
    quench_maker=quench_maker
)
```
for fast quench

```python
quench_maker = SlowQuenchMaker(**quench_maker_kwargs)
MPMorphMLFFMaker(
    ...,
    quench_maker=quench_maker
)
```
for slow quench

src/atomate2/forcefields/flows/mpmorph.py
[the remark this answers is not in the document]
orionarcher: > we wrote it with the post init function to make it user friendly

This makes a lot of sense. Figuring out all of the parameters that need to be set for a maker is confusing and I think having easy and good defaults is ideal. I wonder if you could get the same effect with a user-friendly class method `MPMorphMLFFMaker.from_temp_and_steps` method (or some other `from_X` name) that automatically instantiates, `_post_init_updates`, and returns the makers. That would allow for a streamlined `__init__` signature and a beginner-friendly instantiation option.

```python
morph_maker = MPMorphMLFFMDMaker.from_temp_and_steps(
    temperature=300,
    steps_convergence=5000,
    steps_total_production=10000
```

```
morph_maker = MPMorphMLFFMDMaker(
    convergence_md_maker=...,
    quench_maker=...,
    production_md_maker=...,
)
```

It'd also allow for removing the `md_maker` arg. Currently, if `md_maker` and `convergence_md_maker` are both set, `md_maker` is silently ignored. Removing `md_maker` and instead having users set it within `convergence_md_maker` seems a bit cleaner.

> I do worry if it causes too much codes to be built up in atomate2 over time.

I definitely share this concern, there are a great many classes that are highly similar. I personally like the paradigm of having one class with multiple `from_XYZ` methods but there are tradeoffs there too.

src/atomate2/forcefields/flows/mpmorph.py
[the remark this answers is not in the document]
orionarcher: As I responded above, I think a `MPMorphMLFFMaker.from_temp_and_steps` could be the solution here. Curious to hear what y'all think.

src/atomate2/forcefields/flows/mpmorph.py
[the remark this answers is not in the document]
esoteric-ephemera: We removed `quench_maker_kwargs` entirely in favor of the user explicitly setting the `quench_maker` - no post init updates needed

src/atomate2/forcefields/flows/mpmorph.py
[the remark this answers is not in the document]
esoteric-ephemera: We've added a `from_temperature_and_steps` method to the common, MLFF, and VASP `MPMorphMDMaker`'s. This lets the user input a common temperature. number of steps for the runs,  MDMaker / MD set, and then optionally a quench step.

The class itself is now instantiated by setting `convergence_md_maker`, `production_md_maker`, and `quench_maker` explicitly

src/atomate2/forcefields/flows/mpmorph.py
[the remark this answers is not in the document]
esoteric-ephemera: Implemented as recommended - see above

src/atomate2/forcefields/flows/mpmorph.py
```diff
@@ -0,0 +1,760 @@
...
+    quench_temperature_step : float = 500
+        Temperature step for quench; default 500K drop
+    quench_n_steps : int = 1000
+        Number of steps for quench; default 1000 steps
+    """
+
+    name: str = "ForceField slow quench"
+    md_maker: ForceFieldMDMaker = field(default_factory=ForceFieldMDMaker)
+
+    def call_md_maker(
```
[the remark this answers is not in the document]
esoteric-ephemera: I think so - the VASP and forcefield MD ensemble attributes ($N$, $V$, $T$, and $p$) are set in different locations (the `VaspInputSetGenerator` vs. just top-level `Maker` kwargs) so this might have to stay. Open to suggestions

src/atomate2/vasp/flows/mpmorph.py
esoteric-ephemera: We've cleaned up the names a lot to make these more consistent, let us know if this looks better!

src/atomate2/vasp/flows/mpmorph.py
esoteric-ephemera: Cleaned up with previous comments, I think

BryantLi-BLI:
@utf @janosh @JaGeo: Should be ready for review!

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,518 @@
...
+                    - self.postprocessor.min_data_points
+                )
+                >= self.max_attempts
+            ):
+                # If the equilibrium volume is within the range of fitted volumes,
+                # or if the maximum number of attempts has been performed, stop
+                # and return structure at estimated equilibrium volume
+                final_structure = structure.copy()
+                final_structure.scale_lattice(working_outputs["V0"])
+                return final_structure
```
orionarcher, line 135: Consider

```python
            # Fit EOS to the running volume and energy from previous calculations
            self.postprocessor.fit(working_outputs)
            working_outputs = dict(self.postprocessor.results)

            # breaks whole flow here if EOS is not fitted properly
            if working_outputs.get("V0") is None:
                return Response(output=working_outputs, stop_children=True)

            # Check if the number of calculations has reached the maximum
            n_calcs = len(working_outputs["relax"]["volume"])
            max_calcs = self.postprocessor.min_data_points + (self.max_attempts or 1000)
            max_calcs_reached = n_calcs >= max_calcs

            v_0 = working_outputs.get("V0")
            v_min = working_outputs.get("Vmin")
            v_max = working_outputs.get("Vmax")

            # return final structure if calculated volume is in interpolation regime
            if max_calcs_reached or (v_min <= v_0 <= v_max):
                final_structure = structure.copy()
                final_structure.scale_lattice(v_0)
                return Response(output=final_structure)

            # Else, if the extrapolated equilibrium volume is outside the range of
            # fitted volumes, scale appropriately
            v_ref = v_min if v_0 < v_min else v_max
            eps_0 = (v_0 / v_ref) ** (1 / 3) - 1.0

            linear_strains = [np.sign(eps_0) * (abs(eps_0) + self.min_strain)]
```

orionarcher: LGTM, I might save intermediate variables for the `if` statements but ultimately a matter of taste.

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+    def make(
+        self,
+        structure: Structure,
+        prev_dir: str | Path | None = None,
+    ) -> Flow:
+        """
+        Create a flow with MPMorph molecular dynamics (and relax+static).
+
+        By default, production run is broken up into multiple smaller steps.
+        Converegence and quench are optional and may be used to equilibrate
```
janosh, line 215: typo

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+        Relax Maker for a second relaxation; useful for tighter convergence
+    static_maker : Maker
+        Static Maker
+    """
+
+    name: str = "fast quench"
+    relax_maker: Maker = Maker
+    relax_maker2: Maker | None = None
+    static_maker: Maker = Maker
+
```
janosh, line 324: is there any point in having generic `Makers` as defaults? nothing would happen unless you pass in explicit model/DFT `Makers`, no? in which case, maybe better not to set defaults at all?

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+            path to copy files from
+        working_outputs : dict or None
+            contains the outputs of the flow as it recursively updates
+
+        Returns
+        -------
+        .Flow, an MPMorph flow
+        """
+        if working_outputs is None:
+            if isinstance(self.initial_strain, float | int):
```
janosh, line 89: this `make` looks like it could benefit from more checking of assumptions and raising helpful error messages if they fail.

e.g. 

```
       if isinstance(self.initial_strain, float | int):
          self.initial_strain = (
              -abs(self.initial_strain),
              abs(self.initial_strain),
          )
      elif not isinstance(self.initial_strain, tuple) or len(self.initial_strain) != 2:
         raise ValueError(f"initial_strain must be float | tuple[float, float], got {initial_strain}")
```

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+        ensure the simulation does not hit wall time limits.
+
+    Check atomate2.vasp.flows.mpmorph for MPMorphVaspMDMaker
+
+    Parameters
+    ----------
+    name : str
+        Name of the flows produced by this maker.
+    convergence_md_maker : EquilibrateVolumeMaker
+        MDMaker to generate the equilibrium volumer searcher
```
janosh, line 190: typo

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+            high to low temperature
+        FastQuenchMaker - DoubleRelaxMaker + Static that
+            "quenches" structure at 0K
+    """
+
+    name: str = "Base MPMorph MD"
+    convergence_md_maker: Maker | None = None  # check logic on this line
+    # May need to fix next two into ForceFieldMDMakers later..)
+    production_md_maker: Maker | None = None
+    quench_maker: FastQuenchMaker | SlowQuenchMaker | None = None
```
janosh, line 204: maybe add a `__post_init__` to raise early if all (convergence_md_maker production_md_maker quench_maker) are left as `None`. based on the doc string sounds like maybe `production_md_maker` shouldn;t be allowed to be `None` anyway?

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+        cls,
+        temperature: float,
+        n_steps_convergence: int = 5000,
+        n_steps_production: int = 10000,
+        end_temp: float | None = None,
+        md_maker: Maker = None,
+        quench_maker: FastQuenchMaker | SlowQuenchMaker | None = None,
+    ) -> Self:
+        """
+        Create an MPMorph flow from a temperature and number of steps.
```
janosh, line 273: looks like this is missing an `@abstractmethod` decorator?

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+            name=self.name,
+        )
+
+    def call_md_maker(
+        self,
+        structure: Structure,
+        temp: float,
+        prev_dir: str | Path | None = None,
+    ) -> Flow | Job:
+        """Call MD maker for slow quench.
```
janosh, line 476: same here, should be marked @abstractmethod

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+        The production run can be broken up into smaller steps to
+        ensure the simulation does not hit wall time limits.
+
+    Check atomate2.vasp.flows.mpmorph for MPMorphVaspMDMaker
+
+    Parameters
+    ----------
+    name : str
+        Name of the flows produced by this maker.
+    convergence_md_maker : EquilibrateVolumeMaker
```
janosh, line 189: the name `convergence_md_maker` seems a bit inconsistent. maybe call this `equilibrium_volume_maker`?

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+        MDMaker to generate the equilibrium volumer searcher
+    production_md_maker : Maker
+        MDMaker to generate the production run(s)
+    quench_maker :  SlowQuenchMaker or FastQuenchMaker or None
```
janosh, line 193: double space after :

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,493 @@
...
+    min_strain : float, default = 0.5
+        Minimum absolute percentage linear strain to apply to the structure
+    max_attempts : int | None = 20
+        Number of times to continue attempting to equilibrate the structure.
+        If None, the workflow will not terminate if an equilibrated structure
+        cannot be determined.
+    """
+
+    name: str = "Equilibrium Volume Maker"
+    md_maker: Maker | None = None
```
janosh, line 59: `md_maker` is allowed to be `None` but looks like `make` will fail if it is? why not make this a required arg?

janosh (changes requested):

src/atomate2/common/flows/mpmorph.py
esoteric-ephemera: Changed the docstr generally since it was confusing

src/atomate2/common/flows/mpmorph.py
esoteric-ephemera: The idea was probably to enforce that these are required, but the workflow itself will fail if `relax_maker` and `static_maker` are unspecified. Changed their defaults to `None`

src/atomate2/common/flows/mpmorph.py
esoteric-ephemera: Added a `__post_init__` to this and the base quench maker classes in the same file

src/atomate2/common/flows/mpmorph.py
esoteric-ephemera: Added this decorator to a few other places where I can see it's needed (eos common flows, ASE base classes, OpenMM base class)

src/atomate2/common/flows/mpmorph.py
esoteric-ephemera: See above, this has a `__post_init__` for checking this is not `None` now

janosh: i think adding a `__post_init__` can be considered boilerplate. if you make `md_maker` a required arg and the user doesn't pass it, you `dataclass` should raise a similar error message 

```
    def __post_init__(self) -> None:
        if self.md_maker is None:
            raise ValueError("You must specify `md_maker` to use this flow.")
```

esoteric-ephemera: Right - that's there already

janosh: i know, i'm suggesting removing it and changing

```diff
- md_maker: Maker | None = None
+ md_maker: Maker
```
to make `md_maker` a required arg

esoteric-ephemera: Ahhh yes sorry! Changed as requested, leaving in the `__post_init__`'s as well tho

src/atomate2/ase/jobs.py
```diff
@@ -94,10 +92,9 @@ def run_ase(
         raise NotImplementedError
 
     @property
-    @abstractmethod
     def calculator(self) -> Calculator:
         """ASE calculator, method to be implemented in subclasses."""
-        raise NotImplementedError
+        return NotImplementedError
```
janosh, line 97: was this change accidental? can't return `NotImplementedError` here, doesn't match the return type `Calculator`. should keep the `raise` and the `@abstractmethod` decorator

BryantLi-BLI: I swapped out `return` with `raise`.

I'm a bit confused on the `@abstractmethod` decorator. It seems that every forcefield that inherits the ASEMaker adapts their calculator as `@property`. Does this imply that all the calculators should be swapped from `@property` -> `@abstractmethod` ? or just the original base class where the calculator should be absent?

esoteric-ephemera: whoops typo, think @BryantLi-BLI  fixed that

janosh: > Does this imply that all the calculators should be swapped from @property -> @abstractmethod ? or just the original base class where the calculator should be absent?

just the base class should have the `@abstractmethod` decorator. but all the `calculator` methods should have the `@property` decorator

pyproject.toml
```diff
@@ -138,6 +139,7 @@ changelog = "https://github.com/materialsproject/atomate2/blob/main/CHANGELOG.md
 
 [tool.setuptools.package-data]
 atomate2 = ["py.typed"]
+"atomate2.common.jobs" = ["*.json.gz"]
```
janosh, line 142: presumably this line is here to include `src/atomate2/common/jobs/mp_avg_vol.json.gz` with the PyPI package? probably best to pass the full file path to not auto-include future `json.gz` file in this dir by accident. but more importantly, that file is 650KB and needs to be downloaded by everyone who installs `atomate2` if we merge the current config, whether they use MPMorph flows or not. up to @utf to make a decision here but i would vote against adding that file to the package. better to auto-download it when the user first needs it with a link where to get it if user doesn't have an internet connection.

finally, i looked into that file. looks like it's partially redundant. at least the `oxi_state` and `count` keys can likely be removed?

Screenshot 2024-10-08 at 17 35 38

src/atomate2/ase/jobs.py
esoteric-ephemera: @BryantLi-BLI : that's my bad, the `AseMaker.calculator` should be both a property and abstractmethod. I accidentally removed this in [00d1434](https://github.com/materialsproject/atomate2/pull/938/commits/00d14347e6dc653b3aab0c4457b7bf7b14a21556) (was trying to see if there was a nice way to serialize ASE calculators without having to subclass AseMaker, and just forgot to restore a few things)

pyproject.toml
esoteric-ephemera: @janosh the oxidation states are I think ignored by default, but the counts are needed. If a given chemical environment doesn't exist in a database, the code looks through subspaces of that chemical environment and does a weighted average of structure volumes. The weighted average depends on the counts

I started rewriting this in parquet to jointly store them - might be better to put the parquet fiel on OpenData and pull down as needed

esoteric-ephemera: OK so for clarity: for the ICSD data, the user can either query from oxidation-state dependent or agnostic data, since ICSD includes manually-assigned charges. I pulled the MP data from summary rather than bonds, so the oxidation state info there is ignored.

That's a slight incongruity that I can change if requested

The counts are needed for the same reason given above (the code will sample subspaces of the total chemical space to perform a weighted averaging of reference volumes/atom if the total chemical space is not present in the dataset)

The data is now pulled / cached from figshare

Stuck with JSON since the parquet speed bump is probably not super relevant here (reading basically once rather than reading many times, and the user has the option to manually specify a DataFrame as input, if generating many random structures)

src/atomate2/common/flows/mpmorph.py
```diff
@@ -0,0 +1,528 @@
...
+            # and if so, if that limit has been reached
+            max_attempts_reached = len(working_outputs["relax"]["volume"]) >= (
+                (self.max_attempts or np.inf) + self.postprocessor.min_data_points
+            )
+
+            # Successful fit: return structure at estimated equilibrium volume
+            if v0_in_range or max_attempts_reached:
+                final_structure = structure.copy()
+                final_structure.scale_lattice(v0)
+                return final_structure
```
orionarcher, line 148: What if the return statements returned dictionaries instead of heterogenous outputs? e.g.

A successful return would look like:

```python
            if v0_in_range or max_attempts_reached:
                final_structure = structure.copy()
                final_structure.scale_lattice(v0)
                return {"structure": final_structure, "working outputs": working_outputs}
```

And a failing return:

```python
            if (v0 := working_outputs.get("V0")) is None:
                return Response(output={"working_outputs": working_outputs}, stop_children=True)

```

Advantages:

1. The API more closely mirrors the rest of atomate2 jobs, where we access the structure with `flow.output.structure`

2. By returning the working outputs at the end, there is better insight into how the job is working. I am currently finding the EOS fit is yielding unphysical densities when used with MLIPS. As is, it's a bit hard to collect the child jobs and inspect why the EOS fit is failing. If the working outputs were returned, that'd be very helpful.

3. The outputs can be evolved over time without breaking changes. If someone decided to include additional information in the output, that could be added as a new key:value pair without breaking anything.

esoteric-ephemera: Good suggestion! Added basically the same, but assigned the output prior to the `stress` and `energy` keys being popped from working outputs (that's the full EOS fit output, should be more useful for debugging)

esoteric-ephemera:
Hey @utf, @janosh and @JaGeo: just want to ensure we meet the deadline for including new WFs. This should be ready for review whenever you have time!

janosh (approved):
based on the error message, sounds like the auto-download for `_DEFAULT_AVG_VOL_FILE` returns a 404 HTML page (or some other error code) instead of actual data. maybe mock the data instead of an actual download in CI? other than that, this looks great! 👍 

for context, @orionarcher has been running these workflows for a while. everything working as expected

esoteric-ephemera:
@janosh good call and thanks to @orionarcher for testing this extensively!!

PR #940, opened 2024-08-02
Generalize forcefields for generic ASE calculator support

esoteric-ephemera:
This is a refactor to better support generic ASE calculators in atomate2, as well as molecules with ASE calculators.

Major changes:
- `schemas` / `utils` have been ported from the `forcefields` library to a separate `ase` library
- Forcefield-specific jobs now inherit from ASE base classes, such as the `ForceFieldTaskDocument` from the base `AseStructureTaskDoc` class
- Generic, `Structure`, and `Molecule`-specific ASE task doc classes
- Example ASE jobs using the Lennard-Jones and GFN<i>n</i>-xTB (TBLite) calculators
- The `ForcefieldResult` and `ForcefieldObject` classes alias ASE parent classes + deprecation warning
- Move MD makers from `forcefields` to `ase`
- Lazy load ase `.MolecularDynamics` objects
- XDATCAR support for trajectory output (sometimes useful)

Most of the other moved classes are internals for atomate2, so I don't expect many important consequences there


JaGeo:
@esoteric-ephemera Great! 
Even though the outside usage of `ForcefieldResult` might be limited. Let's maybe still alias - just in case.

esoteric-ephemera:
Hi @JaGeo, restored the `ForcefieldObejct` and `ForcefieldResult` objects for now with a deprecation warning

utf:
Hi @esoteric-ephemera, this is really nice to see in atomate2. Your implementation looks great to me. Would you be also happy to add static makers?

In terms of your todos:

> Replicating all "standard" flows (elastic, eos, gruneisen, phonon) with a base ASE calculator class?

I'm not sure if this makes sense? We'd still need to make subclasses for specific implementations anyway, so I don't think a base implementation would add anything?

> Tests for the ASE calculators currently in atomate2.ase.jobs (having some trouble getting TBLite working on an M1 Mac)

This would be great if you can figure it out.

esoteric-ephemera:
> I'm not sure if this makes sense? We'd still need to make subclasses for specific implementations anyway, so I don't think a base implementation would add anything?

Sounds good, I'll just port the MD maker over to `ase` from `forcefields` and call it

esoteric-ephemera:
Right now, I'm getting completely different behavior from `tblite` when running in CI vs. running locally (as in SCF doesn't converge in CI, works completely fine locally). I'm not sure if it's due to the `pytorch`-`tblite` incompatibility - see [here for example](https://github.com/tblite/tblite/issues/116).

For now, I've made the tblite dependence optional but left the relax / static / MD jobs + tests there. Happy to remove if preferable but I think it's good to have basic implementations of these in atomate2

utf:
Thanks @esoteric-ephemera. Would it be much effort to mock the tblite call so it doesn't have to run in CI? I'll also tag @Andrew-S-Rosen as maybe he has got tblite working for the quacc CI?

Eitherway, if this won't be possible, I'm happy to merge as is.

Andrew-S-Rosen:
It's due to the PyTorch inompatability. Not much you can do there except to split up the tests so you don't have both packages installed in the same env.

JaGeo:
@Andrew-S-Rosen I feel this might be a general problem in the future, especially with all the force fields🫠 

@utf maybe, we should think about installing and testing the forcefields separately as well if further conflicts show up.

src/atomate2/ase/md.py
```diff
@@ -0,0 +1,427 @@
...
+        The pressure in kilobar. If a sequence or 1D array, the pressure
+        schedule will be interpolated linearly between the given values. If a
+        float, the pressure will be constant throughout the run.
+    dynamics : str | ASE .MolecularDynamics | None = None
+        The dynamical thermostat to use. If dynamics is an ASE .MolecularDynamics
+        object, this uses the option specified explicitly by the user.
+        See _valid_dynamics for a list of pre-defined options when
+        specifying dynamics as a string.
+    ase_md_kwargs : dict | None = None
+        Options except for temperature and pressure to pass into the ASE MD function
```
orionarcher, line 113:
```suggestion
    ase_dynamics_kwargs : dict | None = None
        Options except for temperature and pressure to pass into the ASE `.MolecularDynamics` object.
```

I might be wrong about the behavior here but if not this is a bit more explicit.

src/atomate2/ase/md.py
```diff
@@ -0,0 +1,427 @@
...
+            # Use known dynamics if `self.dynamics` is a str
+            self.dynamics = self.dynamics.lower()
+            if self.dynamics not in _valid_dynamics[self.ensemble]:
+                raise ValueError(
+                    f"{self.dynamics} thermostat not available for {self.ensemble}."
+                    f"Available {self.ensemble} thermostats are:"
+                    " ".join(_valid_dynamics[self.ensemble])
+                )
+
+            md_func = _preset_dynamics[f"{self.ensemble}_{self.dynamics}"]
```
orionarcher, line 301: Maybe `dynamics` instead of `md_func` for consistency?

src/atomate2/ase/md.py
```diff
@@ -0,0 +1,427 @@
...
+    "nvt_nose-hoover": NPT,
+    "npt_berendsen": NPTBerendsen,
+    "npt_nose-hoover": NPT,
+}
+
+_preset_dynamics_defaults: dict = {
+    "nve": "velocityverlet",
+    "nvt": "langevin",
+    "npt": "nose-hoover",
+}
```
orionarcher, line 65: Could be good `Enums`

orionarcher (commented):
This is great and makes the interplay of ASE and atomate2 much clearer.

The programmatic logic all seems solid and well tested. Most of my comments are just "I was confused about the naming choices here". Feel free to take it or leave it.

src/atomate2/ase/jobs.py
```diff
@@ -0,0 +1,275 @@
...
+            self._make(ionic_configuration, prev_dir=prev_dir),
+            self.steps,
+            relax_kwargs=self.relax_kwargs,
+            optimizer_kwargs=self.optimizer_kwargs,
+            fix_symmetry=self.fix_symmetry,
+            symprec=self.symprec,
+            **self.task_document_kwargs,
+        ).to_meta_task_doc()
+
+    def _make(
```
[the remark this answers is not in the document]
esoteric-ephemera: Agreed, made this a public `run_ase` method

src/atomate2/ase/schemas.py
```diff
@@ -0,0 +1,479 @@
...
+class AseObject(ValueEnum):
+    """Types of ASE data objects."""
+
+    TRAJECTORY = "trajectory"
+
+
+class AseBaseModel(BaseModel):
+    """Base document class for ASE input and output."""
+
+    ionic_configuration: Optional[Union[Structure, Molecule]] = Field(
```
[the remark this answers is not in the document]
esoteric-ephemera: I think `mol_or_struct` is more informative / good implicit type hinting. I don't think we can change IonicStep to avoid breaking changes on the forcefield side of things

src/atomate2/ase/schemas.py
```diff
@@ -0,0 +1,479 @@
...
+            ase_calculator_name=ase_calculator_name,
+            included_objects=list(objects.keys()),
+            objects=objects,
+            is_force_converged=result.get("is_force_converged"),
+            energy_downhill=result.get("energy_downhill"),
+            dir_name=result.get("dir_name"),
+            **task_document_kwargs,
+        )
+
+    def to_meta_task_doc(self) -> Union[AseStructureTaskDoc, AseMoleculeTaskDoc]:
```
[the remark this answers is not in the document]
esoteric-ephemera: We need a public `from_ase_compatible_result` at least in the forcefield doc to prevent breaking changes. The `to_meta_task_doc` method is only used on the ase side, I'll try to clean that up a bit

src/atomate2/ase/md.py
esoteric-ephemera: You're right about the behavior and clarity of naming, but this is another situation where I'm trying to avoid breaking changes. I'll change the docstr tho, your suggestion is clearer

src/atomate2/ase/md.py
esoteric-ephemera: Agreed, I've changed this around a little bit and made sure that MD dynamics objects are lazy loaded

utf:
Is this still WIP or can it be merged now?

esoteric-ephemera:
@utf should be ready for review!

src/atomate2/ase/schemas.py
```diff
@@ -0,0 +1,495 @@
...
+    def __init__(self, **kwargs: Any) -> None:
+        kwargs = {
+            "final_mol_or_struct": None,
+            "trajectory": None,
+            "is_force_converged": None,
+            "energy_downhill": None,
+            "dir_name": None,
+            **kwargs,
+        }
+        super().__init__(**kwargs)
```
orionarcher, line 49: I am curious why this is a `dict` rather than a `BaseModel`, I am guessing it's for compatibility with `ASE`?

src/atomate2/ase/md.py
orionarcher: Got it, trying to adhere to prior API is very good reason to keep as is.

src/atomate2/ase/schemas.py
```diff
@@ -0,0 +1,526 @@
...
+            # PARTIAL options.
+            objects[AseObject.TRAJECTORY] = trajectory  # type: ignore[index]
+
+        output_doc = OutputDoc(
+            mol_or_struct=output_mol_or_struct,
+            energy=final_energy,
+            energy_per_atom=final_energy_per_atom,
+            forces=final_forces,
+            stress=final_stress,
+            ionic_steps=ionic_steps,
```
orionarcher, line 469: `ionic_steps` stores the forces on every atom at every step and thus is roughly the size of the trajectory. In my tests with a moderately sized system (100 atoms, 400 timesteps) the TaskDocument size is exceeding the 16 MB limit and causing a memory error. I don't think this is coming up in testing because it's all being run with a `MemoryStore`.

The easiest way I can think to solve this would be by adding `"ionic_steps"` to the `_ASE_DATA_OBJECTS` and `_FORCEFIELD_DATA_OBJECTS` before they are passed to `ase_job` and `forcefield_job`. More generally, I think anything that scales with `n_steps` should probably be included in `data` so that long trajectories don't run into memory errors. It'd be nice if everything that scales with n_steps was included in a single object but that'd be a significant and probably unnecessary re-work.

In the future, (not needed in this PR) a way to only save a subset of trajectory steps would be desirable.

orionarcher (changes requested):

src/atomate2/ase/schemas.py
esoteric-ephemera: Thanks @orionarcher , I'll take a look at your PR. For MD runs, only the energy should be saved in ionic steps by default

> In the future, (not needed in this PR) a way to only save a subset of trajectory steps would be desirable.

This is already possible with the `traj_interval` top-level kwarg to the ASE / forcefield MD makers, and by passing `relax_kwargs = {"interval" = <int> ,...}` through the `AseRelaxMaker`

orionarcher: Ok awesome, apologies for missing that. Switching to sparser reporting has ameliorated some of the speed and database issues I was having.

janosh:
@esoteric-ephemera @utf what's the timeline for merging this PR? with latest `numpy`, `scipy` and `pymatgen` (https://github.com/materialsproject/pymatgen/pull/4009) now being 3.10+, would make sense imo for `atomate2` to follow suit. given the size of this PR, that would likely result in merge conflicts so better the get this into `main` first

janosh:
question from looking at the changes in this PR: is the long-term goal to prevent further proliferation of `Makers` for each new MLFF that comes along (worthy goal imo)? having a single `ASE(MD|Static|Relax)Maker` which take a model name and a dict of model specific kwargs or (maybe simpler still) just a ready-to-go ASE-compatible `Calculator` seems much more maintainable. if that's the goal, i think this PR should deprecate `CHGNetRelaxMaker` and friends

https://github.com/materialsproject/atomate2/blob/808de7da064c488511326256552c3a7076f58d73/src/atomate2/forcefields/jobs.py#L202

JaGeo:
@janosh I would be happy about a deprecation period in such a case. We are using these Makers quite a lot and adapting codes/documentation will take time.

esoteric-ephemera:
>@esoteric-ephemera @utf what's the timeline for merging this PR? 

This is ready to go as is. I'd like to add NEB workflows to this (started for [VASP here](https://github.com/esoteric-ephemera/atomate2/tree/neb)) if I had a few more days, but no need if we'd like to merge sooner.

>  is the long-term goal to prevent further proliferation of Makers for each new MLFF that comes along (worthy goal imo)? having a single ASE(MD|Static|Relax)Maker which take a model name and a dict of model specific kwargs or (maybe simpler still) just a ready-to-go ASE-compatible Calculator seems much more maintainable. 

Yes exactly! The `Calculator` solution (+ model kwargs) is implemented in this PR. Having a LibXC-style lookup for forcefields is worthy of its own project, and is probably covered by MatGL for the graph-based MLFFs right?

It's early enough that we _could_ deprecate the existing MLFF makers but I don't see a strong reason to because they're in high use, as @JaGeo said. Think it's better to focus on preventing future bloat with more MLFF-specific calculators.

JaGeo:
@esoteric-ephemera @janosh I am fine with deprecation. I would just prefer a deprecation period in this case.

janosh:
> Having a LibXC-style lookup for forcefields is worthy of its own project, and is probably covered by MatGL for the graph-based MLFFs right?

i don't think so. matGL has chgnet and m3gnet but is unlikely to implement new models like 7net, ORB, Nequip, etc. at the rate they appear. many of these models have their own code bases. MatGL afaik aims to implement models with a consistent API internally, not act as as registry mapping model names to implementations in external code bases.
but that's all we need and it already exists in `atomate2` in the function `ase_calculator`

https://github.com/materialsproject/atomate2/blob/fdd2c9e2eed3f1fc631a3061f84d93e0898b2347/src/atomate2/forcefields/utils.py#L386-L458

> I don't see a strong reason to because they're in high use

the strong reason imo would be to get a leaner/cleaner `atomate2` code base and offer a single idiomatic way of using MLFFs through `atomate2`. to quote "The Zen of Python":

> “There should be one—and preferably only one—obvious way to do it.”

esoteric-ephemera:
@janosh Ok! I see now - so yes we'll still maintain `ase_calculator` and add models to that as needed, but we won't necessarily make subclasses where the MLFF is pre-assigned. That's totally OK with me, let me get the deprecation warnings in there

Also re: NEB, do we want to wait on that for now or push forwards? It could easily go in a separate NEB-related PR so no rush on that

janosh:
> It could easily go in a separate NEB-related PR

i think a separate PR makes a lot of sense

esoteric-ephemera:
Sounds good on NEB. I've added deprecation warnings to all the classes, and incorporated a suggestion from @orionarcher about the `force_field_name` attr of the static, relax and MD makers: the user can input these as either `str(MLFF.<attr>)`, `MLFF.<attr>` as an enum, or the forcefield name as a string (e.g., `"CHGNet"`). Makes sense to have that now since we're deprecating these classes, and it's easier for the user this way

.github/workflows/testing.yml
```diff
@@ -62,8 +63,8 @@ jobs:
           mkdir -p ~/.abinit/pseudos
           cp -r tests/test_data/abinit/pseudos/ONCVPSP-PBE-SR-PDv0.4 ~/.abinit/pseudos
           pip install .[strict,tests,abinit]
-          pip install torch-runstats
           pip install --no-deps nequip==0.5.6
+          pip install tblite==0.3.0
```
janosh, line 67: should `tblite` be added to pyproject

.github/workflows/testing.yml
```diff
@@ -32,6 +32,7 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: true
```
janosh, line 35: IIUC `fail-fast: true` will reduce the amount of signal you get from a CI run. e.g. if a job running on 3.9-3.12 fails on 3.9 first, you won't know if it also would have failed on 3.10, 3.11, etc. is this intended?

.github/workflows/testing.yml
```diff
@@ -75,7 +76,49 @@ jobs:
...
+          pip install --no-deps nequip==0.5.6
+
+      - name: Install pymatgen from master if triggered by pymatgen repo dispatch
+        if: github.event_name == 'repository_dispatch' && github.event.action == 'pymatgen-ci-trigger'
+        run: pip install --upgrade 'git+https://github.com/materialsproject/pymatgen@${{ github.event.client_payload.pymatgen_ref }}'
+
+      - name: Test
+        env:
+          MP_API_KEY: ${{ secrets.MP_API_KEY }}
+        run: pytest --cov=atomate2 --cov-report=xml tests/forcefields
```
janosh, line 121: should we merge the force field tests into the main test run? ideally, we'll use `pytest-split` later to split workload into multiple runners equally to speed up CI

.github/workflows/testing.yml
esoteric-ephemera: I mostly added this for my ease of figuring out why tests were failing initially - I can revert if you want

.github/workflows/testing.yml
esoteric-ephemera: The way it's partitioned now is because `tblite` is incompatible with `torch`. The safest approach, to me, seemed like completely quarantining the `torch`-dependent modules (just `forcefields`) from everything else. Open to suggestions

.github/workflows/testing.yml
esoteric-ephemera: Yeah, I'll put this in an `ase-ext` requirements. The base ASE stuff does not require `tblite` in any way

src/atomate2/ase/md.py
```diff
@@ -0,0 +1,444 @@
...
+    zero_angular_momentum : bool = False
+        Whether to initialize the atomic velocities with zero angular momentum
+    verbose : bool = False
+        Whether to print stdout to screen during the MD run.
+    """
+
+    name: str = "ASE MD"
+    time_step: float | None = None
+    n_steps: int = 1000
+    ensemble: Literal["nve", "nvt", "npt"] = "nvt"
```
janosh, line 157: might make sense to define

```py
class Ensemble(Enum):
    nve = "nve"
    nvt = "nvt"
    npt = "npt"
```

instead of `Literal["nve", "nvt", "npt"]`

.github/workflows/testing.yml
```diff
@@ -75,7 +74,48 @@ jobs:
...
-        run: pytest --cov=atomate2 --cov-report=xml
+        run: pytest --ignore=tests/forcefields --cov=atomate2 --cov-report=xml
+
+      - uses: codecov/codecov-action@v1
+        if: matrix.python-version == '3.10' && github.repository == 'materialsproject/atomate2'
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: ./coverage.xml
+
+  test-forcefields:
```
janosh, line 85: i'd recommend merging all tests back into a single job (which we can then split with `pytest-split`)

tests/ase/test_utils.py
```diff
@@ -0,0 +1,156 @@
...
+
+if TYPE_CHECKING:
+    from pymatgen.core import Structure
+
+
+from atomate2.ase.utils import AseRelaxer, FrechetCellFilter, TrajectoryObserver
+
+
+def test_safe_import():
+    assert FrechetCellFilter is None or FrechetCellFilter.__module__ == "ase.filters"
```
janosh, line 21: do we still need this test? given that we pin `"ase>=3.23.0"` in pyproject, `FrechetCellFilter` will be importable always

janosh (commented):
spotted a few more minor things

.github/workflows/testing.yml
esoteric-ephemera: It seems like anything `torch`-dependent and `tblite` can't be installed in the same environment without the `tblite` tests crashing in CI. If I use `pytest-split`, I'd still need to separately run the forcefield tests to have separate environments, no?

Outside of CI, having torch installed but not loaded seems OK with `tblite`. I can also just split off the `tblite` / `ase-ext` tests and then use `pytest-split` on the rest

Let me know what you think is best

tests/ase/test_utils.py
esoteric-ephemera: Don't think so - removed this and the logic around safely trying to import `FrechetCellFilter`

src/atomate2/ase/md.py
esoteric-ephemera: Added as suggested!

.github/workflows/testing.yml
janosh: > It seems like anything torch-dependent and tblite can't be installed in the same environment without the tblite tests failing in CI.

oh, i forgot about that. in that case, could you verbatim copy your above comment into the `test.yml` to leave a paper trail on why the tests are split into 2 jobs? i do think it would be better to run the `tblite` in its own job and move the force fields into the main job. will put some pressure on me to actually implement the `pytest-split` to warrant the effort. 😅

esoteric-ephemera: OK! Split off the ase and jupyter notebook tests into separate test instances. Also set up the `pytest-split` logic (3 splits per version seems OK), might need some help in getting that optimized

janosh: `pymatgen` uses 10 splits. i think `atomate2` could use more than 3, maybe 5 is the sweet spot? how did you generate the durations file? ideally, all splits should take about the same amount of time. currently, 1 is ~35 min, 2 is 15-20 min and 3 is 5 - 10 min so i think there's something off. for comparison, `pymatgen` uses

```
pytest --store-durations --durations-path tests/files/.pytest-split-durations
```

to update the file

esoteric-ephemera: I copied all the pytest split logic from pymatgen 😂 same command for the split durations test, the CI tests just take a very different amount of time in CI vs. my machine. Guessing there's MPS hardware acceleration turned on by default in the forcefields

I'll play around with this a bit - maybe setting the default device to be CPU would help for getting a better CI timing estimate

janosh: the other option would be to temp change the CI run to

```sh
pytest --store-durations --durations-path=DURATIONS_PATH
cat DURATIONS_PATH
```

and then copy pasting the actual CI durations from the log into your local file and committing that

janosh:
@esoteric-ephemera might be cleanest and fastest to branch off a `pytest-split` branch from this PR and open that in a separate PR, then `git reset --hard` to back out the splitting commits from here so we can review the force field overhaul and CI splitting stuff separately

janosh:
would give better git history but i'm happy either way (if it saves time)

esoteric-ephemera:
Hi @utf should be ready to merge provided the tests pass (need to make sure these still run OK after resolving merge conflicts with OpenMM)

src/atomate2/ase/jobs.py
```diff
@@ -0,0 +1,308 @@
...
+        prev_dir : str or Path or None
+            A previous calculation directory to copy output files from. Unused, just
+                added to match the method signature of other makers.
+
+        Returns
+        -------
+        AseStructureTaskDoc or AseMoleculeTaskDoc
+        """
+        return AseTaskDoc.to_mol_or_struct_metadata_doc(
+            getattr(self.calculator, "name", self.calculator.__class__),
```
janosh, line 169: i think this line will fail `pydantic` validation if a calculator should ever not have a name. `__class__` is not a string, we want the `__name__` attribute

https://github.com/materialsproject/atomate2/pull/940/files#diff-9eab185b405c88913d9a52dc90990b06235a6955a91c6c06ccbc899a4e3ec5b7R204-R207

janosh (approved):

JaGeo:
I that conext: should we document this and a few other of our new workflows that are not VASP-based more clearly? I think a lot of people have interest in this general interface but might not discover it easily

src/atomate2/ase/jobs.py
esoteric-ephemera: good catch!

esoteric-ephemera:
@JaGeo agreed, this should probably go in the docs for clarity. Also happy to add to the working paper

JaGeo:
> @JaGeo agreed, this should probably go in the docs for clarity. Also happy to add to the working paper

Personally, I think it should be also in the paper. Thoughts, @utf ?

utf:
If you're happy to add to the paper, I think it should be highlighted.

JaGeo:
And thanks @esoteric-ephemera (bit confused today 🙈)

PR #992, opened 2024-09-25
Bump min supported Python to 3.10

janosh:
following suit with `pymatgen` which in turn follows `numpy`, `scipy` and the rest of the NumFOCUS stack, this PR drops Python 3.9 support.

@utf curious to hear your thoughts. we may want to add 3.12 to CI in the same PR. another open question is whether we want to set `zip`'s new `strict` keyword to `True`. it defaults to `False` which preserves the old behavior of stopping whenever the 1st iterator is exhausted. `True` instead throws an error if all iterators don't have the same length

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ atomate2 workflows can be run using the [FireWorks] software. See the
 
 ## Installation
 
-Atomate2 is a Python 3.8+ library and can be installed using pip. Full installation
+Atomate2 is a Python 3.10+ library and can be installed using pip. Full installation
 and configuration instructions are provided in the [installation tutorial][installation].
 
 ## Tutorials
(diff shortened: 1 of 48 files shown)


utf:
It's already time for a new conda env 😨. Thanks for this @janosh. It makes sense to follow suit with pymatgen.

utf (approved):

README.md
```diff
@@ -5,6 +5,7 @@
 [![pypi version](https://img.shields.io/pypi/v/atomate2?color=blue)](https://pypi.org/project/atomate2)
 ![supported python versions](https://img.shields.io/pypi/pyversions/atomate2)
```
QuantumChemist, line 6: Here is a second python badge still listing 3.9.

janosh: that badge pulls the data it displays from PyPI, so will auto-update once this PR is merged and released

QuantumChemist: Oki, good to know :)

janosh:
@utf the required CI runs for auto-merging need updating. 3.9 is still required but should be replaced with 3.12

PR #993, opened 2024-09-26
Excise openff dependency from openmm testing

orionarcher:
This PR removes the openff dependency from the openmm test suite. This conflict originally arose in PR #923.


orionarcher:
@janosh would you mind taking a look?

src/atomate2/openmm/utils.py
```diff
@@ -2,16 +2,28 @@
...
+from emmet.core.openmm import OpenMMInterchange
+from openmm import LangevinMiddleIntegrator, XmlSerializer
+from openmm.app import PDBFile
+
 if TYPE_CHECKING:
+    from contextlib import suppress
+
     from emmet.core.openmm import OpenMMTaskDocument
 
+    with suppress(ImportError):
```
janosh, line 24: `TYPE_CHECKING` is always `False` at run time so no need to `suppress` errors here. also, `suppress` is quite slow compared to `try/except` so would usually opt for that

orionarcher: I initially did try/except but it was caught by the linter. should I override then?

orionarcher:
All fixed, thanks for review @janosh

src/atomate2/openmm/utils.py
janosh: yes, we should ignore that `ruff` rule that recommends `contextlib.suppress`. bad rule imo

janosh (approved):
👍

janosh (approved):
looks great! 👍 thanks for the quick turnaround on https://github.com/materialsproject/atomate2/pull/923#issuecomment-2377150112 @orionarcher

PR #996, opened 2024-09-28
Fix `prev_dir` behavior in input set generator of `MPGGAStaticMaker`

### LINKED ISSUE #995: BUG: INCAR parameter inheritance with `prev_dir` has changed for `MPGGAStaticMaker`
**Describe the bug**

There is a change in behavior for how INCAR parameters are inherited from previous directories. I believe it to be potentially undesired.

Take the following example:

```python
from atomate2.vasp.jobs.mp import MPGGAStaticMaker

structure = Structure.from_file("test/POSCAR")
input_set = MPGGAStaticMaker().input_set_generator.get_input_set(structure=structure, potcar_spec=True, prev_dir="test")

print(input_set.incar["ENCUT"])
```
[test.zip](https://github.com/user-attachments/files/17175684/test.zip)

The above example is generating an `MPStaticSet` of parameters while building upon a prior directory's run via the `prev_dir` keyword argument. The prior directory in this case (`test`) has some parameters that are intentionally distinct from the `MPStaticSet` for the sake of demonstration. For instance, let's track ENCUT, which is set to 123 in `test/vasprun.xml` whereas `MPGGAStaticMaker` has it set to 520.

In atomate2 0.0.14, I get ENCUT = 520. In other words, the `MPGGAStaticMaker` parameters are taking priority here.

In atomate2 0.0.15+, I get ENCUT = 123. In other words, the previous directory takes priority and overrides the settings by the `MPGGAStaticMaker`.

My question is: is this intentional? My gut feeling is no. For instance, if you are continuing a calculation from a previous directory but are now running with tighter settings, those tighter settings will get ignored.   We can compare with what was [in atomate2 <=0.0.14](https://github.com/materialsproject/atomate2/blob/58703efccaebe87376ef2fdb58ef8bcea3816e09/src/atomate2/vasp/sets/base.py#L422-L425). As you can see in the linked section of the codebase, `prev_dir` previously did not update the INCAR parameters by default and instead was only used to update things such as KSPACING due to the prior calculation's band gap. The only time it would update the INCAR parameters was if `inherit_incar=True` in `VaspInputSetGenerator`, which is `False` by default.

This is ultimately due to the switch to pymatgen-based input sets, but then why is nothing breaking in CI as a result? CC'ing @esoteric-ephemera.

Andrew-S-Rosen:
Closes #995.

### DIFF
--- a/src/atomate2/vasp/jobs/mp.py
+++ b/src/atomate2/vasp/jobs/mp.py
@@ -55,7 +55,7 @@ class MPGGARelaxMaker(BaseVaspMaker):
 
     name: str = "MP GGA relax"
     input_set_generator: VaspInputGenerator = field(
-        default_factory=lambda: MPRelaxSet(force_gamma=True, auto_metal_kpoints=True)
+        default_factory=lambda: MPRelaxSet(force_gamma=True, auto_metal_kpoints=True, inherit_incar=False)
     )
 
 
@@ -91,7 +91,7 @@ class MPGGAStaticMaker(BaseVaspMaker):
     name: str = "MP GGA static"
     input_set_generator: VaspInputGenerator = field(
         default_factory=lambda: MPStaticSet(
-            force_gamma=True, inherit_incar=False, auto_metal_kpoints=True
+            force_gamma=True, auto_metal_kpoints=True, inherit_incar=False,
         )
     )
 
@@ -129,6 +129,7 @@ class MPPreRelaxMaker(BaseVaspMaker):
     input_set_generator: VaspInputGenerator = field(
         default_factory=lambda: MPScanRelaxSet(
             auto_ismear=False,
+            inherit_incar=False,
             user_incar_settings={
                 "EDIFFG": -0.05,
                 "GGA": "PS",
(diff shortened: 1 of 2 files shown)


janosh (commented):
glad you caught this @Andrew-S-Rosen. we absolutely need a regression test. this is as a big a gotcha as they come.

maybe we want to pass `inherit_incar=False` here as well to be explicit. it already defaults to `False` but that's reliant on a default value two parent classes away in a different repo. better be safe and repeat

https://github.com/materialsproject/atomate2/blob/9600cef746957aae34dfa4a4df1f1bb68deca5cf/src/atomate2/vasp/jobs/mp.py#L58

Andrew-S-Rosen:
@janosh: I was also wondering about that --- I had to dig through classes to see what value `inherit_incar` was being set to just to be sure. I will add `inherit_incar=False` to the other sets here that way when people make a new one, they will be more likely to ensure that this is still the case regardless of what happens upstream.

Definitely agree about the regression test. I wish I could be more help on this. If nobody is able to chip in, I'll see if I can carve out some time in the next week or two.

Andrew-S-Rosen:
@janosh: Under-promise and over-delivering over here --- I added a regression test for the `prev_dir` behavior with `MPGGAStaticMaker`. 

However, what is concerning is that there does not appear to be any pre-existing tests for `MPGGAStaticMaker` in `test/vasp/jobs/test_mp.py` so we may only be testing that `Maker` indirectly. That's for another PR though.

janosh (approved):
excellent, thanks a lot @Andrew-S-Rosen! 👍

Andrew-S-Rosen:
Would it be possible to get a bugfix release after this PR?

utf:
v0.0.17 should be released shortly @Andrew-S-Rosen

PR #997, opened 2024-09-28
Allow `MatPesStaticFlowMaker(static2=None)`

janosh:
enables running MatPES PBE-only statics, e.g. for quick debug runs

@esoteric-ephemera would be good to get your eyes on this

### DIFF
--- a/src/atomate2/vasp/flows/matpes.py
+++ b/src/atomate2/vasp/flows/matpes.py
@@ -50,7 +50,7 @@ class MatPesStaticFlowMaker(Maker):
             ),
         )
     )
-    static2: Maker = field(
+    static2: Maker | None = field(
         default_factory=lambda: MatPesMetaGGAStaticMaker(
             # start from pre-conditioned WAVECAR from static1 to speed up convergence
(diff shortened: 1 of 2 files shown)


esoteric-ephemera:
Looks fine to me, but wouldn't this be the same as `MatPesStaticFlowMaker(static1=None, static2=MatPesGGAStaticMaker, static3=None)`?

janosh:
> Looks fine to me, but wouldn't this be the same as `MatPesStaticFlowMaker(static1=None, static2=MatPesGGAStaticMaker, static3=None)`?

yes. just looks a bit unidiomatic to have `static2` if `static1=None`

esoteric-ephemera:
Fair but isn't the easier solution would be to use just `MatPesGGAStaticMaker` (no flow) or the flow with only `static2` set if it's more convenient set makers via kwargs in debugging

Also TBC, the PR is fine, just want to discuss other possible solutions

janosh:
all valid points. i agree the changes in this PR are not essential, but maybe nice to have. i'll leave that up to popular vote if anyone wants to weigh in over the next few days.

PR #999, opened 2024-09-29
Feat: Add DeepMD MLFF support

MoseyQAQ:
## Summary

I have integrated atomate2 with the DeepMD MLFF, enabling efficient high-throughput calculations using this powerful tool.

## Checklist

### DIFF
--- a/tests/forcefields/test_jobs.py
+++ b/tests/forcefields/test_jobs.py
@@ -543,7 +543,6 @@ def test_deepmd_static_maker(sr_ti_o3_structure: Structure, test_dir: Path):
     importorskip("deepmd")
 
     # generate job
-    # NOTE the test model is not trained on Si, so the energy is not accurate
     job = ForceFieldStaticMaker(
         force_field_name="DeepMD",
         ionic_step_data=("structure", "energy"),
@@ -556,7 +555,7 @@ def test_deepmd_static_maker(sr_ti_o3_structure: Structure, test_dir: Path):
     # validation the outputs of the job
     output1 = responses[job.uuid][1].output
     assert isinstance(output1, ForceFieldTaskDocument)
-    assert output1.output.energy == approx(-44.40017, rel=1e-4)  # change here
+    assert output1.output.energy == approx(-3723.09868, rel=1e-4)
     assert output1.output.n_steps == 1
     assert output1.forcefield_version == get_imported_version("deepmd")
 
@@ -576,7 +575,7 @@ def test_deepmd_relax_maker(
 ):
     importorskip("deepmd")
     # translate one atom to ensure a small number of relaxation steps are taken
-    sr_ti_o3_structure.translate_sites(0, [0, 0, 0.2])
+    sr_ti_o3_structure.translate_sites(0, [0, 0, 0.01])
     # generate job
     job = ForceFieldRelaxMaker(
         force_field_name="DeepMD",
@@ -594,11 +593,11 @@ def test_deepmd_relax_maker(
     output1 = responses[job.uuid][1].output
     assert isinstance(output1, ForceFieldTaskDocument)
     if relax_cell:
-        assert output1.output.energy == approx(-44.407, rel=1e-3)
-        assert output1.output.n_steps == 5
+        assert output1.output.energy == approx(-3723.098823, rel=1e-3)
+        assert output1.output.n_steps == 14
     else:
-        assert output1.output.energy == approx(-44.40015, rel=1e-4)
-        assert output1.output.n_steps == 5
+        assert output1.output.energy == approx(-3723.100111, rel=1e-4)
+        assert output1.output.n_steps == 9
 
     # fix_symmetry makes no difference for this structure relaxer combo
     # just testing that passing fix_symmetry doesn't break
--- a/tests/forcefields/test_jobs.py
+++ b/tests/forcefields/test_jobs.py
@@ -10,6 +10,8 @@ from pytest import approx, importorskip
 from atomate2.forcefields.jobs import (
     CHGNetRelaxMaker,
     CHGNetStaticMaker,
+    DeepMDRelaxMaker,
+    DeepMDStaticMaker,
     ForceFieldRelaxMaker,
     ForceFieldStaticMaker,
     GAPRelaxMaker,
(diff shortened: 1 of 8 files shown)


JaGeo:
@MoseyQAQ Thanks for your contribution.

As you might have seen, we want to deprecate the concrete implementations of the force fields in the future and just provide a general interface. I am not sure if it makes sense to add this new implementation just now and directly deprecate it in a few months.

MoseyQAQ:
@JaGeo Hi, thanks for your comment! I opted for concrete implementations because the [atomate2 workflow for calculating phonons](https://github.com/JaGeo/TutorialAtomate2Forcefields/blob/main/atomate2_workflow_tutorial/phonon.ipynb) tutorial uses concrete implementations. I aim to maintain consistency with the established practices in these resources to ensure a seamless user experience.

JaGeo:
@utf Opinions?

utf:
Thanks for this @MoseyQAQ. I agree with @JaGeo, there doesn't seem much point adding something already with a deprecation notice. I would remove the concrete implementations but leave the rest of your code. DeepMD is great to have in atomate2.

In the future, we should think about convenience functions to create entire flows with a specific forcefield. E.g., something like:
```python

from atomate2.forcefields.flows import ElasticMaker

maker = ElasticMaker.from_forcefield("DeepMD")
```

MoseyQAQ:
@utf Thank you!

MoseyQAQ:
@utf @JaGeo , Hi! I’m currently working on removing the old implementations as suggested and addressing the issue preventing the code from passing the unit tests (which might be related to the dependency version). I’ll keep you updated as I make progress.

tests/forcefields/test_md.py
```diff
@@ -143,8 +142,10 @@ def test_ml_ff_md_maker(
         for step in task_doc.objects["trajectory"].frame_properties
     )
 
-    with pytest.warns(FutureWarning):
-        name_to_maker[ff_name]()
+    # Skip the following test for DeepMD, since it doesn't have concrete implementations
```
JaGeo, line 145: Is it possible to implement this? Or do you see any hurdles?

MoseyQAQ: This part of the code checks whether the concrete Maker classes raise a `FutureWarning`, indicating they will be deprecated. As per our earlier discussions, I’ve removed the concrete implementations for `DeepMD`, so this class no longer exists. Therefore, I’ve addressed this by skipping the test for `DeepMD`.

MoseyQAQ:
@utf @JaGeo Hi! This PR is ready for review and could be merged.

tests/test_data/forcefields/deepmd/graph.pb
janosh, line 1: 5MB is big for a test file. can we swap this out for a smaller checkpoint? or maybe auto-download the checkpoint on the fly like some of the other models?

tests/forcefields/test_jobs.py
```diff
@@ -535,3 +535,66 @@ def test_nequip_relax_maker(
...
+        optimizer_kwargs={"optimizer": "BFGSLineSearch"},
+        relax_cell=relax_cell,
+        fix_symmetry=fix_symmetry,
+        calculator_kwargs={"model": test_dir / "forcefields" / "deepmd" / "graph.pb"},
+    ).make(sr_ti_o3_structure)
+
+    # run the flow or job and ensure that it finished running successfully
+    responses = run_locally(job, ensure_success=True)
+
+    # validation the outputs of the job
```
janosh, line 587: typo

janosh (commented):
thanks @MoseyQAQ! two minor comments, otherwise this looks great! 👍

tests/test_data/forcefields/deepmd/graph.pb
MoseyQAQ: I have removed the original file and set it to download from GitHub when the test starts automatically.

tests/forcefields/test_jobs.py
MoseyQAQ: fixed

tests/forcefields/conftest.py
```diff
@@ -13,3 +16,16 @@ def pytest_runtest_setup(item: Any) -> None:
...
     # For consistent performance across hardware, explicitly set device to CPU
     torch.set_default_device("cpu")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def download_deepmd_pretrained_model(test_dir: Path) -> None:
+    # Download DeepMD pretrained model from GitHub
+    file_url = "https://raw.github.com/sliutheorygroup/UniPero/main/model/graph.pb"
+    local_path = test_dir / "forcefields" / "deepmd" / "graph.pb"
+    response = requests.get(file_url, timeout=10)
```
janosh, line 26: `atomate2` doesn't currently depend on `requests` so not sure it's installed. if it is, only as transitive dep. maybe better to use std lib `urllib.request.urlretrieve`

MoseyQAQ:
@janosh Hi, I found that the test can't pass because `MACE` is unable to download the foundation model from `tinyurl.com`. In the latest release of `MACE`, the original `tinyurl.com` link has been replaced with `GitHub`. The previous `tinyurl.com` URL returns a `403` error, which prevents the model from being downloaded properly.

janosh:
thanks for letting me know. could you replace tinyurl.com with [one of the checkpoints from here](https://github.com/ACEsuit/mace-mp/releases/tag/mace_mp_0)? maybe [`2023-12-10-mace-128-L0_epoch-199.model`](https://github.com/ACEsuit/mace-mp/releases/download/mace_mp_0/2023-12-10-mace-128-L0_epoch-199.model)

esoteric-ephemera:
@utf : I added that kind of [convenience function for the forcefield EOS flows](https://github.com/materialsproject/atomate2/blob/68f7e071f5cfa7dc426af088578d0e0359cf08bd/src/atomate2/forcefields/flows/eos.py#L61) but didn't for more complex ones - happy to do that in a separate PR

utf:
Thanks @esoteric-ephemera and @MoseyQAQ. Is this ready for final review?

MoseyQAQ:
@utf Yes. @esoteric-ephemera Thank you!

janosh (approved):
this looks great. thanks @MoseyQAQ! 👍

PR #1006, opened 2024-10-04
Fix bump phonopy

naik-aakash:
# Changes

1. Update phonopy version
2. Switch off the Gruneisen parameter computations at gamma point following (https://github.com/phonopy/phonopy/issues/455#issuecomment-2495156556)

# Todo

- [x] Investigate what cause the change in gruneisen parameter value leading to failure in test
- [x] Fix failing phonon workflow with using latest phonopy version (v2.30.1)


JaGeo:
How large are the changes? Could you give more details? Thanks!

naik-aakash:
The average Gruneisen value now comes out to be  `1.271535057337552` using v2.28 (using v2.27 we used to get `1.1882292157682082`)

JaGeo:
Yeah, I think we need a bit more analysis here. This is quite a large change

naik-aakash:
The latest version of phonopy causes phonon workflow test to also fail (cause of failure seems to originate from force constants computation default being changed)

JaGeo:
I would need more details to act. E.g., How large is the change? Why does this only affect the Grüneisen parameter workflow? What is different?

We can do this privately if you prefer.

naik-aakash:
> I would need more details to act. E.g., How large is the change? Why does this only affect the Grüneisen parameter workflow? What is different?
> 
> We can do this privately if you prefer.

As I have mentioned above, now with latest version phonon workflow test is also affected. I will try to share the details in next days

JaGeo:
@naik-aakash one last question: do you need all test files still or can one set be removed?

naik-aakash:
> @naik-aakash one last question: do you need all test files still or can one set be removed?

Hi @JaGeo, Thanks for spotting this. I Have now removed the test files that are no longer used

JaGeo:
@naik-aakash thanks. This looks good. I will merge.

JaGeo:
For everyone else seeing this and wondering why we are changing tests:

By accident, one of the phonon tests used wrong test files. This only appeared as a problem when phonopy added more strict handling of the forces. 

The Grüneisen parameter computation was also affected by a phonopy change. However, as this mostly affected an unrealiable value at gamma, we simply remove it from the mesh computation. We should have likely done this already before.

PR #1008, opened 2024-10-06
`PhononMaker` add options to `calculate_pdos` and save force constants to file

chouyoudou:
## Summary

- Feature: Added functionality to compute and save projected phonon density of states (PDOS). The feature can be controlled by the `calculate_pdos` parameter in `BasePhononMaker.`

- Fix: Added functionality to control the saving of force constants in the phonopy parameters schema through the `store_force_constants` parameter in `BasePhononMaker`. This feature determines whether force constants are saved as part of the phonon data schema, and now it functions correctly as intended.


JaGeo:
Hi @chouyoudou ,

Thanks. I will review this. Just to mention, the `store_force_constants` parameter was referring to the schema and not the file. Should maybe be made clearer in the documentation.

chouyoudou:
> Hi @chouyoudou ,
> 
> Thanks. I will review this. Just to mention, the `store_force_constants` parameter was referring to the schema and not the file. Should maybe be made clearer in the documentation.

Thanks for pointing that out. I've updated the documentation to clarify that it refers to controlling whether the force constants are saved in the schema, rather than directly to a file.

src/atomate2/common/schemas/phonons.py
```diff
@@ -406,6 +406,16 @@ def from_forces_born(
...
+        # projected dos 
+        if kwargs.get("calculate_pdos", False):
+            phonon.run_mesh(kpoint.kpts[0], with_eigenvectors=True, is_mesh_symmetry=False)
+            phonon_dos_sigma = kwargs.get("phonon_dos_sigma", None)
+            dos_use_tetrahedron_method = kwargs.get("dos_use_tetrahedron_method", True)
+            phonon.run_projected_dos(
+                sigma=phonon_dos_sigma, use_tetrahedron_method=dos_use_tetrahedron_method
+            )
+            phonon.write_projected_dos()
+
```
JaGeo, line 418: Does it make sense to also add an automatic plot here? Likely a bit more work. I am fine with just adding this for now.

JaGeo: This directly relates to the next question: do you want the option to save it in the schema? If so, you would need to develop a parser.

chouyoudou: > Does it make sense to also add an automatic plot here? Likely a bit more work. I am fine with just adding this for now.

I think it could be useful to provide summed PDOS, for example by element type or symmetry-equivalent atoms. However, user needs in this area are often quite diverse, making it challenging to implement something that meets all needs as straightforwardly as the TDOS plot.


> This directly relates to the next question: do you want the option to save it in the schema? If so, you would need to develop a parser.

The current approach saves the PDOS in an external file, similar to the TDOS format. I believe a simple adjustment to the default filename to match the style of atomate2 (e.g., `phonon_pdos.yaml` instead of the default `projected_dos.dat` in phonopy) would suffice.

JaGeo: I am fine with this for now. But is it really a yaml? If not, rather keep the dat

chouyoudou: Actually, it is not a yaml; it was just renamed with a `.yaml` extension in atomate2 for TDOS.

chouyoudou (commented):
Apply pre-commit auto-fix for code formatting and linting issues

chouyoudou (commented):
add force_constants2file parameter for saving the force constants in a file

src/atomate2/common/flows/phonons.py
```diff
@@ -150,6 +154,8 @@ class BasePhononMaker(Maker, ABC):
     kpath_scheme: str = "seekpath"
     code: str = None
     store_force_constants: bool = True
+    force_constants2file: str = None
+    calculate_pdos: bool = False
```
JaGeo, line 158: Could you please make them part of "generate_frequencies_eigenvectors_kwargs" or at least the filename? Additionally, please use "create_force_constants_file" as an additional kwarg in "generate_frequencies_eigenvectors_kwargs" to write the force constant file? Otherwise, this will getvery messy. I would also rename "force_constants2file" to "force_constants_filename"

janosh:
@JaGeo does this PR have approval?

JaGeo:
@janosh with your updates, it is fine with me. 😀

JaGeo (approved):

PR #1012, opened 2024-10-08
Ferroelectric Workflow 2

fraricci:
Porting the ferroelectric workflow from atomate to atomate2.

This is a second pull request, since the first one #196 got  broken in some way.


src/atomate2/vasp/jobs/ferroelectric.py
```diff
@@ -0,0 +1,181 @@
...
+        Number of interpolations calculated from polar to nonpolar structures,
+        including the nonpolar.
+
+    Returns
+    -------
+    Path
+        Path to the directory containing the interpolated structures.
+    """
+    interp_structures = p_st.interpolate(np_st, nimages, autosort_tol=0.0)
+    dumpfn(interp_structures, "interp_structures.json")
```
utf, line 116: Rather than doing this, can you just have the `interpolate_structures` job return the list of structures and the `add_interpolation_flow` take a list of structures as input rather than a folder.

utf (commented):
Thanks @fraricci, I just noticed one final hangover from atomate1 (writing intermediate objects in a directory). Once that's fixed I'm happy to merge.

src/atomate2/vasp/jobs/ferroelectric.py
```diff
@@ -139,7 +139,7 @@ def add_interpolation_flow(
     outputs = {}
 
     for i, interp_structure in enumerate(interp_structures[1:-1]):
-        lcalcpol_maker.write_additional_data[""] = {
+        lcalcpol_maker.write_additional_data["structures:json"] = {
             "st_polar": interp_structures[0],
             "st_nonpolar": interp_structures[-1],
             "st_interp_idx": i + 1,
```
fraricci, line 145: Question: may I add the interpolated structure directly instead of its index?

utf: I think it makes sense to only keep the index to keep track. The structure will be available anyway as the input.

utf:
Thanks @fraricci

PR #1017, opened 2024-10-15
Allow custom mace model by specifying "model" in calculator kwargs"

orionarcher:
Currently it's only possible to use the mace_mp model. This PR allows users to use a custom model by making the "model" kwarg a path to the model. If `model` is not a `Path`, then `mace_mp` is used, thus this change should be backwards compatible.

EDIT: More context:

`mace==0.3.3`, the version currently used by atomate2, allows custom mace models in `mace_mp`. So the current code does not need any changes to use custom models.

`mace==0.3.7`, the most recent version does not allow custom models in mace_mp, which is a breaking change. This PR will be needed when `atomate2` bumps the mace version. When that happens, many tests will likely need fixing.

### DIFF
--- a/src/atomate2/forcefields/utils.py
+++ b/src/atomate2/forcefields/utils.py
@@ -67,12 +67,8 @@ def ase_calculator(calculator_meta: str | dict, **kwargs: Any) -> Calculator | N
             model = kwargs.get("model")
             if isinstance(model, str | Path) and os.path.isfile(model):
                 model_path = model
-                device = kwargs.get("device") or (
-                    "cuda" if torch.cuda.is_available() else "cpu"
-                )
                 calculator = MACECalculator(
                     model_paths=model_path,
-                    device=device,
                     **kwargs,
                 )
             else:
--- a/src/atomate2/forcefields/utils.py
+++ b/src/atomate2/forcefields/utils.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import json
-import os
 from contextlib import contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -64,7 +63,7 @@ def ase_calculator(calculator_meta: str | dict, **kwargs: Any) -> Calculator | N
             from mace.calculators import MACECalculator, mace_mp
 
             model = kwargs.get("model")
-            if isinstance(model, str | Path) and os.path.isfile(model):
+            if isinstance(model, str | Path) and Path(model).exists():
                 model_path = model
                 device = kwargs.get("device") or "cpu"
                 if "device" in kwargs:
--- a/src/atomate2/forcefields/utils.py
+++ b/src/atomate2/forcefields/utils.py
@@ -68,7 +68,8 @@ def ase_calculator(calculator_meta: str | dict, **kwargs: Any) -> Calculator | N
             if isinstance(model, str | Path) and os.path.isfile(model):
                 model_path = model
                 device = kwargs.get("device") or "cpu"
-                del kwargs["device"]
+                if "device" in kwargs:
+                    del kwargs["device"]
                 calculator = MACECalculator(
                     model_paths=model_path,
                     device=device,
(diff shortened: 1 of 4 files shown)


orionarcher:
More context:

`mace 0.3.3`, the version currently used by atomate2, allows custom mace models in `mace_mp`. So the current code does not need any changes to use custom models.

`mace 0.3.7`, the most recent version does not allow custom models in `mace_mp`, which is a breaking change. This PR will be needed when atomate2 bumps the mace version. When that happens, many tests will likely need fixing.

@janosh @utf, how does atomate2 intend to evolve the `mace` dependency over time? They introduced some breaking changes but also useful new features.

janosh:
> @janosh @utf, how does atomate2 intend to evolve the mace dependency over time? They introduced some breaking changes but also useful new features.

i think the latest version of `atomate2` should strive to be compatible with the latest version of `mace-torch`, even if that means adapting to breaking changes

JaGeo:
@orionarcher I will remove the orb stuff here. Could you make a separate pull request for the orb part and add a test there? E.g., by installing org separately? I think there should be at least one test.

JaGeo:
Oh, there seems to be something wrong with the kwargs. Will check later and then merge.

JaGeo:
@esoteric-ephemera Could you shortly have a look at this failing test? To me, it looks like the symmetry does not change during relaxation with the latest mace version. At least, within the ASERelaxer, the correct "relax_cell" and "fix_symmetry" are set.

esoteric-ephemera:
@JaGeo sure! Will look at this soon

src/atomate2/forcefields/utils.py
```diff
@@ -59,9 +61,18 @@ def ase_calculator(calculator_meta: str | dict, **kwargs: Any) -> Calculator | N
...
-            from mace.calculators import mace_mp
-
-            calculator = mace_mp(**kwargs)
+            import torch
+            from mace.calculators import MACECalculator, mace_mp
+
+            model = kwargs.get("model")
+            if isinstance(model, str | Path) and os.path.isfile(model):
+                model_path = model
+                calculator = MACECalculator(
```
esoteric-ephemera, line 70: Think we need to change this to:
```
device = kwargs.get("device") or "cpu"
calculator =  MACECalculator(
    model_paths=model_path,
    device = device,
    **kwargs
)
```

esoteric-ephemera:
OK so if the symmetry of a cell is fixed, we certainly should assert that the initial and final symmetries of the cell are the same

But if we don't fix the symmetry, there's a distinct possibility that the symmetry of the cell remains the same. It could change, but unless the force convergence criterion is made tighter, there's a decent chance the structure barely changes

For example, with the previous setting of `fmax` = 0.04 eV/Å, the atoms don't move much. With that tightened to 0.01 eV/Å, the symmetry of the cell changes during relaxation 

Also worth noting that it seems like ASE's FixSymmetry isn't holding for more stringent force convergence criteria. I pushed `fmax` to 0.001 eV/Å and the cell symmetry wasn't preserved with `fix_symmetry = True`

JaGeo:
@esoteric-ephemera so, you are okay with simply adapting the test, right?

src/atomate2/forcefields/utils.py
JaGeo: Thanks. One needs to remove the "device" key from the kwargs. Otherwise, it will complain that it is set twice. Will do that.

JaGeo:
@esoteric-ephemera @orionarcher Okay, I figured it out. `mace_mp` has a default floating type precision of float32. If we set this in the `MACECalculator`, we get the same results as before. Will merge this now.  (Sorry, took a little longer for me to see this even though it is kind of obvious)

JaGeo:
@utf @janosh I have one question regarding backward compatibility: I decided to go with the new default (floating point precision of the model that one trained). This might influence the results a bit, but I think it is smarter than setting float32 by default. If we should change this, let me know.

esoteric-ephemera:
@JaGeo  good with adapting the test as needed but good to know it was a dtype issue

PR #1031, opened 2024-10-28
Orb_support

orionarcher:
## Summary

Add support for the Orb forcefield calculator

This adds a dependency for `orb_models` and `pynanonflann`. The `orb_models` repo has a strict dependency on torch 2.2.0, though locally I have it working with 2.4.0. I wrote a test and added an `importorskip` call but did not modify the pyproject.toml. Having all models tested and installed in a single environment feels increasingly untenable. I don't want to add a new dependency that will mess with the resolver.

## Linting

  type check your code.

### DIFF
--- a/src/atomate2/forcefields/__init__.py
+++ b/src/atomate2/forcefields/__init__.py
@@ -16,6 +16,7 @@ class MLFF(Enum):  # TODO inherit from StrEnum when 3.11+
     NEP = "NEP"
     Nequip = "Nequip"
     SevenNet = "SevenNet"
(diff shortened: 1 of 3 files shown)


JaGeo:
@orionarcher agreed, pytorch conflicts are starting to get a nightmare. 

@utf any ideas on how to solve this in the future?

JaGeo:
I would recommend adding one test where you do not install all forcefields, just orb. 
Also, all other tests need to run through before we can merge here.

PR #1036, opened 2024-11-01
Single source of truth for optional deps in `pyproject.toml`

janosh:
plus some code cleanup

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_language_version:
 exclude: ^(.github/|tests/test_data/abinit/)
 repos:
 - repo: https://github.com/charliermarsh/ruff-pre-commit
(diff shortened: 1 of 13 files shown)


JaGeo:
Hi @janosh , I am a bit confused by this. As far as I understood strict, it was there to give exact version numbers. With this pull request, this would not be true anymore.

janosh:
@JaGeo you're right. i wanted to use this PR as a test to look into adding a [`uv.lock`](https://docs.astral.sh/uv/concepts/projects/#project-lockfile) file which would replace the `strict` optional deps entirely. it might be a step towards addressing the pytorch version conflicts you mentioned https://github.com/materialsproject/atomate2/pull/1031#issuecomment-2441823827

JaGeo:
@janosh ah, thank you for the explanation! 😅

PR #1037, opened 2024-11-01
Make testing utils sf

jmmshn:
## Summary

I think we can move some of the testing utils into the main code so that they are importable by other packages.

* Isolated the code used to mock vasp
* Moved the code to `atomate2.utils.testing.vasp.py`
* Modified imports in different places to point the the new location.

None


JaGeo:
Yes, very good idea!

JaGeo:
@utf I would merge it. I think it is a good idea. Just want to make sure you are fine with it as well.

utf:
Thanks @jmmshn. Looks good to me.

PR #1053, opened 2024-11-11
Add trajectory reporter to openmm workflow

orionarcher:
## Summary

* Allows the OpenMM trajectories to report pymatgen trajectories

## Checklist

title.

  type check your code.


orionarcher:
Would you mind taking a look @janosh?

src/atomate2/openmm/utils.py
```diff
@@ -171,3 +172,148 @@ def openff_to_openmm_interchange(
...
+        )
+        box_vectors = state.getPeriodicBoxVectors(asNumpy=True).value_in_unit(
+            omm_unit.angstrom
+        )
+
+        ev_to_kj_per_mol = 96.485
+
+        # Get energies in eV
+        kinetic_energy = (
+            state.getKineticEnergy()._value / ev_to_kj_per_mol  # noqa: SLF001
```
janosh, line 269: is there no public openMM API to get these values? seems fragile to rely on `_value` here which i'm guessing has no backwards compat gurantees?

src/atomate2/openmm/utils.py
```diff
@@ -171,3 +172,148 @@ def openff_to_openmm_interchange(
...
+            {
+                "kinetic_energy": kinetic_energy,
+                "potential_energy": potential_energy,
+                "total_energy": kinetic_energy + potential_energy,
+            }
+        )
+
+        self._nextModel += 1
+
+    def __del__(self) -> None:
```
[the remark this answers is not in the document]
orionarcher: Good point. I was mirroring an external API and being a bit. I replaced it with a save method to be more robust. 

The core problem is that the `Reporters` usually append to a file at each step but trajectory doesn't easily support that.

src/atomate2/openmm/utils.py
orionarcher: Good catch. I was a bit more clever and fixed it.

orionarcher:
There are other `Reporters` that all use the camelCase args. I think OpenMM call's some of the methods internally too, so it's not possible to change them. I'd advocate for sticking with that, even though I agree it's a bit heretical.

janosh:
looks good! thanks @orionarcher  👍

orionarcher:
This will need one more loop back @janosh

PR #1060, opened 2024-11-19
ABINIT DFPT SHG

VicTrqt:
## Summary

* ABINIT workflow to compute the second-harmonic generation (SHG) tensor in DFPT with its test
* Output files saving (DDB and POT)
* Job to remove and compress files from a previous Flow

## TODO

* Add the possibility of a scissor shift

## Note

This DFPT SHG workflow is mentioned in the draft as being under review. The initial plan was to include it fully in the upcoming paper, but I would understand this might be unfeasible at this stage of the process. Also, since the GW WF is explained in the draft, priority should be given to #816 .


JaGeo:
@VicTrqt Hi, thank you! 

There seem to be some failing abipy imports in the ase part of the tests. Could you check them? It might be that you should only import abipy when abipy is installed.

pyproject.toml
```diff
@@ -38,7 +38,8 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-abinit = ["abipy>=0.9.3"]
+#abinit = ["abipy>=0.9.3"]
+abinit = ["abipy @ git+https://github.com/abinit/abipy.git"] # tmp
```
JaGeo, line 42: Before we merge this, we would definitely need a new abipy version.

src/atomate2/abinit/schemas/anaddb.py
```diff
@@ -0,0 +1,391 @@
...
+        The electronic contribution to the dielectric tensor
+    """
+
+    structure: Union[Structure] = Field(
+        None, description="The final structure from the calculation"
+    )
+    dijk: Optional[list] = Field(
+        None, description="Conventional SHG tensor in pm/V (Chi^(2)/2)"
+    )
+    epsinf: Optional[list] = Field(
```
JaGeo, line 48:
```suggestion
    eps_inf: Optional[list] = Field(
```
?

src/atomate2/abinit/schemas/mrgddb.py
```diff
@@ -0,0 +1,391 @@
...
+#         self.ddb_as_str: str = ddb_as_str
+
+#     @classmethod
+#     def from_ddbfile(cls, ddbfile: DdbFile) -> Self:
+#         """Create a DdbFileStr object from the native DdbFile abipy object."""
+#         with open(ddbfile.filepath) as f:
+#             ddb_as_str = f.read()
+#         return cls(ddbfilepath=ddbfile.filepath, ddb_as_str=ddb_as_str)
+
+
```
JaGeo, line 48: ?

src/atomate2/abinit/schemas/task.py
```diff
@@ -13,17 +13,97 @@
...
+        return cls(
+            basename=dct_abi_psp["basename"],
+            type_psp=dct_abi_psp["type"],
+            symbol=dct_abi_psp["symbol"],
+            Z=dct_abi_psp["Z"],
+            Z_val=dct_abi_psp["Z_val"],
+            l_max=dct_abi_psp["l_max"],
+            md5=dct_abi_psp["md5"],
+            filepath=dct_abi_psp["filepath"],
+            repo_name=repo_name,
```
JaGeo, line 102: I think the naming and the hypens are a bit inconsistent. "basename" vs. "repo_name".

JaGeo (commented):
I added some initial comments but would need to take a closer look at another point.

VicTrqt:
Hello Janine!
Thanks for your comments! I'll try addressing them as they come.

> There seem to be some failing abipy imports in the ase part of the tests. Could you check them? It might be that you should only import abipy when abipy is installed.

I'm not sure what you mean... Just in case, I modified the required abipy version (temporarily) so that the tests could pass with the github version. Seems that triggered something else with the pseudos so I'll fix that.

JaGeo:
@VicTrqt the tests never passed. Not even before the modification

VicTrqt:
@JaGeo Yes it was expected since new abipy factories are required and a version of abipy with those new factories has not been released yet.

> ## Note
>
>     * The new tests will fail until a new version of Abipy with my changes is released.

VicTrqt:
I was able to fix the tests (still temporarily relying on the github abipy). However, in test_dfpt, abinit is called to generate the perturbations. It could be mocked, but it would not be trivial in terms of work, especially considering future tests... so we decided to install abinit as a dependency so that the tests could pass online. This means that there is now a mixture of mock and real abinit calls. I also added a skip of the test_dfpt in case abinit was not in the PATH so that this test would not be triggered locally by developers without an abinit executable.

JaGeo:
@VicTrqt i have been wondering if there is interest to merge this soon?

VicTrqt:
Hi @JaGeo, no because thanks to @gpetretto and @emarazzi the ABINIT DFPT workflow has been improved to include a phonons workflow, so we are hoping to open a new PR soon (waiting for an abipy release) that will include this new generic DFPT workflow along with the specific SHG and Phonons workflows.

JaGeo:
@VicTrqt i see! Thanks!

esoteric-ephemera:
Can I close this out then @VicTrqt?

VicTrqt:
@esoteric-ephemera sure ! I can do it rn

PR #1062, opened 2024-11-22
Harmonic lattice dynamics workflow using hiPhive

hrushikesh-s:
## Summary

Lattice dynamics workflow using hiPhive, rebuilt on top of the pheasy workflow merged in #1063.

The two workflows are now identical except for the force-constant fit. Displacement generation, force assembly with residual subtraction, band structure, DOS, thermal displacements and the output document are shared code paths. Only the fit differs. The four `pheasy` subprocess calls are replaced by a hiPhive cluster space fitted with trainstation, with the Huang and Born-Huang rotational sum rules applied, which is what `--rasr BHH` did on the pheasy side.

The test reuses the `Si_pheasy` reference data rather than adding its own. Both workflows are therefore validated against identical DFT inputs, and the diff adds no test data.

Previous version of this PR was 114 files. This is 8.

New `hiphive` extra. hiPhive is already a dependency of the pheasy extra, so this pins the same version. `alamode` and the build toolchain are already installed in the `test-non-ase` CI job, so the workflow file changes by one word.

The `git+https://gitlab.com/jsyony37/hiphive.git@personal` dependency from the old version of this PR is gone. That fork existed to patch `hiphive/fitting/fit_methods.py`, a module that no longer exists in any current hiPhive. Fitting moved to the separate `trainstation` package, which depends on scikit-learn directly.

## Deferred

Anharmonic force constants, phonon renormalization, Grüneisen parameters, thermal expansion and lattice thermal conductivity are not in this PR. pheasy's `--fix_fc2` has no hiPhive equivalent, so the anharmonic fit needs a design decision that is better made on its own.

## Checklist

Work-in-progress pull requests are encouraged, but please put [WIP] in the pull request title.

  type check your code.


hrushikesh-s:
@janosh or @JaGeo , can you pls tell me how I can fix the following error -- 
`testing / test-non-ase (3.11, 1) (pull_request) Failing after 2m`

janosh:
`phono3py` can't currently be installed with `uv`/`pip`. you can upvote https://github.com/phonopy/phono3py/issues/295 which hopefully will make this possible soon 🤞

hrushikesh-s:
Oh I see. Thanks for sharing that info!
> `phono3py` can't currently be installed with `uv`/`pip`. you can upvote [phonopy/phono3py#295](https://github.com/phonopy/phono3py/issues/295) which hopefully will make this possible soon 🤞

janosh:
this looks very close to mergeable. anything i can do to help this over the finish line?

hrushikesh-s:
@JaGeo, this PR is ready for review.

I rebuilt it on top of the pheasy workflow from #1063. The two now share everything except the force constant fit. That took the diff from 114 files down to 8, and it adds no new test data since the test reuses `Si_pheasy`. The scope of this PR is harmonic only. 
Does this look okay to you? If so, can it be merged?

JaGeo:
@hrushikesh-s thanks! Did you run a slightly larger test to ensure stability of the workflow?

I will take a closer look in the next days

hrushikesh-s:
Thanks Janine. 

I ran a larger test as you suggested. I picked 14 materials from the Materials Project, two per crystal system. All of them already have pheasy phonons in the database. For each material, I compared three things. 
First, the DFT pheasy phonons from MP. 
Second, an MLIP phonopy run. 
Third, an MLIP hiPhive run. 
Both MLIP runs use MACE-OMAT-0-medium on the same relaxed cell, with the same supercell that DFT pheasy phonons on MP have used, and also used the same 0.01 A displacement. 
phonopy is the control. Where phonopy and hiPhive agree with each other, the fit is sound. What still separates both of them from DFT could be the error in MACE-OMAT-0-medium.

The test found four bugs in this PR. All four are fixed and pushed.

-- The pair cutoff could land on a neighbour shell, which hiPhive rejects. It now sits in the gap between shells.
-- The refit after imaginary modes reused the cutoff from the first fit, so it just repeated it. It now steps in by one neighbour shell.
-- A supercell too small to reach any neighbour shell now gives a clear error instead of a confusing one from hiPhive.
-- The default fit method was rfe. It does not scale past a few hundred parameters, and it is not what DFT pheasy phonon uses. The default is now lasso.

hiPhive agrees with phonopy on 10 of the 14. It disagrees on HfS2, NaAuO2, and K2ZnSi3O8. Ti2Ag fails because a 2x2x2 supercell of its body centred cell gives a cutoff that reaches no neighbour shell.

Everything else follows DFT pheasy phonons workflow -- same displacement count, same 0.01 A amplitude, same symprec.

Full numbers attached:
[hiphive_pr1062_validation.json](https://github.com/user-attachments/files/32299932/hiphive_pr1062_validation.json)

JaGeo:
Thank you! That's very helpful

Could you maybe add a hint to the documentation that the code is only tested on a smaller dataset so far?

hrushikesh-s:
Added the note about the limited testing of the hiphive workflow to the docs, and also added a tutorial notebook that reproduces the 14 material test set.

src/atomate2/common/flows/hiphive.py
```diff
@@ -0,0 +1,338 @@
...
+    """Maker to calculate harmonic phonons with the cluster-expansion code hiPhive.
+
+    Calculate the zero-K harmonic phonons of a material and higher-order FCs.
+    Initially, a tight structural relaxation is performed to obtain a structure
+    without forces on the atoms. Subsequently, supercells with all atoms displaced
+    by a small amplitude (generally using 0.01 A) are generated and accurate forces
+    are computed for these structures for the second order force constants. hiPhive
+    builds a cluster space from the supercell symmetry and fits the force constants
+    to those forces by regression. In this Workflow, we separate the harmonic phonon
+    calculations and anharmonic force constants calculations. To correct for
```
JaGeo: is the anharmonic part covered?

src/atomate2/common/flows/hiphive.py
```diff
@@ -0,0 +1,338 @@
...
+    polarization effects, a correction of the dynamical matrix based on BORN charges
+    can be performed. Finally, phonon densities of states, phonon band structures and
+    thermodynamic properties are computed. For the anharmonic force constants, the
+    supercells with all atoms displaced by a larger amplitude (generally using 0.08 A)
+    are generated and accurate forces are computed for these structures. The third-
+    and fourth-order force constants are fitted from the same cluster space.
```
JaGeo: I understood that this was actually not the scope of the PR. Did you test this?

src/atomate2/common/jobs/hiphive.py
```diff
@@ -0,0 +1,814 @@
...
+    positions = supercell_ph.scaled_positions
+    numbers = supercell_ph.numbers
+    natom = len(numbers)
+
+    # get the number of free parameters of 2ND FCs from ALM, labeled as n_fp
+    with ALM(lattice, positions, numbers) as alm:
+        alm.define(1)
+        alm.suggest()
+        n_fp = alm._get_number_of_irred_fc_elements(1)  # noqa: SLF001
+
```
JaGeo, line 367: This is also required for hiphive?

JaGeo:
@hrushikesh-s I believe this s nearly ready to merge. Could you clarify the documentation when it comes to higher order force constants?

hrushikesh-s:
@JaGeo, good catch on all three. These are fixed in the latest commit.

The docstring was copied from the pheasy maker and never trimmed. It described an anharmonic branch that does not exist here. Nothing generates 0.08 A displacements, and nothing fits third or fourth order, so there was nothing to test. The docstring now says harmonic only and that anharmonic is deferred.

On ALM, yes it is needed. It counts the irreducible second-order force constants for the supercell, and that sets how many displaced configurations to generate. Using ALM also keeps the displacement count the same as pheasy, which is what made the 14 material comparison meaningful. I added a comment saying this.

JaGeo:
@hrushikesh-s thanks. I will merge now. Great, no open phonon-related PRs!

One more thing: is our preprint on the Pheasy database already included somewhere in the documentation? If not, should we add it?

hrushikesh-s:
Thanks!

on citing the database preprint -- it's not in the docs anywhere yet, so yes, let us add it. I was thinking the Pheasy section of docs/user/codes/vasp.md, next to the pheasy code paper citation. Is that fine, or do you recommend adding it in some other location?

JaGeo:
Sounds good for now. Potentially we could start with using duecredit as we do in pymatgen.

JaGeo:
Examples:
https://github.com/materialsproject/pymatgen-core/blob/73af4e53f5f24e1dcf11e0d94ca13be10ea956ad/src/pymatgen/util/due.py
https://github.com/materialsproject/pymatgen-core/blob/73af4e53f5f24e1dcf11e0d94ca13be10ea956ad/src/pymatgen/io/lobster/inputs.py#L49

JaGeo:
Correction, we do already use it: https://github.com/materialsproject/atomate2/pull/1442/changes

PR #1063, opened 2024-11-22
Lattice dynamics workflow using Pheasy

hrushikesh-s:
## Summary

* Feature 1 -- Add lattice dynamics workflow using pheasy for harmonic & anharmonic phonon calculations

* pheasy – used to generate symmetry-aware cluster spaces, construct null-space constraints, build displacement/sensing matrices, and fit harmonic force constants from displacement–force data; also used for optional higher-order IFC fitting, phonon renormalization, and force-constant conversion workflows.
* alamode (alm) – used to estimate the number of irreducible force-constant parameters and thereby determine how many displaced supercells are needed for harmonic and anharmonic IFC extraction; also used to estimate higher-order anharmonic parameter counts.
* hiphive – used as a post-processing and recovery backend for force constants: it enforces rotational sum rules on second-order IFCs when imaginary modes are present, reconstructs corrected force constants, and is also used to convert third-order force constants to phono3py format for thermal-conductivity workflows.

  type check your code.


hrushikesh-s:
Pheasy workflow depends on [Alamode](https://alamode.readthedocs.io/en/latest/install.html). @janosh, can you pls help in installing Alamode via `.github/workflows/testing.yml` file?

janosh:
no idea, never used or installed Alamode before. the best solution would be for them to publish pre-compiled wheels to PyPI  so the code doesn't need to be re-compiled from scratch on every CI run as that's both slow and tends to be brittle

JaGeo:
@naik-aakash set up a docker for https://github.com/autoatml/autoplex but it has been a massive amount of work. Before this, we did the compilation on the repo. The slowness was a real problem. But not sure it is realistic to get the wheels.

janosh:
Could offer an issue bounty to get buy in. Happy to pitch 50$

JaGeo:
@janosh thank you for offering this.

I am unsure about the timescales for such solutions. We would like to have the workflow available as soom as possible due to the planned atomate2 paper.

hrushikesh-s:
In that case, @JaGeo & @naik-aakash can you pls help with the compilation of alamode on the repo?
Also tagging in @leslie-zheng.
> @naik-aakash set up a docker for https://github.com/autoatml/autoplex but it has been a massive amount of work. Before this, we did the compilation on the repo. The slowness was a real problem. But not sure it is realistic to get the wheels. To sum up: compilation on the repo might be a realistic and okay solution if it isn't too slow.

hrushikesh-s:
These are the exact commands to install Alamode:

```
1. conda install -c conda-forge numpy scipy h5py compilers “libblas=*=*mkl” spglib boost eigen cmake ipython mkl-include
2. git clone https://github.com/ttadano/ALM.git
3. cd ALM
4. python setup.py build
5. pip install -e .
```

JaGeo:
Just add it to the workflow file. You can see how micromamba is used for installation here:
https://github.com/materialsproject/atomate2/blob/3d55b2b56b2dfb938ab60b6deb198b8e682fcd7f/.github/workflows/testing.yml#L61

src/atomate2/common/flows/phonons.py
```diff
@@ -162,6 +162,7 @@ class BasePhononMaker(Maker, ABC):
 
     kpath_scheme: str = "seekpath"
     code: str = None
+    mp_id: str = None
```
JaGeo, line 165: This can be added as metadata.

tests/vasp/flows/test_phonons.py
```diff
@@ -202,6 +202,8 @@ def test_phonon_wf_vasp_only_displacements_no_structural_transformation(
     assert_allclose(
         responses[job.jobs[-1].uuid][1].output.primitive_matrix,
         ((0, 1, 0), (0, 0, 1), (1, 0, 0)),
+        rtol=1e-8,
```
JaGeo, line 205: @hrushikesh-s @naik-aakash and I are working on fixing these tests. I think I messed something up with the test files. I might have time later tonight. So, maybe skip those tests first.

JaGeo: Aah. Sorry! Different test 😅. But I meant the update to the latest phonopy. Sorry!

hrushikesh-s: oh okay okay. I was wondering why suddenly the test_phonons.py file was failing :sweat_smile:

hrushikesh-s:
@JaGeo , The test fails because GH can't find FORCECONSTANTS file in the temp folder. However, when I run the same test locally, I am able to successfully run it. Can you pls help?

JaGeo:
@hrushikesh-s I will try to look into it. Could you add all installation instructions to the documentation as well?

.github/workflows/testing.yml
```diff
@@ -67,8 +67,20 @@ jobs:
...
           uv pip install --no-deps nequip==0.5.6
+          
+
+      - name: Install ALA-Mode
+        run: |
+          micromamba activate a2
+          micromamba install -n a2 -c conda-forge numpy scipy h5py compilers “libblas=*=*mkl” spglib boost eigen cmake ipython mkl-include openmpi --yes  
+          git clone https://github.com/ttadano/ALM.git # do I need to modify this?
+          cd ALM # do I need to modify this?
+          cd python
```
JaGeo, line 81: don't you need the develop branch? I currently cannot install it locally

JaGeo:
@hrushikesh-s Locally, it fails for me with exactly the same error. Are you sure you have the same pheasy, ALM and phonopy versions? I only have a force_matrix.pkl file in the last folder but no FORCE_CONSTANT file

JaGeo:
I also got a numpy error. Something in the pheasy commands seems to go wrong.

hrushikesh-s:
> @hrushikesh-s I will try to look into it. Could you add all installation instructions to the documentation as well?
Yes. 

Doing this now

> @hrushikesh-s Locally, it fails for me with exactly the same error. Are you sure you have the same pheasy, ALM and phonopy versions? I only have a force_matrix.pkl file in the last folder but no FORCE_CONSTANT file

Checking this now

src/atomate2/common/jobs/pheasy.py
```diff
@@ -31,6 +34,40 @@
 
 logger = logging.getLogger(__name__)
 
+@job
```
JaGeo, line 37: What is the key difference here? Happy to modify the normal supercell search as well if there are weaknesses

hrushikesh-s: I wanted the get_supercell_size() to accept only min_length, max_atoms, force_90_degrees, force_diagonal, and allow_orthorhombic, but it currently also insists on a max_length argument.

That's why, as a quick fix, I created this new function. 😅

But if the original get_supercell_size() from jobs/phonons.py can be modified such that it doesn't insists on max_length argument, then that'll also work 😅

PS: For my current HT phonon runs (nsites ≥ 41) I specifically wanted:
min_length = 8.0
max_atoms = 200
force_90_degrees = True
force_diagonal = True
allow_orthorhombic = False

Tagging @leslie-zheng as well

JaGeo: I see. Thanks. Yes, maybe we should make the modification once this workflow is completely implemented and tested!

JaGeo:
@hrushikesh-s Thank you for this pull-request.
I have a list of TODOs:

- [x] Solve conflicts with the current main branch
- [x] Check if the tests run through here on github actions 
- [x] Add a real test for the whole pheasy workflow (ML potential and VASP) (only see schema tests so far)
- [ ] Add new documentation

Let me know if you need help with individual parts.

leslie-zheng:
> @hrushikesh-s Thank you for this pull-request. I have a list of TODOs:
> 
> * [ ]  Solve conflicts with the current main branch
> * [ ]  Check if the tests run through here on github actions
> * [ ]  Add a real test for pheasy (only see schema tests so far)
> * [ ]  Add new documentation
> 
> Let me know if you need help with individual parts.

Hi Janine, Could you please tell us more about the requirement of new document? @JaGeo 
thanks 
jiongzhi

JaGeo:
@leslie-zheng you mean documentation? Just some info on the approach as a part of the docs. Happy to help you write it once the conflicts are solved and a new test is there.

src/atomate2/common/flows/pheasy.py
```diff
@@ -0,0 +1,446 @@
...
+            Instead of min_length, also a supercell_matrix can be given, e.g.
+            [[1.0,0.0,0.0],[0.0,1.0,0.0],[0.0,0.0,1.0]
+        """
+        use_symmetrized_structure = self.use_symmetrized_structure
+        kpath_scheme = self.kpath_scheme
+        valid_structs = (None, "primitive", "conventional")
+        if use_symmetrized_structure not in valid_structs:
+            raise ValueError(
+                f"Invalid {use_symmetrized_structure=}, use one of {valid_structs}"
+            )
```
JaGeo, line 251: There is some code repetition here from the harmonic PhononMaker. I think we should either inherit from this Maker or move those code blocks into functions that can be easily reused.

src/atomate2/common/flows/pheasy.py
```diff
@@ -0,0 +1,446 @@
...
+            cal_anhar_fcs=self.cal_anhar_fcs,
+            displacement_anhar=self.displacement_anhar,
+            num_disp_anhar=self.num_disp_anhar,
+            fcs_cutoff_radius=self.fcs_cutoff_radius,
+            sym_reduce=self.sym_reduce,
+            symprec=self.symprec,
+            use_symmetrized_structure=self.use_symmetrized_structure,
+            kpath_scheme=self.kpath_scheme,
+            code=self.code,
+        )
```
JaGeo, line 362: This part is now different than in the harmonic phonon maker. Maybe, one could add a method to the phonon.BasePhononMaker and overwriting this here with the new implementation.

src/atomate2/common/flows/pheasy.py
```diff
@@ -0,0 +1,446 @@
...
+        # perform the phonon displacement calculations
+        displacement_calcs = run_phonon_displacements(
+            displacements=displacements.output,
+            structure=structure,
+            supercell_matrix=supercell_matrix,
+            phonon_maker=self.phonon_displacement_maker,
+            socket=self.socket,
+            prev_dir_argname=self.prev_calc_dir_argname,
+            prev_dir=prev_dir,
+        )
```
JaGeo, line 374: This is also different than in phonon.BasePhononMaker

src/atomate2/common/flows/pheasy.py
hrushikesh-s: Okay, makes sense. How should we do it? Should these code blocks be moved into a new function in common/jobs/phonons.py ?

hrushikesh-s:
> @hrushikesh-s Thank you for this pull-request. I have a list of TODOs:
> 
> * [x]  Solve conflicts with the current main branch
> * [x]  Check if the tests run through here on github actions
> * [x]  Add a real test for the whole pheasy workflow (ML potential and VASP) (only see schema tests so far)
> * [ ]  Add new documentation
> 
> Let me know if you need help with individual parts.

@JaGeo , can you pls check if the PR looks okay?

src/atomate2/common/flows/pheasy.py
hrushikesh-s: Didnt understand this. Do you mean writing a method in common/flows/phonons.py ?

src/atomate2/common/flows/pheasy.py
hrushikesh-s: This is currently getting imported like this:
```
from atomate2.common.jobs.pheasy import (
    generate_frequencies_eigenvectors,
    generate_phonon_displacements,
    get_supercell_size,
    run_phonon_displacements,
)
```

> This is also different than in phonon.BasePhononMaker

What should be done to fix this then?

src/atomate2/common/flows/pheasy.py
```diff
@@ -0,0 +1,446 @@
...
+            num_displaced_supercells=self.num_displaced_supercells,
+            cal_anhar_fcs=self.cal_anhar_fcs,
+            displacement_anhar=self.displacement_anhar,
+            num_disp_anhar=self.num_disp_anhar,
+            fcs_cutoff_radius=self.fcs_cutoff_radius,
+            renorm_phonon=self.renorm_phonon,
+            renorm_temp=self.renorm_temp,
+            cal_ther_cond=self.cal_ther_cond,
+            ther_cond_mesh=self.ther_cond_mesh,
+            ther_cond_temp=self.ther_cond_temp,
```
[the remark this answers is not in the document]
JaGeo: This would be one option but it might be a breaking change.

src/atomate2/common/flows/pheasy.py
JaGeo: I will draft something 😃

src/atomate2/common/flows/phonons.py
hrushikesh-s: Done!

.github/workflows/testing.yml
```diff
@@ -60,16 +60,28 @@ jobs:
...
+      - name: Install ALA-Mode
+        run: |
+          micromamba activate a2
+          micromamba install -n a2 -c conda-forge "numpy<=2.2" scipy h5py compilers “libblas=*=*mkl” spglib boost eigen cmake ipython mkl-include openmpi --yes  
+          git clone https://github.com/ttadano/ALM.git # do I need to modify this?
+          cd ALM # do I need to modify this?
+          cd python
+          python setup.py build # do I need to modify this?
+          uv pip install -e . # do I need to modify this?
+
```
JaGeo, line 72: Could you add these instructions to the documentation?

.github/workflows/testing.yml
```diff
@@ -60,16 +60,28 @@ jobs:
         run: |
           micromamba install -n a2 -c conda-forge enumlib packmol bader --yes
 
+      - name: Install ALA-Mode
+        run: |
+          micromamba activate a2
+          micromamba install -n a2 -c conda-forge "numpy<=2.2" scipy h5py compilers “libblas=*=*mkl” spglib boost eigen cmake ipython mkl-include openmpi --yes  
+          git clone https://github.com/ttadano/ALM.git # do I need to modify this?
```
JaGeo, line 67: https://anaconda.org/conda-forge/alm why not installing it directly from conda? Is a different version needed?

src/atomate2/vasp/flows/pheasy.py
```diff
@@ -0,0 +1,162 @@
...
+    Maker to calculate harmonic phonons with VASP and Phonopy.
+
+    Calculate the harmonic phonons of a material. Initially, a tight structural
+    relaxation is performed to obtain a structure without forces on the atoms.
+    Subsequently, supercells with one displaced atom are generated and accurate
+    forces are computed for these structures. With the help of phonopy, these
+    forces are then converted into a dynamical matrix. To correct for polarization
+    effects, a correction of the dynamical matrix based on BORN charges can
+    be performed. Finally, phonon densities of states, phonon band structures
+    and thermodynamic properties are computed.
```
JaGeo, line 31: Doc strings would need to be updated.

JaGeo:
@hrushikesh-s I increased code reuse in the pheasy code. The other changes already look great!

What is missing:
- please add instructions for the installation in the documentation, please also add a small section in the documentation on the pheasy workflow
- please check all doc strings (vasp.pheasy and forcefield.pheasy) and check if you added all required parameters
- please add a test for the forcefield implementation
- please switch to alm installation via conda-forge which should make the installation more stable
- get rid of code duplication in common.jobs.pheasy and common.jobs.phonons

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,475 @@
...
+        code to perform the computations
+
+    """
+    warnings.warn(
+        "Initial magnetic moments will not be considered for the determination "
+        "of the symmetry of the structure and thus will be removed now.",
+        stacklevel=1,
+    )
+    cell = get_phonopy_structure(
+        structure.remove_site_property(property_name="magmom")
```
JaGeo, line 129: ideally, we export the common parts from the phonons and pheasy jobs into a method and reuse it? This would then also make it easier when we change anything here in the future in either of the jobs in pheasy.py or phonons.py

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,475 @@
...
+@job(data=["forces", "displaced_structures"])
+def run_phonon_displacements(
+    displacements: list[Structure],
+    structure: Structure,
+    supercell_matrix: Matrix3D,
+    phonon_maker: BaseVaspMaker | ForceFieldStaticMaker | BaseAimsMaker = None,
+    prev_dir: str | Path = None,
+    prev_dir_argname: str = None,
+    socket: bool = False,
+) -> Flow:
```
JaGeo, line 390: if possible, i would modify the phonons.py job instead to avoid code duplication

JaGeo:
- [ ] We need to add the Pheasy reference to the documentation or the doc string of this PR: https://arxiv.org/abs/2508.01020

src/atomate2/common/jobs/pheasy.py
```diff
@@ -189,61 +177,41 @@ def generate_phonon_displacements(
...
+            ),
+            random_seed=random_seed,
+        )
 
     supercells = phonon.supercells_with_displacements
     displacements = [get_pmg_structure(cell) for cell in supercells]
 
-    # Here, the ALM module is used to determine how many free parameters of third and
-    # fourth order force constants (FCs) within the specific supercell.
+    # Here, the ALAMODE copde is used to determine the number of
```
JaGeo, line 213:
```suggestion
    # Here, the ALAMODE code is used to determine the number of
```

src/atomate2/common/jobs/pheasy.py
```diff
@@ -341,133 +306,608 @@ def generate_frequencies_eigenvectors(
...
-) -> Flow:
-    """
-    Run phonon displacements.
+    # phonon band structures will always be computed
+    filename_band_yaml = "phonon_band_structure.yaml"
+    filename_band_hdf5 = "phonon_band_structure.hdf5"
 
-    Note, this job will replace itself with N displacement calculations,
-    or a single socket calculation for all displacements.
+    # TODO: potentially add kwargs to avoid computation of eigenvectors
```
JaGeo, line 628:
```suggestion
```

JaGeo:
Linking this issue again https://github.com/materialsproject/atomate2/issues/1208

esoteric-ephemera:
@JaGeo  if you have a suggestion for how to implement the kind of check in #1208, feel free to push later today. Need to make some updates to the test data (there are absolute tolerances of 1000 used so the data is unreliable), don't unfortunately have the bandwidth to add a check like you're describing

JaGeo:
@esoteric-ephemera Sure. No worries. I just wanted to mention this again as a potential source for problems. This can be left out for now. (I also don't have the bandwidth)

esoteric-ephemera:
Also: one of the bigger changes in the phonon schema in going from atomate2 to emmet is to make the temperature, free energy, heat capacity, and entropy on-the-fly computed fields. The rationale was that these are really just computed from the phonon DOS, which we already store, and are fast to compute

I see that a lot of the phonon-derived flows (QHA + Gruneisen) require stored temperature + phonon DOS derived data, which also seems like a bad idea given that the default grid in the older PhononBSDOSDoc was very coarse (100 K steps)

I was able to get the Gruneisen flow working with just an additional job analysis step, for QHA I'll need to add another document model to store the thermal outputs, since the [original phonopy test data](https://github.com/phonopy/phonopy/tree/403f9b224da86cba500408d63a4a981da3248f69/example/Al-QHA) also omits DOSes

JaGeo:
@esoteric-ephemera Thanks for the feedback. We have added QHA and Grüneisen workflows quite recently and both have not been heavily tested in production yet. Thus, default values have not been optimized. I am also sorry that it requires a lot of effort from your side to integrate the document models from emmet into atomate2. However, at the time of development, the future developments for the phonon workflow were not yet clear.

esoteric-ephemera:
No worries! For now then, I'll go with making a slightly different phonon output data structure for QHA which saves some DOS-derived data, and then we can figure out better defaults later

JaGeo:
@esoteric-ephemera Sure and thanks!

hrushikesh-s:
Hi @JaGeo, does this look okay?

JaGeo:
@hrushikesh-s thanks!

I think one point we need to discuss is the deprecation of the phonon related schema. Our codes actually depend on the old schema and i think we should handle this carefully.

hrushikesh-s:
Hi @JaGeo , Thanks for bringing this up, that makes a lot of sense.

I agree that switching directly to the Emmet models could be disruptive since earlier production codes may still rely on the current atomate2 schema. Would it be reasonable to handle this in a phased way?

What I had in mind is:

* In this PR, we transition to using the Emmet models internally
* Keep the existing `atomate2.common.schemas.phonons` import path for backward compatibility
* Preserve any atomate2-specific helper methods so existing workflows don’t break
* Add a `DeprecationWarning` with a clear timeline

This way we can move toward the new schema without introducing a breaking change immediately. Does that sound like a reasonable approach to you? If yes, then I can make these edits to the current PR.

JaGeo:
I do recall a discussion involving @esoteric-ephemera that we wanted to potentially keep both options. 
I am in principle okay with a long-term deprecation. I would need to check the emmet models in detail to judge if they would still fully serve our purposes.

esoteric-ephemera:
I don't remember off hand how much the two schemas differ, can you point me to the fields that are needed in the legacy schema that are missing in the new emmet one @hrushikesh-s?

hrushikesh-s:
Hi @esoteric-ephemera ,

I compared the legacy [atomate2 phonon schema](https://github.com/materialsproject/atomate2/blob/main/src/atomate2/common/schemas/phonons.py) against the new [emmet](https://github.com/materialsproject/emmet/blob/main/emmet-core/emmet/core/phonon.py) one. The main differences are:

1. The new emmet schema replaces the old `PhononBSDOSDoc` with a `PhononBSDOSTask` + built `PhononBSDOSDoc` split.
2. `phonon_bandstructure` and `phonon_dos` are no longer stored as raw pymatgen objects; they use emmet wrapper schemas (`PhononBS`, `PhononDOS`).
3. The following legacy stored fields are not present as explicit stored fields in the new emmet schema:

   * `free_energies`
   * `heat_capacities`
   * `internal_energies`
   * `entropies`
   * `temperatures`
   * `uuids`
   * `jobdirs`
4. In emmet, thermodynamic quantities are computed from the DOS on demand rather than stored as arrays on the main BSDOS doc.
5. `has_imaginary_modes` is now a computed field rather than a plain stored field.
6. `phonopy_settings` was renamed to `post_process_settings`.
7. `force_constants` changed shape from the legacy wrapped `ForceConstants` object to a plain nested matrix list.
8. `total_dft_energy` changed convention: legacy atomate2 stored it per formula unit, while emmet stores it per atom.
9. The new schema also adds fields the legacy one did not have, such as `phonon_method`, `epsilon_electronic`, and sum-rule diagnostics (`sum_rules_breaking`, ASR/CNSR-related computed data).

Aside -- I tried to restore the legacy phonon schema functionality for the phonon, QHA, and Grüneisen workflows, and have fixed the associated test failures. Can you pls take a look and confirm whether these changes align with the goal of continuing to support the legacy schema for these workflows in the short term, with a plan to migrate to the emmet schema later?

hrushikesh-s:
Also, it's worthwhile to mention that the current PR makes use of [my fork of Pheasy repo](https://gitlab.com/hpsahasrabuddhe/pheasy/-/tree/support-disp-force-matrix-input), and the [PR is underway on Pheasy](https://gitlab.com/cplin/pheasy/-/merge_requests/1) to make it compatible with the current atomate2 style reading of .npy files for disp & force matrices

JaGeo:
Thanks!

From the timeline: i will likely look at it after Easter (next week earliest). I have to grade and prepare my teaching this week. I would also like to test and explore the changes in detail.

src/atomate2/common/jobs/gruneisen.py
```diff
@@ -50,7 +53,7 @@ def shrink_expand_structure(structure: Structure, perc_vol: float) -> Response:
     return Response(output={"plus": plus_struct, "minus": minus_struct})
 
 
-@job(data=[PhononBSDOSDoc])
```
JaGeo, line 53: With this new workflow, no object needs to be transferred to "data"?

src/atomate2/common/jobs/pheasy.py
JaGeo: can we merge this with the "get_supercell_size"? I think if it is just about the max_length, it must be possible to get this modified.

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,953 @@
...
+            filename=bs_plot_file,
+            units=kwargs.get("units", "THz"),
+        )
+
+        imaginary_modes = bs_symm_line.has_imaginary_freq(
+            tol=kwargs.get("tol_imaginary_modes", 1e-5)
+        )
+
+    # Using a shorter cutoff (10 A) to generate the force constants to
+    # eliminate the imaginary modes near Gamma point in phesay code
```
JaGeo:
```suggestion
    # eliminate the imaginary modes near Gamma point in pheasy code
```

src/atomate2/common/jobs/qha.py
```diff
@@ -56,7 +60,7 @@ def get_supercell_size(
     )
 
 
-@job(data=[PhononBSDOSDoc])
```
JaGeo, line 59: Is this not needed?

JaGeo: Or at least a replacement for this one?

tests/common/schemas/test_phonons.py
```diff
@@ -1,60 +0,0 @@
...
-from monty.json import MontyEncoder
-from pydantic import ValidationError
-
-from atomate2.common.schemas.phonons import (
-    PhononBSDOSDoc,
-    PhononComputationalSettings,
-    PhononJobDirs,
-    PhononUUIDs,
-    ThermalDisplacementData,
-)
```
JaGeo, line 14: I understood that we keep this for now, yes? Why not keeping the tests as long as we keep the schema?

src/atomate2/common/flows/pheasy.py
```diff
@@ -0,0 +1,362 @@
+"""Flow for calculating (an)harmonic FCs and phonon energy renorma. with pheasy."""
```
JaGeo, line 1:
```suggestion
"""Flow for calculating (an)harmonic FCs and phonon energy renormalisation with pheasy."""
```

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,953 @@
...
+
+    supercells = phonon.supercells_with_displacements
+    displacements = [get_pmg_structure(cell) for cell in supercells]
+
+    # Here, the ALAMODE code is used to determine the number of
+    # third and fourth-order FCs are needed for the supercell
+    if cal_anhar_fcs:
+        # Due to the cutoff radius of the force constants use the unit of Borh in ALM,
+        # we need to convert the cutoff radius from Angstrom to Bohr.
+        with ALM(lattice * 1.89, positions, numbers) as alm:
```
JaGeo, line 241: I assume this is accurate enough for a cutoff but maybe more digits?

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,953 @@
...
+        # phonon band structures will always be computed
+        phonon.run_band_structure(
+            qpoints, path_connections=connections, with_eigenvectors=True
+        )
+        phonon.write_yaml_band_structure(filename=_DEFAULT_FILE_PATHS["band_structure"])
+        bs_symm_line = get_ph_bs_symm_line(
+            _DEFAULT_FILE_PATHS["band_structure"],
+            labels_dict=kpath_dict,
+            has_nac=born is not None,
+        )
```
JaGeo, line 829: can more code be reused from phonon.py? If we have an update for the plotting, we would need to fix it here as well. Same for the code below.

src/atomate2/openff/__init__.py
```diff
@@ -1,5 +1,13 @@
 """Module for classical md workflows."""
 
+from importlib.util import find_spec
+
+if not find_spec("openff"):
+    raise ImportError(
+        "openff must be installed via conda forge to use with atomate2:\n"
+        "conda install -c conda-force openff"
+    )
+
```
JaGeo, line 10: can you check if this is consistent with how we handle this accross atomate2?

tests/vasp/flows/test_pheasy.py
```diff
@@ -0,0 +1,130 @@
...
+    assert ph_doc.code == "vasp"
+    assert isinstance(
+        ph_doc.post_process_settings,
+        PhononComputationalSettings,
+    )
+    assert ph_doc.post_process_settings.npoints_band == 101
+    assert ph_doc.post_process_settings.kpath_scheme == "seekpath"
+    assert ph_doc.post_process_settings.kpoint_density_dos == 7000
+
+    assert ph_doc.chemsys == "Si"
```
JaGeo, line 130: I don't see a test that actually checks the new implemented functionality. How do we make sure that the results stay consistent?

JaGeo:
Hi  @hrushikesh-s ,

I left a few comments.

My main points:

- [ ] pheasy functionality must be tested, e.g., by testing some of the phonon results.
- [ ] could you check that both Grüneisen and QHA workflow still work in practice when they are run with fireworks or jobflow-remote. The data objects are a really important point
- [ ] should we provide a forcefield implementation for testing?
- [ ] should we provide a tutorial for pheasy? if we had a forcefield implementation, this would be much easier to do.
- [ ] conflicts need to be solved :)

esoteric-ephemera:
Hey @hrushikesh-s sorry I missed your messages. These schema changes are very much intentional:

> 1. The new emmet schema replaces the old PhononBSDOSDoc with a PhononBSDOSTask + built PhononBSDOSDoc split.

This is to explicitly allow for storing these as separate objects. `PhononBSDOSTask` should go to regular mongo collections, whereas band structures and DOS should go to GridFS

> 2. phonon_bandstructure and phonon_dos are no longer stored as raw pymatgen objects; they use emmet wrapper schemas (PhononBS, PhononDOS).

There is an [opt-in setting](https://github.com/materialsproject/emmet/blob/85b7edc3e1a8b4e49d7cc36ec4a7b7465bacc0dd/emmet-core/emmet/core/settings.py#L176) for using the new emmet models. The default is still to use the pymatgen models in atomate2 (opt-out).

> 3. The following legacy stored fields are not present as explicit stored fields in the new emmet schema:

    free_energies
    heat_capacities
    internal_energies
    entropies
    temperatures
    uuids
    jobdirs

There is no reason to store `free_energies`, `heat_capacities`, `internal_energies`, `entropies`, and `temperatures` in the document, since these are computed from a set of temperatures. What if the user wants a different set of temps than the one prescribed in the document?

The `uuids` and `jobdirs` are stored in a [`calc_meta` field](https://github.com/materialsproject/emmet/blob/85b7edc3e1a8b4e49d7cc36ec4a7b7465bacc0dd/emmet-core/emmet/core/phonon.py#L403)

> 5. has_imaginary_modes is now a computed field rather than a plain stored field.

A model computed field will still be saved in the document if it has been calculated. It is just a way to [defer calculating an expensive model field](https://docs.pydantic.dev/latest/concepts/fields/#the-computed_field-decorator) until the user requests it

Regarding the other fields being renamed, the document model explicitly handles the migration internally. You don't have to do anything to coerce the fields. Put a decent amount of work into it to do that since it needed to migrate our existing data seamlessly too

Please revert these changes regarding the schema or communicate what changes you need in the schema to get the workflows to run correctly

hrushikesh-s:
Hi @JaGeo and @esoteric-ephemera, thanks for your detailed comments. I will start working on the comments this Friday and will keep you updated.

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,953 @@
...
+    logger.info("Start running pheasy in cluster")
+
+    subprocess.call(shlex.split(pheasy_cmd_1))
+    subprocess.call(shlex.split(pheasy_cmd_2))
+    subprocess.call(shlex.split(pheasy_cmd_3))
+    subprocess.call(shlex.split(pheasy_cmd_4))
+
+    # When this code is run on Github tests, it is failing because it is
+    # not able to find the FORCE_CONSTANTS file. This is because the file is
+    # somehow getting generated in some temp directory. Can you fix the bug?
```
JaGeo, line 548: i guess this works now as well

JaGeo:
@hrushikesh-s do you have a suggestion for the timeline? More and more conflicts start to show up and it will only get harder to finally merge

tests/forcefields/flows/test_phonon.py
```diff
@@ -216,47 +217,56 @@ def test_phonon_wf_force_field(
...
+        "internal_energy": [
+            5058.44158791,
+            5385.88058579,
+            6765.19854165,
+            8723.78588089,
+            10919.0199409,
+        ],
         "free_energy": (
-            [11500.0, 11000.0, 9100.0, 6100.0, 2200.0]
+            [3164.0, 3053.0, 2351.0, 999.0, -868.0]
```
JaGeo, line 244: This changed so much as Chgnet was updated, right?

hrushikesh-s: Yes, right. I think CHGNet changed from the old MPtrj-trained model to the new MatPES-trained model that matgl now serves.

src/atomate2/common/jobs/gruneisen.py
JaGeo, line 53: Are you sure this should be removed?

src/atomate2/common/jobs/pheasy.py
```diff
@@ -0,0 +1,916 @@
...
+
+logger = logging.getLogger(__name__)
+
+try:
+    from alm import ALM
+except ImportError:
+    ALM = None
+
+# CODATA 2018: 1 Angstrom = 1 / 0.529177210903 Bohr
+ANGSTROM_TO_BOHR = 1.8897261246257702
```
JaGeo, line 51: Are you using this at multiple places? If you, please make a module for it. Eg "utils.units.py"

src/atomate2/common/jobs/gruneisen.py
JaGeo: @hrushikesh-s i believe this is the very last open point

JaGeo:
@hrushikesh-s can you please resolve the very last open point? I am happy to merge if this is done

JaGeo:
@hrushikesh-s are you happy with it? can i merge? :)

hrushikesh-s:
Hi @JaGeo, yes, pls merge it 😅

JaGeo:
Done and massive thank you @hrushikesh-s !!

PR #1079, opened 2024-12-07
Add MatterSim, Allegro and OCP models (`fairchem-core`) to `ase_calculators`

janosh:
MatterSim does very well on the new thermal conductivity modeling task ($\kappa_\rm{SRME} = 0.554$, see https://github.com/MPA2suite/k_SRME/pull/12, thanks @AntObi), almost as good as the current SOTA GRACE ($\kappa_\rm{SRME} = 0.525$, see https://matbench-discovery.materialsproject.org). 

the small (1M) model interestingly does better than the large (5M) one, which is a good sign that $\kappa_\rm{SRME}$ could prove more robust to overfitting

Screenshot 2024-12-07 at 14 28 05

either way, MatterSim seems like a useful addition to the set of supported ASE calculators in `atomate2`, hence this PR.

adding `Allegro` and `OCPCalculator` as well since I've been using them with my local fork for a while

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ forcefields = [
     "matgl>=1.1.3",
     "torchdata<=0.7.1",
     # quippy-ase support for py3.12 tracked in https://github.com/libAtoms/QUIP/issues/645
+    "mattersim>=1.0.0rc10.dev1",
     "quippy-ase>=0.9.14; python_version < '3.12'",
     "sevenn>=0.9.3",
     "torchdata<=0.7.1",                            # TODO: remove when issue fixed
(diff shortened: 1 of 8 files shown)


gpetretto:
Sorry, I think it would be convenient to have these additional potentials among the easily accessible options. (maybe also including the latest UMA from Meta would be good). Is there any plan to merge this?

esoteric-ephemera:
@gpetretto and @JaGeo I'm in favor of including these - right now, we have dependency conflicts between `mattersim` and `mace_torch` that make it infeasible to test both in CI

In the longer term, I think we will need separate test blocks for different sets of MLFFs with compatible dependencies. Since we are not testing Allegro nor fairchem, I think it's also safe to avoid testing mattersim for now / skip any test if it's not installed (the default CI behavior)

JaGeo:
@esoteric-ephemera fair for now. 

Definitely, we need to come up with a testing strategy here in the future.

esoteric-ephemera:
I'll post an issue for the missing tests after, I don't want to let it linger for too long

JaGeo:
@esoteric-ephemera sure! And thank you!

PR #1095, opened 2025-01-04
Bump monty and pymatgen 

naik-aakash:
# Changes

Upgrade monty to latest version

# Reason 

See https://github.com/materialsproject/pymatgen/issues/4243, due to accidental oversight with deprecation warning raise behaviour in monty, all downstream packages where an earlier version of monty is used, CI tests fails

PR #1106, opened 2025-01-21
Fixed Failing Abinit tests

jmmshn:
## Summary

The abinit tests are failing due to a float point comparison that gave the following error:
["The variable 'tsmear' is different in the two files:\n - this file: '0.0036749322175665 Ha'\n - other file: '0.0036749322175655 Ha'"]

This is being patched upstream 
https://github.com/abinit/abipy/commit/dc6d09766da5ffc9f1e5a5d9b9ad1688c0c28563

but I have added bypass here for now.

There is a `_get_differences_tol` function that mimic what the updated function is suppose to do but implemented differently to avoid other updates in `abipy`.
Once the `abipy` is bumped we can remove this helper function in the tests.

Also changed AIMS EOS test to allow for a `1e-4` tolerance.

### DIFF
--- a/tests/abinit/conftest.py
+++ b/tests/abinit/conftest.py
@@ -130,7 +130,7 @@ def check_run_abi(ref_path: str | Path):
         ref_str = file.read()
     ref = AbinitInputFile.from_string(ref_str.decode("utf-8"))
     # Ignore the pseudos as the directory depends on the pseudo root directory
-    diffs = user.get_differences(ref, ignore_vars=["pseudos"])
+    diffs = _get_differences_tol(user, ref, ignore_vars=["pseudos"])
     # TODO: should we still add some check on the pseudos here ?
     assert diffs == [], f"'run.abi' is different from reference.\n{diffs}"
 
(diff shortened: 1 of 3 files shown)


JaGeo:
@jmmshn Thank you for working on this. Could you build on the discussion here? https://github.com/materialsproject/atomate2/pull/1095

jmmshn:
@JaGeo , my thinking (feel free to push back) is that see any direct usage of `all_close` or any kind of fractional coordinate checking has an implicit dependence on the atom order which again relies on comparison of exact floating point values.

If we want to "fuzzy" check structure we need some kind of actual fingerprinting --- short of that I think composition + volume will have to be good enough for now.

JaGeo:
@jmmshn I understand. However, I think this would leave the structure generation untested which I find dangerous.

I think we already have a solution for the problem here.  https://github.com/materialsproject/atomate2/blob/95ea0600e00bcded2a0cb6cfe6d190fa6a980c39/src/atomate2/utils/testing/vasp.py#L237

jmmshn:
> I think we already have a solution for the problem here.

Ooo this works I was not aware of this.

JaGeo:
> > I think we already have a solution for the problem here.
> 
> Ooo this works I was not aware of this.

So far, there hasn't been an indication for me that it doesn't 😅. In any case safer than not comparing at all.

JaGeo:
@jmmshn could you maybe extract the comparison function for structures to a common testing module, import it from there to vasp.py and the abinit tests? Then, I would be happy to merge.

VicTrqt:
Hello @JaGeo and @jmmshn , 
Thanks a lot for looking into this as well. 
My fix does not account for a shuffling of the atom order and it is a good idea to include it indeed. After discussing with @gpetretto , we don't see any reason not to adapt and use ```check_poscar```. We also think it would be better to keep my check on the dict ```abivars```, except for the key ```xred``` (the fractional coordinates). Since we are allowing a shuffling of the atoms, it also needs to be adapted to allow a shuffling of ```znucl``` (which element) and ```typat``` (the ordering of the elements in the coordinates). As this is more ABINIT related, I will modify it and then you can include it in this PR if this is okay with you (or I will open a new one depending on the time).
Regarding the fix in Abipy mentioned at the beginning [abinit/abipy@dc6d097](https://github.com/abinit/abipy/commit/dc6d09766da5ffc9f1e5a5d9b9ad1688c0c28563) for the floating point comparison, it would be nice to avoid too many version releases since, depending on the review, my open SHG PR might require modifications in Abipy as well...

VicTrqt:
- I adapted check_poscar [here](https://github.com/VicTrqt/atomate2/blob/65b657b940e0e102abe8b2b5ad5f01a79d17fcd6/tests/abinit/conftest.py#L163). Feel free to modify it and put it in the common testing module as suggested.
- I implemented a small check on the type of elements and their number of atoms [here](https://github.com/VicTrqt/atomate2/blob/65b657b940e0e102abe8b2b5ad5f01a79d17fcd6/tests/abinit/conftest.py#L184)
- So the final ABINIT check looks like [this](https://github.com/VicTrqt/atomate2/blob/65b657b940e0e102abe8b2b5ad5f01a79d17fcd6/tests/abinit/conftest.py#L210). I can open a new PR for it if needed, but it might as well be part of this PR by copying it.

JaGeo:
@VicTrqt maybe open a pull request and i will merge this. Then, this is fixed.

VicTrqt:
Here it is #1108 .
I also copied the rest of the fixes implemented by @jmmshn.

JaGeo:
Closing this as it is fixed in #1108

JaGeo:
Thank you both @jmmshn and @VicTrqt !!

jmmshn:
Thanks @VicTrqt, @JaGeo I couldn't get to it yesterday but thanks for pushing this through the finish line!

JaGeo:
@jmmshn thanks and don't worry! I also only was able to review and not do it myself 😉

PR #1110, opened 2025-01-25
Added a MlffVaspLobsterMaker

QuantumChemist:
## Summary

As a lot of MLIPs can reproduce important structure characteristics at DFT accuracy, I have added a `MlffVaspLobsterMaker`, that replaces the `BaseVASPMaker` with `ForceFieldRelaxMaker` for the `relax_maker` to enable the user to speed up the calculations.

I could have also simply made:
```python
@dataclass
class VaspLobsterMaker(Maker):
    name: str = "lobster"
    relax_maker: BaseVaspMaker | ForceFieldRelaxMaker | None = field(
        default_factory=lambda: DoubleRelaxMaker.from_relax_maker(RelaxMaker())
    )
```
but I wanted to make the optimization step mandatory in this Maker and it would not have been really reflected in the name. Let me know if you prefer to simply add the `ForceFieldRelaxMaker` type to `relax_maker`.

Anyway, this will be a very convenient way to do similar calculations as I have done for [Figure 4](https://pubs.acs.org/doi/abs/10.1021/acs.chemmater.1c03349).

## TODO (if any)

Maybe add a unit test?

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -89,6 +89,7 @@ jobs:
           micromamba activate a2
           pytest --splits 3 --group ${{ matrix.split }} --durations-path tests/.pytest-split-durations --splitting-algorithm least_duration --ignore=tests/ase --cov=atomate2 --cov-report=xml
 
+
       - uses: codecov/codecov-action@v1
         if: matrix.python-version == '3.10' && github.repository == 'materialsproject/atomate2'
         with:
@@ -150,7 +151,7 @@ jobs:
       - name: Test Notebooks
         run: |
           micromamba activate a2
-          pytest --nbmake ./tutorials --ignore=./tutorials/openmm_tutorial.ipynb
+          pytest --nbmake ./tutorials --ignore=./tutorials/openmm_tutorial.ipynb --ignore=./tutorials/force_fields
 
       - name: Test ASE
         env:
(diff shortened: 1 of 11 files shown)


JaGeo:
@QuantumChemist thankyou! I think this might be something for the MP compatible Lobster workflow. Now, PBE in the ML potential (at least in the universal ones) is combined with PBEsol.

However, I would prefer allowing the ML potential in the workflow as there are so many options to combine ML potentials with DFT that we likely cannot implement all of them.

What I think would be really useful: a combined relaxmaker that uses the ML potential first and then runs a MP compatible relaxmaker. This could then also be combined with our workflow and many others. Maybe, you could implement this?

We could also add the the speed-up due to the non-scf run to the MP compatible workflow.

JaGeo:
@QuantumChemist Thinking about it further: this PR could be the start of a common workflow for Lobster. While it now maybe only allows VASP/MLIP in the relaxation, it could be more open in the future, e.g. in connection with abinit. 😀


Let me know what you think and what you prefer.

QuantumChemist:
@JaGeo I like the idea of making it a common LOBSTER wf. 😃 
Converted the PR to a draft for now. Will work on it whenever I have time on the weekends or so.
Wanted to draft the idea with my first commit.

Issue #1102, opened 2025-01-13
atomate2 / OpenMM Enhancements: Added OPLS Support


shehan807:
In the coming week, I plan to implement enhancements to the atomate2 / OpenMM features specifically through support for OPLS-based high-throughput pipelines. Any feedback or concern for overlap with existing efforts would be greatly appreciated: 

#### 1. Extend `download_opls_xml` to using BOSS v5.1 via Docker
Details mentioned [here](https://github.com/materialsproject/atomate2/pull/1054#issuecomment-2498477918), to use BOSS v5.1 one requires a local install of the executable provided by the Jorgensen group. Given the number of dependencies to run the program (i.e., 32-bit compatible libraries), Docker images are suitable. I would defer to anyone who is more familiar with the licensing aspect of distributing code (here is the BOSS license: [NonProfitLicense.pdf](https://github.com/user-attachments/files/18399352/NonProfitLicense.pdf)), but having a self-contained environment of ligpargen would seem easier than requiring users setting it up locally. 

#### 2. Implement Lennard-Jones OPLS combination rules via a Nonbonded flag
The CustomNonbondedForce in OpenMM implements the Lorentz-Berthelot combination rules to estimate sigma and epsilon parameters for 1-2, 1-3, and 1-4 interactions. However, correctly implementing the OPLS-AA geometric combination rules requires modification of the OpenMM system.

dspoel:
I am just googling for OPLS combination rules in OpenMM. Wasn't that already implemented in OpenMM?
https://github.com/openmm/openmm/issues/1918

shehan807:
The `NonbondedForce` class still does not support alternate combing rules (https://github.com/openmm/openmm/issues/4384#issuecomment-1881491329) nor does LigParGen create the necessary `CustomNonbondedForce` class when you generate .xml files. Instead, this is done via `opls_lj()`, as recommended by the [LigParGen](https://traken.chem.yale.edu/ligpargen/openMM_tutorial.html) website and now implemented in PR #1111.

PR #1111, opened 2025-01-25
atomate2 / OpenMM OPLS-AA Enhancements

shehan807:
## Summary

This PR resolves the first two enhancements proposed in Issue #1102: 

* **Feature 1**: Implements the correct geometric combination rules for OPLS MD simulations via [`opls_lj()`](https://github.com/shehan807/atomate2/blob/b45af33739649813997abe52336b00368407d75d/src/atomate2/openmm/utils.py#L261)

I've changed `create_system_from_xml` to `create_ff_from_xml` in `src/atomate2/openmm/jobs/generate.py`. Why? The OPLS geometric combination rules have two requirements, (1) a `CustomNonbondedForce` in OpenMM that combines the `sigma` and `eps` parameters geometrically, and (2) the `coulomb14scale` and `lj14scale` attributes of the `ForceField` are **both** `0.5`. The latter was enough justification for raising an error if not the case--to do so, the `ForceField` object would need to be accessible in `generate_openmm_interchange` (instead of adding unsuitable arguments to `create_system_from_xml`). Originally, `create_system_from_xml` created both the `ForceField` and `System` within the same function; now, `system` and `ff` are separate variables in `generate_openmm_interchange`. 

* **Feature 2**: Allows users to create OPLS forcefield files via [`generate_opls_xml()`](https://github.com/shehan807/atomate2/blob/b45af33739649813997abe52336b00368407d75d/src/atomate2/openmm/utils.py#L135) via LigParGen and BOSS source code directly. To do so, users would need to follow the steps in the [docs](https://github.com/shehan807/atomate2/blob/b45af33739649813997abe52336b00368407d75d/docs/user/codes/openmm.md?plain=1#L449) from this fork. 

Currently, `generate_opls_xml` is only functional so long as the user has a local installation of Docker and created a *private* (~~for now, pending approval for pushing to public registries~~ approval received to save on public registries) image of LigParGen. The documentation is provided to completion in `docs/user/codes/openmm.md`. 

EDIT: We've received written approval from Dr. William Jorgensen to host the LigParGen image on a public registry, i.e., the user no longer has to build this themself. 

* **Feature 3**: All changes from #1054 have been merged into this fork (i.e., extending the `download_opls_xml` to handle charges, number of optimization iterations, and charge method).

* ~~[defusedxml](https://pypi.org/project/defusedxml/), resolves ruff failures for safely loading XML files, e.g.,~~ 

* ~~[textwrap](https://docs.python.org/3/library/textwrap.html), enables easier text wrapping of long warnings/errors printed for the user, e.g.,~~

## Checklist
  type check your code.
 
CC'ing @utf, @orionarcher


src/atomate2/openmm/utils.py
```diff
@@ -23,27 +32,35 @@
 
 
 def download_opls_xml(
-    names_smiles: dict[str, str], output_dir: str | Path, overwrite_files: bool = False
+    names_params: dict[str, dict[str, str]],
```
orionarcher, line 35: can you document what params are allowed?

orionarcher:
@janosh or @utf could you get the tests running here?

src/atomate2/openmm/jobs/generate.py
```diff
@@ -5,25 +5,29 @@
 import copy
 import io
 import re
+import textwrap
 import warnings
 import xml.etree.ElementTree as ET
 from pathlib import Path
 from xml.etree.ElementTree import tostring
 
+import defusedxml.ElementTree as DiffET
```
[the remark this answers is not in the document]
shehan807: ruff gave me the following error for `ET.parse`,

```bash
S314 Using `xml` to parse untrusted data is known to be vulnerable to XML attacks; use `defusedxml` equivalents
```

I now see that `# noqa: S314` was there in the original version to suppress this error, so I've updated the try/except statement, now without defusedxml.

src/atomate2/openmm/utils.py
shehan807: Done.

shehan807:
@janosh all pre-commit/pytests are working on my end as of my last commit, could you try running the tests again? Thanks @orionarcher for the comments!

shehan807:
@janosh The last failed test is due to an environment variable dependence introduced by the `generate_opls_xml` function. The function requires the user to follow the [OPLS docs](https://github.com/shehan807/atomate2/blob/feature/OPLS-AA/docs/user/codes/openmm.md#varying-forcefields-opls) to create a `CONTAINER_SOFTWARE` and `LPG_IMAGE_NAME` environment variable:

```bash
export LPG_IMAGE_NAME="USERNAME/ligpargen:latest"
export CONTAINER_SOFTWARE="podman-hpc" # e.g.
```

Based on what I've observed from the VASP/AIMS tests, I just updated `test_generate_opls_xml` so it [monkeypatches ](https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables) environment variables. The new tests are now broken up into `test_generate_opls_xml_no_container` and `test_generate_opls_xml_success`. The latter is skipped since, even if the env variables were set, it would fail because success depends on whether podman-hpc or docker are available.

orionarcher:
@janosh or @utf the tests are frozen again here, could we get them running here?
 
 Sorry to bother, I'm not sure why another approval is required.

orionarcher (approved):
Thanks for addressing my comments, this looks good on my end!

PR #1118, opened 2025-02-10
Update README.md and check current test failures

JaGeo:
Check where current test failures come from.


JaGeo:
@Andrew-S-Rosen i think we solved it. They changed the default model to the alexandria one in mace. I will close this here to avoid confusion.

PR #1122, opened 2025-02-12
Tutorials for QHA and Grüneisen workflows

JaGeo:
## Summary

- [x] Include tutorials for qha
- [x] include tutorial for grüneisen parameter
- [x] fix potential omissions in the qha schema 
- [x] finish my fight with the linting (our hard drive looses all file permissions which makes the pre-commit really messy...)


JaGeo:
Note here: i had to exclude one file from ruff as ruff-format would do the exact opposite.

janosh:
always good to see more docs and tutorials! 👍 

curious though, is it really necessary to commit ~400 test files alongside these tutorials? seems like this could bloat the repo over time

JaGeo:
I think i had a look at the data size: it was not more than 4 MB

JaGeo:
@janosh To add further, i think this is currently one of our most complex workflows. We can, unfortunately, not run it with most of the forcefields as the accuracy is not good enough yet. I might be able to reduce the number of points along the ev curve a bit to get still a good fit but that would mean running yet another workflow just for test data. Not sure it is worth the effort in this one case. 

Btw, we now have: https://materialsproject.github.io/atomate2/tutorials/tutorials.html

Issue #1123, opened 2025-02-14
BUG:ForceFieldRelaxMaker does not compatible with molecule


yaoyi92:
**Describe the bug**
ForceFieldRelaxMaker does not compatible with molecule. Is there any effort in making them compatible with each other?

**To Reproduce**
The code to generate the problem:
```
from atomate2.forcefields.jobs import ForceFieldRelaxMaker
from pymatgen.core import Molecule
from jobflow import run_locally
from atomate2.forcefields import MLFF

molecule = Molecule.from_str(
"""3
Properties=species:S:1:pos:R:3 pbc="F F F"
O 0.0 0.0 0.0
H 1.0 0.0 0.0
H 0.0 1.0 0.0
""",
fmt="xyz",
)
job_relax = ForceFieldRelaxMaker(
    force_field_name = MLFF.MACE,
    relax_kwargs = {'fmax':0.025}).make(molecule)
response_relax = run_locally(job_relax)
```
And here is the result.

```
2025-02-14 15:13:45,456 INFO Started executing jobs locally
2025-02-14 15:13:45,463 INFO Starting job - Force field relax (bb32706c-9a5c-4f38-986f-141ccfd08f64)
/home/yiy/Test/000-active-learning-workflow/test_env/mace_env/lib/python3.10/site-packages/e3nn/o3/_wigner.py:10: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  _Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
cuequivariance or cuequivariance_torch is not available. Cuequivariance acceleration will be disabled.
Using medium MPA-0 model as default MACE-MP model, to use previous (before 3.10) default model please specify 'medium' as model argument
[...]

PR #1127, opened 2025-02-18
[WIP]: Molecule for forcefield

yaoyi92:
## Summary

Discussed in #1123. Following the same strategy as ase tasks, output ForceFieldStructureTaskDocument or ForceFieldMoleculeTaskDocument based on the input type of mol_or_struct. 

## Checklist

title.

* [] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org) to
  type check your code.


JaGeo:
Thank you! 

I am tagging @utf and @esoteric-ephemera as well as we have to decide how to handle molecular and structural outputs in general.

yaoyi92:
Here is my understanding of how to handle molecular and structural outputs. I saw in atomate2, two approaches exist. In FHI-aims, and cp2k engines, their taskdoc inherents from both StructureMetadata, MoleculeMetadata and the output structure takes the same type of the input structure. While, in ase engine, the implementation is to output StructureTaskDoc or MoleculeTaskDoc based on the input structure. Since, the forcefield inherents mostly from ase, I picked the second approach. The FHI-aims/cp2k approach is  cleaner in software engineering. However, StructureMetadata and MoleculeMetadata seems to have different symmetry types. I am not sure whether a conflict will happen is symmetry is used. My understanding is to make them fully compatible with each other would require a rework at the emmet-core level.

esoteric-ephemera:
Hey @yaoyi92, thanks for making this extension to the forcefields! The split that occurs for `StructureMetadata` and `MoleculeMetadata`-based task documents in ASE is to ensure there are no collisions between attributes with the same names.

I would avoid merging the `ememt-core` classes - we're already in the process of reworking the `TaskDoc` and other schemas in `emmet-core` to enforce stricter typing

To @JaGeo's point: we have too many disparate schemas for each code, and it would likely be best to have them originate with the same base document models. Maybe we should start an issue for this?

JaGeo:
@esoteric-ephemera I agree with your points and I also believe we should think about the task docs more broadly.

src/atomate2/forcefields/md.py
```diff
@@ -126,19 +130,18 @@ def __post_init__(self) -> None:
...
         data=[*_FORCEFIELD_DATA_OBJECTS, "ionic_steps"],
-        output_schema=ForceFieldTaskDocument,
     )
     def make(
         self,
-        structure: Structure,
+        structure: Molecule | Structure,
         prev_dir: str | Path | None = None,
-    ) -> ForceFieldTaskDocument:
+    ) -> ForceFieldStructureTaskDocument | ForceFieldMoleculeTaskDocument:
```
JaGeo, line 138: I am not sure why this isn't done for all other ForceFieldMakers? There, I only see a replacement with `ForceFieldStructureTaskDocument`.

src/atomate2/ase/schemas.py
```diff
@@ -233,27 +233,6 @@ class AseStructureTaskDoc(StructureMetadata):
...
-            Additional keyword args passed to :obj:`.AseStructureTaskDoc()`.
-        """
-        task_document_kwargs.update(
-            {k: getattr(ase_task_doc, k) for k in _task_doc_translation_keys},
-            structure=ase_task_doc.mol_or_struct,
-        )
-        return cls.from_structure(
-            meta_structure=ase_task_doc.mol_or_struct, **task_document_kwargs
-        )
-
```
JaGeo, line 256: This is not needed anymore?

yaoyi92: This function is only called from the forcefields schemas.
I moved this function to [src/atomate2/forcefields/schemas.py](https://github.com/materialsproject/atomate2/pull/1127/files#diff-bcdf417f8360fa7de45fdc429f55fc77042f2e18200fd89125da1ca619777abe).

JaGeo: I see. However, it is a general static method. Why not keeping it here?

src/atomate2/forcefields/md.py
yaoyi92: Which makers are you referring to? I only found two makers that have an explicit make function, ForceFieldRelaxMaker and ForceFieldMDMaker. It seems, the other makers inherent from these two.

src/atomate2/ase/schemas.py
yaoyi92: I have add the method back in this commit. 7b321660f1956b9d58807f7c264a90c1396f5ed2

src/atomate2/forcefields/jobs.py
```diff
@@ -118,7 +121,7 @@ class ForceFieldRelaxMaker(AseRelaxMaker):
     tags : list[str] or None
         A list of tags for the task.
     task_document_kwargs : dict (deprecated)
-        Additional keyword args passed to :obj:`.ForceFieldTaskDocument()`.
+        Additional keyword args passed to :obj:`.ForceFieldStructureTaskDocument()`.
```
JaGeo, line 124:
```suggestion
        Additional keyword args passed to :obj:`.ForceFieldStructureTaskDocument()` or :obj:`.ForceFieldMoleculeTaskDocument()`.
```

yaoyi92: Thanks, I updated these descriptions in 4e9f061.

JaGeo:
I am happy to merge after these changes. I just fear we will have a CI failure now. mp-api and monty need to be updated

yaoyi92:
Dear @JaGeo, just check, do we have the CI issue fixed? I think this pull request is read to merge.

JaGeo:
Thank you. If the tests workflow run through, it will get merged!

JaGeo:
@yaoyi92 still some failures, unfortunately.

yaoyi92:
I fixed the incompatibility issue, how ever it seems there is still some error related to this merge of version bump of mdanalysis #1176.

yaoyi92:
Dear @JaGeo, can you help run the workflow again, I believe the testing can be fixed by this commit 95fd127.

JaGeo:
@yaoyi92 hey, do you still want to get it merged? Sorry for not following up!

yaoyi92:
> @yaoyi92 hey, do you still want to get it merged? Sorry for not following up!

Yes, if the CI works fine now. I can try to merge it.

yaoyi92:
Looks like all checks passed. Can you help merge it @JaGeo ?

esoteric-ephemera:
Thanks a lot @yaoyi92  for this! And very sorry to get back to this so late. I've opted to avoid breaking changes as much as possible, and have rearranged the class structures you started slightly.

For backwards compatibility, I've combined your `ForceFieldStructureTaskDocument` and `ForceFieldTaskDocument` into the existing `ForceFieldTaskDocument`. The `ForceFieldMoleculeTaskDocument` is more or less as you drafted

PR #1134, opened 2025-03-07
Fix mp api version

JaGeo:
## Summary

This should fix the current CI failures.


JaGeo:
Hi @gpetretto, @VicTrqt ,  could you take a look what is going on with abipy? mp-api forced us to update monty and new failures have been showing up. I would like to have an error-free github repo for the school next week.

JaGeo:
https://github.com/materialsproject/atomate2/pull/1134/commits/944070e7deb6208ad9f065f7e1b2d2871341ed86 can you have a look at this commit and the tests

gpetretto:
Hi @JaGeo, thanks for pointing it out. However it seems that the same tests pass in your last commit https://github.com/materialsproject/atomate2/pull/1134/commits/283ebe3b7f00a0e00cb9d1be8761c6f783236f01. Do you get the failure locally? Or do you need to revert to the https://github.com/materialsproject/atomate2/commit/944070e7deb6208ad9f065f7e1b2d2871341ed86 commit?

JaGeo:
I see @gpetretto . 

The whole purpose of this PR is to fix the mp-api issues that suddenly showed up. I had overlooked that the pymatgen abinit tests are passing now. Then, I will try to fix the new errors if possible...

orionarcher:
Hmmm, let me have a look. The discrepancy is quite surprising. I'll loop back around to it later this week.

JaGeo:
@orionarcher thank you! We have a larger atomate2 school next week. It would if we could get this fixed before next week.

esoteric-ephemera:
Hey @JaGeo we should probably split off the tests into isolated environments - since the openff tests are brittle, they could live in their own env, or an easier solution for now would be moving the `mp_api`-related tests to the forcefield test (that's completely isolated from abinit and openff)

Should I take a look at that?

JaGeo:
Yes, please, @esoteric-ephemera . This would be very much appreciated. It might also be worth to test several pymatgen versions etc. Maybe one is compatible with both mp-api and openff ...

PR #1158, opened 2025-03-24
Add MatPES forcefields

esoteric-ephemera:
- Adds MatPES-trained PBE and r<sup>2</sup>SCAN MLFFs with options to use different pre-trained model architectures
- Adds tests
- Slight tweak to base `AseMaker` class per the atomate2 workshop - this class just now needs a calculator plugged in to run a static / external interface to another code. Useful for examples in training / simple ASE stuff

Would also close #1156

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install  "setuptools<78"
+          pip install "setuptools==77.0.3"
           pip install .[strict,docs]
 
       - name: Copy tutorials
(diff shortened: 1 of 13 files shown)


JaGeo:
I am not sure how to continue with the torch conflicts. Moving every ml potentials in a separate environment is also not exactly ideal

esoteric-ephemera:
Yeah, I'm seeing if it's possible to use the matgl version of CHGNet instead of its standalone package. That would resolve the current set of conflicts but is still a band aid solution

Not sure it's an issue we can solve without a common interface between all the models (like matgl, but broader)

esoteric-ephemera:
OK so there's a big [breakage in the newer versions of setuptools](https://github.com/pypa/setuptools/issues/4910) and I'm not able to pin the setuptools version using pyproject.toml nor the github action. so for right now, the docs action is going to fail

JaGeo:
Thanks @esoteric-ephemera !

@utf I think we were all fine with giving @esoteric-ephemera the merge rights, right? 😀

PR #1159, opened 2025-03-28
Add example MP-compatible phase diagram tutorial

esoteric-ephemera:
It came up during the CECAM atomate2 workshop that it would be useful to have an example of how to perform MP-compatible workflow calcs, including how to apply corrections to these.

This adds a notebook to the tutorials on how to do this and a reference set of task docs. To handle the larger amount of raw VASP data used, I've added a `VaspTestData` class to `atomate2.utils.testing.vasp` which takes an input calculation directory and handles creation of a compressed JSON archive of the files for better organization / data usage
- I also tested using HDF5 and parquet to compress VASP test data, but the header / metadata fields take up a lot of space in these formats

Also added a utility function for extracting job UUID : name mappings from generic flows with replacements/etc. Can remove if requested, it's been useful for analyzing tests with complex flows when run locally.

Other changes:
- Repin pymatgen version in `strict-openff` dependencies - this keeps breaking tests
- Bumps pydantic and ensures AIMS models are fully defined, supersedes #1161


src/atomate2/utils/testing/vasp.py
```diff
@@ -15,7 +15,14 @@
 from pymatgen.util.coord import find_in_coord_list_pbc
 
 import atomate2.vasp.jobs.base
-import atomate2.vasp.jobs.defect
+
+try:
```
JaGeo, line 19: Thanks. Should have fixed this a long time ago..

tests/forcefields/flows/test_eos.py
```diff
@@ -2,6 +2,7 @@
 from jobflow import run_locally
 from monty.serialization import loadfn
 
+from atomate2.common.utils import get_job_uuid_name_map
```
JaGeo, line 5: Maybe rather in utils.testing?

JaGeo:
@esoteric-ephemera that's great. Thanks. 
The idea behind the test data was to ensure that the tutorials will not break in the future which happens quite frequently otherwise. But I get the point with the large files ...

tests/forcefields/flows/test_eos.py
esoteric-ephemera: Sounds good - moved / made a common testing utils file

esoteric-ephemera:
That makes sense, I'll see how much test data I can remove

esoteric-ephemera:
@JaGeo , @utf : I got the test data for this down to about 10MB so I'm adding it here. I also added convenience classes to make json-format archives of VASP calculations (`atomate2.utils.testing.vasp.VaspTestData`) that interface with `mock_vasp`

So from the user's side, to make VASP test data archive, they have to run two lines of code:

```python
from atomate2.utils.testing.vasp import VaspTestData

vtd = VaspTestData.from_directory("calculation directory")
vtd.to_file("name of the json.gz or json.bz2 file")
```

`mock_vasp` now checks for an archived test data format and appropriately unpacks it. The new classes handle making the directory structure, removing POTCAR info, etc.

Hoping this helps with file bloat / size as we grow tests out. If you want me to re-/move this, just let me know

JaGeo:
Great @esoteric-ephemera . I like it. I am fine with leaving it.
Can you maybe add this way to generate test data in the documentation as well?

JaGeo:
Thank you for the great work.  I will merge!

PR #1203, opened 2025-05-22
Remove walrus operator to avoid unassigned vmax

Jonas-Finkler:
## Summary

Fixes a bug in `mpmorph.py`. 

```Python
v0_in_range = (
    (vmin := working_outputs.get("Vmin"))
    <= v0
    <= (vmax := working_outputs.get("Vmax"))
)
```
If `v0 > vmin`, the expression evaluates to false and the second comparison is skipped, leaving `vmax` unassigned. 

I moved the two assignments out of the expression, avoid the walrus operators. 

## Checklist

title.

  type check your code.

### DIFF
--- a/src/atomate2/common/flows/mpmorph.py
+++ b/src/atomate2/common/flows/mpmorph.py
@@ -130,11 +130,9 @@ class EquilibriumVolumeMaker(Maker):
                 return Response(output=flow_output, stop_children=True)
 
             # Check if equilibrium volume is in range of attempted volumes
-            v0_in_range = (
-                (vmin := working_outputs.get("Vmin"))
-                <= v0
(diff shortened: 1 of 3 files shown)


Jonas-Finkler:
Hi @JaGeo, 
I just started using this workflow when I encountered the bug. 
I don't know my ways around the code base well enough to add a test for this with a reasonable amount of work. 
I was hoping that it would not be necessary since it's only a few lines.

esoteric-ephemera:
Thanks for the fix! I'll take a look today

esoteric-ephemera:
Also noticed that there was a bug in the keywords for the EOS initial fitting that has been corrected. Should be gtg once tests pass. Thanks @Jonas-Finkler!

PR #1204, opened 2025-05-23
switch to CONTCAR read for LOBSTERTaskDoc

naik-aakash:
LOBSTER run results sometimes not consistent when using POSCAR as VASP sometimes shifts origin leading to inconsistent neighbour detections


JaGeo:
Thanks @naik-aakash !

PR #1206, opened 2025-05-30
add file output xyz file including final atoms object to ase and forcefield jobs

JaGeo:
## Summary

This adds a default output file to all ase and forcefield jobs. The output file is an XYZ file including the final structure, energy, forces and stresses.

I think it is useful to have the final results of the ase runs. This then also allows a short file-based exploration if all intermediate steps look reasonable.


JaGeo:
cc @esoteric-ephemera for information in case you are not happy with this change.

esoteric-ephemera:
All good with me! Is it certain optimizers which go beyond the requested number of steps? That can be something we override in the static ASE jobs

JaGeo:
Added that fix in #1207 . Should be fine now, I think. I mainly wanted to have an output written for force fields. Makes it easier if you use memory databases

Issue #1212, opened 2025-06-06
BUG: atomate2 failed to compute the 3rd-order elastic tensor.


hongyi-zhao:
I try to use [jobflow-remote](https://github.com/Matgenix/jobflow-remote) to compute the 3rd-order elastic tensor with the following code:

```python
import numpy as np
from jobflow import run_locally
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
from atomate2.vasp.flows.elastic import ElasticMaker
from atomate2.vasp.powerups import (
    update_user_incar_settings,
    update_user_potcar_functional,
    update_user_kpoints_settings,
)
from pymatgen.io.vasp.inputs import Kpoints
from pymatgen.core import Structure
from mp_api.client import MPRester
from atomate2.vasp.powerups import add_metadata_to_flow
#https://github.com/materialsproject/atomate2/issues/1014#issuecomment-2454955891
from atomate2.vasp.flows.mp import MPGGADoubleRelaxMaker, MPGGAStaticMaker

from jobflow import Flow
from jobflow_remote import submit_flow

# Based on the following checking:
#werner@x13dai-t:~/Public/hpc/vasp/pot$ ug -r 'TITEL  = PAW_PBE Pt ' -g POTCAR
#potpaw_PBE.54/Pt/POTCAR:   TITEL  = PAW_PBE Pt 04Feb2005
#potpaw_PBE.64/Pt/POTCAR:   TITEL  = PAW_PBE Pt 04Feb2005
#potpaw_PBE/Pt/POTCAR:   TITEL  = PAW_PBE Pt 05Jan2001
#potpaw_PBE/Pt_new/POTCAR:   TITEL  = PAW_PBE Pt 04Feb2005
#potpaw_PBE/Pt_ZORA/POTCAR:   TITEL  = PAW_PBE Pt 05Jan2001
#potpaw_PBE.52/Pt/POTCAR:   TITEL  = PAW_PBE Pt 04Feb2005
#potpaw_LDA_PBE_52_54_orig/potpaw_PBE.54/Pt/POTCAR:   TITEL  = PAW_PBE Pt 04Feb2005
#potpaw_LDA_PBE_52_54_orig/potpaw_PBE.52/Pt/POTCAR:   TITEL  = PAW_PBE Pt 04Feb2005

#So, all the following versions are the same one in this case:
#user_potcar_functional = "PBE_64"
#user_potcar_functional = "PBE_52"
[...]

PR #1213, opened 2025-06-07
Fix: Support 3rd-order elastic tensor computation

hongyi-zhao:
- Add support for complex tensor structures in ElasticTensorDocument
- Implement recursive _safe_voigt_to_list() function
- Enable order=3 elastic tensor calculations
- Maintain backward compatibility for order=2 tensors

Resolves issue with 'list' object has no attribute 'tolist' error

Fix https://github.com/materialsproject/atomate2/issues/1212


esoteric-ephemera:
Thanks @hongyi-zhao!

PR #1221, opened 2025-06-14
Bump Pymongo

CompRhys:
Dump pymongo

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
     "pydantic-settings>=2.0.3",
     "pydantic>=2.0.1",
     "pymatgen>=2024.11.13",
-    "pymongo<=4.10.1",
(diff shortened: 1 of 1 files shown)


esoteric-ephemera:
Thanks @CompRhys but I think the underlying issue is maggma's MemoryStore depending on mongomock, which hasn't seen a release in about a year.

CompRhys:
I think I also went down that rabbit hole when looking at this but just left this open. I think that the actual way forward would be to just rip out all those tests and just run with a real version on mongo for testing both locally and in CI. Locally `docker compose up mongodb` isn't much of an overhead and then it's also fairly easy to spin up mongo in CI. I believe that this is already done for some of the tests.

esoteric-ephemera:
Yeah that would be better for the long term. Going to discuss with the rest of the MP team and move forward more on the maggma and emmet side of things to fix this

PR #1247, opened 2025-07-16
Fix RelaxConstVolSetGenerator

gpetretto:
## Summary

`RelaxConstVolSetGenerator` was not setting the INCAR values correctly.
Fixing this and adding tests for all the input sets in core to avoid this kind of issues.


JaGeo:
@gpetretto thank you. After looking at this though, I am wondering if this is really a "constant volume" Maker or rather a "constant lattice" Maker as it uses ISIF=2. I might open a new issue

Issue #959, opened 2024-08-20
Issue with VASP_VDW_KERNEL_DIR Setup Not Locating vdw_kernel.bindat File


YuanbinLiu:
Hi,

I have been working on setting up Atomate2 for my VASP calculations, and I encountered an issue related to the VASP_VDW_KERNEL_DIR variable. I followed the instructions provided in the Atomate2 documentation (available [here](https://materialsproject.github.io/atomate2/reference/atomate2.settings.Atomate2Settings.html#atomate2.settings.Atomate2Settings.VASP_VDW_KERNEL_DIR)) to configure the VASP_VDW_KERNEL_DIR variable. The goal was to ensure that the vdw_kernel.bindat file could be correctly located during my VASP calculations.

Despite following the documentation, it appears that the vdw_kernel.bindat file is still not being found during the calculations. I have double-checked the path and confirmed that the file exists at the specified location.

Could you please provide some guidance on what might be going wrong? Specifically:

1. Are there any common pitfalls or additional steps that I might have overlooked in setting up this variable?
2. Is there a way to verify within Atomate2 that the VASP_VDW_KERNEL_DIR variable is correctly pointing to the intended directory?

utf:
Hi @YuanbinLiu, can I just check if you're using a relative or absolute path, and whether you're using `~/` or putting the full path.

JaGeo:
@utf I had just looked for the constant in the code base and only found it showing up twice: https://github.com/search?q=repo%3Amaterialsproject%2Fatomate2+VDW_KERNEL_DIR&type=code

I am currently not sure how and where it should be used. I assume I am missing something here.

YuanbinLiu:
> Hi @YuanbinLiu, can I just check if you're using a relative or absolute path, and whether you're using `~/` or putting the full path.

Hi @utf, I am using the absolute/full path.

JaGeo:
Sorry for closing. This was a mistake from my side.

utf:
I agree @JaGeo. It looks like this option doesn't currently do anything. I think it would make sense to actually have this option in pymatgen and then automatically copy the kernel when writing the input set.

@YuanbinLiu Do you have access to VASP v6.4.3? If so, the kernel dir is not necessary as described on this page: https://www.vasp.at/wiki/index.php/Nonlocal_vdW-DF_functionals

YuanbinLiu:
@utf I have access. Implementing this feature in the future would enhance the program's flexibility.

[...]

PR #1267, opened 2025-08-12
Remove `VASP_VDW_KERNEL_DIR`

Andrew-S-Rosen:
Currently, it does nothing as noted in https://github.com/materialsproject/atomate2/issues/959. The better PR would be to make it work, but we shouldn't have an unused setting around since that will confused users more.


JaGeo:
Agreed. Also newer vasp versions don't need it anymore

PR #1269, opened 2025-08-14
Add support for a `CUSTODIAN_GZIPPED_OUTPUT` setting

### LINKED ISSUE #1268: Expose the Custodian `gzipped_output` keyword argument
In Atomate2's call to the `Custodian` class, there is no direct mechanism for the user to specify the `gzipped_output` keyword argument in the `Custodian` class.

https://github.com/materialsproject/atomate2/blob/34dcf1c93272856a055804c77e293bd0aa08d6fb/src/atomate2/vasp/run.py#L83-L94

Now, you might ask why one would want to access the `gzipped_output` keyword argument in the `Custodian` class if Atomate2 gzips files automatically. The reason is that if the user sets `SETTINGS.CUSTODIAN_SCRATCH_DIR` to a highly performant directory, the job will run there, but the files will be copied back to the initial directory where they will then be gzip'd. This is slow. Ideally, the gzipping should happen while the files are still in `SETTINGS.CUSTODIAN_SCRATCH_DIR`. Having Custodian do the gzipping would be ideal. Atomate2 will then ignore these when they're copied back.

Andrew-S-Rosen:
Closes #1268.

### DIFF
--- a/src/atomate2/settings.py
+++ b/src/atomate2/settings.py
@@ -44,6 +44,10 @@ class Atomate2Settings(BaseSettings):
     CUSTODIAN_SCRATCH_DIR: Optional[str] = Field(
         None, description="Path to scratch directory used by custodian."
     )
+    CUSTODIAN_GZIPPED_OUTPUT: bool = Field(
+        False,
+        description="Whether Custodian should gzip the files. This parameter "
+        "may make sense to use if you are setting CUSTODIAN_SCRATCH_DIR"
 
     # VASP specific settings
     VASP_CMD: str = Field(


esoteric-ephemera:
Overall fine with me, but isn't the same possible by setting `custodian_kwargs = {"gzipped_output": False}`?

src/atomate2/cp2k/run.py
```diff
@@ -118,6 +118,7 @@ def run_cp2k(
         validators=validators,
         max_errors=max_errors,
         scratch_dir=scratch_dir,
+        gzipped_output=SETTINGS.CUSTODIAN_GZIPPED_OUTPUT,
```
esoteric-ephemera, line 121: Can you move these lines to the `def run(...):`? That way they're not frozen by the atomate2 settings and can still be modified by kwargs

Andrew-S-Rosen:
> Overall fine with me, but isn't the same possible by setting `custodian_kwargs = {"gzipped_output": False}`?

How would the user pass in the `custodian_kwargs`? I think I may just be missing how to do that if it's not accessible by the settings.

esoteric-ephemera:
That's the more confusing part sorry: in a `BaseVaspMaker`-derived job, you would do set it via [run_vasp_kwargs](https://github.com/materialsproject/atomate2/blob/34dcf1c93272856a055804c77e293bd0aa08d6fb/src/atomate2/vasp/jobs/base.py#L229):
```
BaseVaspMaker(run_vasp_kwargs = {"custodian_kwargs":  {"gzipped_output": False}})
```

The design of `BaseVaspMaker` is opaque for this, might be best to expose `custodian_kwargs` there instead

Andrew-S-Rosen:
Ah! Okay, with that being said I think this PR should be closed. I don't think there is a need to make this setting more accessible. Thank you for the review and info!

esoteric-ephemera:
Sure thing! I do feel like kwargs for the Makers could be clearer, or we should have a few examples of how to use them

PR #1304, opened 2025-10-08
Fix ValueEnum import inconsistency - Replace emmet.core.utils.ValueEn…

tjaffrel:
Fix ValueEnum import error by using jobflow.utils.ValueEnum instead of emmet.core.utils.ValueEnum


esoteric-ephemera:
Thanks but this is being taken care of in this PR: https://github.com/materialsproject/atomate2/pull/1301

PR #1313, opened 2025-10-17
fix(docs): revise python version example

ramonmi:
## Summary

As mentioned above, _Make sure to create a Python 3.10+ environment as recent versions of atomate2 only support Python 3.10 and higher._


JaGeo:
Thanks, @ramonmi !

PR #1322, opened 2025-10-29
Transformation makers and SQS

esoteric-ephemera:
- Add makers to run pymatgen transformations as a job, with a `Transformer` class which only needs a transformation as input to run for ~90% of pymatgen transformations
- Add specific SQS implementation + tests

### DIFF
--- a/src/atomate2/common/schemas/transform.py
+++ b/src/atomate2/common/schemas/transform.py
@@ -1,12 +1,7 @@
 """Define schemas for SQS runs."""
 
+from emmet.core.types.enums import ValueEnum
 from pydantic import BaseModel, Field
-
-try:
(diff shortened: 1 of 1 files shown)


utf:
Just noting the [`TransmuterMaker`](https://github.com/materialsproject/atomate2/blob/058a9df0577184ec4d589e73d53484e9bb668fb4/src/atomate2/vasp/jobs/core.py#L576) in case you weren't already aware.

There is some overlap here the existing class both applies the transformation and then runs a calculation which is more limiting.

esoteric-ephemera:
Yep! The idea here was more flexibility like you're saying, specifically for SQS transformations (the output of the SQS transformation in particular is different from the other transformation classes)

PR #1332, opened 2025-11-17
Improve type safety and documentation in powerups modules

tomdemeyere:
## Summary

Made changes to `atomate2/vasp/powerups.py` and `atomate2/common/powerups.py`

- Improved docstrings
- Added a `JobType` TypeVar for functions accepting union of types such as `Job | Flow | Maker` so that the return type by the static type checker is the same as the input type.

Potential problem: at some point we can find `flow: Flow, additional_fields: dict, class_filter: Maker = BaseVaspMaker`, the class filter then get send to `update_custodian_handlers` where we can find `code = class_filter.name.split(" ")[1]`

Here an instance property is called on what is supposed to be a type, I assume this is done because in practice the property is only implemented in the `Maker` base class and conventionally overridden by a class attribute in most subclasses (hacky).

@JaGeo @naik-aakash 

## Checklist

  type check your code.


JaGeo:
Hi @tomdemeyere , 

Thanks for working on this.

Could you elaborate on how this new change affects usage of the powerups, and also check if the failing tests have anything to do with your changes? Thanks in advance!

src/atomate2/vasp/powerups.py
```diff
@@ -1,98 +1,111 @@
...
+        Default is None (no filtering).
+    class_filter : type[Maker] | None, optional
+        Filter to apply updates only to makers of this class or its subclasses.
+        Default is BaseVaspMaker.
 
     Returns
     -------
-    Job or Flow or Maker
-        A copy of the input flow/job/maker modified to use the updated incar settings.
+    Job | Flow | Maker
```
esoteric-ephemera: The `Job or Flow or Maker` is I think preferred since atomate2 uses [numpy-style docstr](https://numpydoc.readthedocs.io/en/latest/format.html) - can you change the pipes back to `or`?

esoteric-ephemera:
Thanks for the docstr updates @tomdemeyere! Some comments for stylistic consistency, ~also @JaGeo the failing tests were just the emmet-core incompatibility which should be resolved now~

esoteric-ephemera:
@tomdemeyere can you look at why the cp2k tests are failing from these changes?

tomdemeyere:
Thanks for you comments @esoteric-ephemera I did not know about "or" I always thought NumPy style was with pipes, it's nice to know!

Tests should be passing now.

src/atomate2/vasp/powerups.py
tomdemeyere: Done!

tomdemeyere:
> Could you elaborate on how this new change affects usage of the powerups

@JaGeo No runtime changes in this PR, only type hinting. The `TypeVar` introduced in functions ensure that if users send an object A to the function, they will get an object A back:

```python
# Previous behavior
my_job: Job
my_job = update_user_kpoints_settings(my_job, ...)
# `my_job` is now Job | Flow | Maker type due to the return type of the function

#Current behavior
my_job: Job
my_job = update_user_kpoints_settings(my_job, ...)
# `my_job` is still the same exact type `Job` due to the TypeVar
```

JaGeo:
To me everything looks fine 😀.

Are any comments still open from your side @esoteric-ephemera ?

JaGeo:
Thanks @tomdemeyere ! :)

PR #1354, opened 2025-12-18
Clean up structure/molecule aliases in ASE / MLFF jobs

### LINKED ISSUE #1353: Ionic steps data in ASE Job
While running MLFF calculation with ASE based jobs I encountered two problems with the data being stored in the ionic steps. 

The first is that the base `AseRelaxMaker` has a `ionic_step_data` attribute that contains `"mol_or_struct"` in its default:

https://github.com/materialsproject/atomate2/blob/1822f9d6f7ad3fdebd44a81e835fdcfefb8222a9/src/atomate2/ase/jobs.py#L84-L90

Based on this and on the docstrings, I assumed that if I provided a list excluding `"mol_or_struct"` the structure would have not been saved. However, in the code, as long as there are some data for ionic steps stored, the `mol_or_struct` element is always stored: https://github.com/materialsproject/atomate2/blob/1822f9d6f7ad3fdebd44a81e835fdcfefb8222a9/src/atomate2/ase/schemas.py#L530

I think that it would be correct to exclude `mol_or_struct` from the output if not present in `ionic_step_data`, but if not, at least this should be made clear in the docstrings and `mol_or_struct` removed from the default.

The second issue is that in `IonicStep` the `structure` and `molecule` are automatically filled in based on `mol_or_struct`:
https://github.com/materialsproject/atomate2/blob/1822f9d6f7ad3fdebd44a81e835fdcfefb8222a9/src/atomate2/ase/schemas.py#L124
However, this makes so that when the output is serialized it contains the structure twice for each ionic step. Is there any reason for `structure` and `molecule` not being `@property`? Or at least they could be set as `Field(..., excluded=True)` so that the same data will not be stored twice. 

I can open a PR to address these issues if needed, but I first wanted to check if these were the result of a specific choice/need and if there is any preference about how to address them.

### LINKED ISSUE #1364: Streamlined keyword arguments for matgl models
https://github.com/materialsproject/atomate2/blob/702724784e1ed09bca0879f9df20a9b36bf78f86/src/atomate2/forcefields/utils.py#L59-L77

Currently, CHGNet and M3GNet use `stress_weight=_GPa_to_eV_per_A3`, but the MatPES models use `stress_unit = "eV/A3"`. I think [these should do the same thing](https://github.com/materialyzeai/matgl/blob/7100676ebdc81f2f0a3da7d431af167d63be6746/src/matgl/ext/_ase_pyg.py#L162-L170) so for consistency and to prevent silly mistakes, maybe we should adopt the same keyword arguments for the different matgl-based models.

esoteric-ephemera:
Close #1353 by:
- Making `structure` and `molecule` properties that get populated by `mol_or_struct` in base ASE / MLFF docs
- Backwards compatibility: `mol_or_struct` gets populated by alias if `structure` or `molecule` is present
- Allow for removing `mol_or_struct` in ionic steps, and by specifying either `mol_or_struct`, `molecule`, or `structure`

Close #1364 by:
- Migrating CHGNet to `matgl` away from the `chgnet` package (slims down dependency stack as well)
- Use same kwargs for all `matgl`-based MLFFs

### DIFF
--- a/docs/dev/dev_install.md
+++ b/docs/dev/dev_install.md
@@ -51,7 +51,7 @@ If you're planning on contributing to the atomate2 source, you should also insta
 the developer requirements with:
 
 ```bash
-pip install -e .[dev]
+pip install -e . --group dev
 pre-commit install
 ```
 
@@ -65,7 +65,7 @@ Unit tests can be run from the source folder using `pytest`. First, the requirem
 to run tests must be installed:
 
 ```bash
-pip install .[tests]
+pip install . --group tests
 ```
 
 And the tests run using:
@@ -79,7 +79,7 @@ pytest
 The atomate2 documentation can be built using the sphinx package. First, install the requirements:
 
 ```bash
-pip install .[docs]
+pip install . --group docs
 ```
 
 Next, the docs can be built to the `docs_build` directory:


gpetretto:
Thanks @esoteric-ephemera for the fix! It looks good to me.

I just wanted to mention that the alias+property at the moment does not create any problem since in `@forcefield_job` no `output_schema` is defined. However, if this changes in the future, the following code can give an error:
```python
job1=ForceFieldRelaxMaker(force_field_name="MACE_MP_0B3").make(s)
job2=ForceFieldRelaxMaker(force_field_name="MACE_MP_0B3").make(job1.output.output.structure)
```

> AttributeError: OutputDoc does not have attribute 'structure'.

In light of this, I think it may be worth to update jobflow to handle cases like this. The change should be relatively small. What do you think @utf?

esoteric-ephemera:
Thanks @gpetretto - we could move the `structure` attribute access over to `mol_or_struct` access instead. That would also make jobs / flows apply to molecules and structures simultaneously

gpetretto:
> Thanks @gpetretto - we could move the `structure` attribute access over to `mol_or_struct` access instead. That would also make jobs / flows apply to molecules and structures simultaneously

This is probably a good idea and would definitely help for the workflows in atomate. 
However, it may not cover the case of private workflows that users may have implemented (I do have a subclass of the Ase job for a project).

Overall, I don't think there will be a downside in supporting properties at the jobflow level, so that the general case can be covered in the future. If I have time I will try to open a PR for jobflow.

PR #1358, opened 2025-12-23
Clarify pressure parameter description in qha.py

Andrew-S-Rosen:
The units weren't mentioned for the `pressure` keyword argument in the `CommonQhaMaker`. I believe it is GPa.


esoteric-ephemera:
@JaGeo can you confirm the units here? Not as familiar with the QHA workflow

JaGeo:
@Andrew-S-Rosen @esoteric-ephemera we use the phonopy interface here: https://github.com/materialsproject/atomate2/blob/717c320806a15950e9f71daa03f4b71f0f42327e/src/atomate2/common/schemas/qha.py#L138

Thus, GPa

PR #1365, opened 2026-01-02
Fix JSONStore and add MemoryStore option in Installation Docs

thomasloux:
## Summary

Fix the installation JSONStore (__init__ requires paths and not uri). Also add MemoryStore in order to quickly test the installation as JSONStore does not work with all workflows (failing with PhononMaker with TypeError: Type is not JSON serializable: ScalarFloat).

## Checklist

title.

  type check your code.

### DIFF
--- a/docs/user/install.md
+++ b/docs/user/install.md
@@ -270,8 +270,19 @@ from) by clicking "Network Access" (under "Security" in the left hand menu) and 
 ````
 
 ````{note}
-If you do not have access to a Mongo database, you can run `atomate2` using a local `.json` file
-to store outputs using the following `jobflow.yaml` file.
+If you do not have access to a Mongo database, you can run `atomate2` without setting anything. Under the hood,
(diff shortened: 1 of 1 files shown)


JaGeo:
@thomasloux Thank you! I kind of forgot that this is still part of the Installation instructions!

I think we should suggest the MemoryStore as a default. Actually, it should work without the setup as well.

Would you mind changing this as part of this PR?

thomasloux:
Indeed it works without setting anything (don't provide jobflow.yaml).

Do you want MemoryStore as default before MongoStore or just before JSONStore ?

JaGeo:
@thomasloux just before JSONStore. 

MongoStore is still the better option if you want to run high-throughput. We have been working on this as well to provide a different option.

JaGeo:
@thomasloux Thanks! 

If you want to, you can add yourself to the list of contributors in your second PR!

PR #1366, opened 2026-01-02
Move pre-commit in dev for automatic install

thomasloux:
## Summary

In current installation, you need to run `uv sync --extra dev` to use pre-commit. I've just installed it using `uv add --dev pre-commit`so that this dev group is automatically installed when running uv sync.

## Checklist

title.

  type check your code.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .[strict,docs]
+          pip install .[strict] --group docs
 
       - name: Build
         run: sphinx-build docs docs_build
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,7 +34,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install "setuptools==77.0.3"
-          pip install .[strict,docs]
+          pip install .[strict] --group docs
 
       - name: Copy tutorials
         run: |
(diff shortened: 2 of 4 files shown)


JaGeo:
@thomasloux Thank you. I am not a pre-commit expert. Is the version update stricly required?

thomasloux:
No, it just updated by running `uv add --dev pre-commit`. But pre-commit runs fine with the new version. I can roll back to the same version to prevent error.

JaGeo:
I see. Thanks. I think we can keep the old version.

Could you check the documentation for the pre-commit installation and update this one?

thomasloux:
What do you want me to change regarding pre-commit? There is nothing about pre-commit in the docs as far as I know: https://materialsproject.github.io/atomate2/about/contributing.html

JaGeo:
It is here: https://materialsproject.github.io/atomate2/dev/dev_install.html

Can you add the uv options and check if all pip ones are updated? That would be great!

esoteric-ephemera:
Just noting: we don't use `uv sync` in CI (see the [tests workflow](https://github.com/materialsproject/atomate2/blob/717c320806a15950e9f71daa03f4b71f0f42327e/.github/workflows/testing.yml#L70)), you can also just `uv pip install atomate2[dev]` to pick up these extra development-only requirements

Not sure this needs to be modified

thomasloux:
This does not need to be updated. I was mainly suggesting an alternative configuration that I find more convenient. This is independent of the CI workflow.

thomasloux:
@esoteric-ephemera in CI you can also replace 'pip install .[extra1, extra2]' with 'uv sync --extra extra1 --extra extra2' if you have uv installed and want to only use uv and drop pip.

esoteric-ephemera:
Thanks! We should probably make a decision regarding using `uv` more completely with atomate2 but that can come in a separate fix, will merge this in for now

JaGeo:
@esoteric-ephemera I think this would be great! Also, adding this to our documentation would be good

PR #1383, opened 2026-02-09
Openff tests

esoteric-ephemera:
Pin setuptools < 81 in openff tests because the `smirnoff99frosst` package used in those tests requires `pkg_resources`. Will need to migrate these tests to a newer forcefield upon a new release of openff

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -132,8 +132,8 @@ jobs:
         run: micromamba run -n a2 pip install uv
 
       - name: Install conda dependencies
-        run: |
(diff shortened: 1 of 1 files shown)


mattwthompson:
FYI
- We re-released (https://github.com/openforcefield/smirnoff99Frosst/releases/tag/1.1.0.1) and rebuilt (https://github.com/conda-forge/smirnoff99frosst-feedstock/pull/2) this package with version 1.1.0.1 that should fix this incompatibility
- The force field itself is hardly more than a historical artifact at this point; there isn't a strong reason to use it over Sage (OpenFF 2.x, 2.3.0 being the most recent release) and Parsley (OpenFF 1.x) lines

esoteric-ephemera:
Thanks @mattwthompson! We're using it only as a lightweight forcefield for testing, but we should probably switch over to something more modern

Issue #1376, opened 2026-01-27
Discussion: Testing ML potentials


esoteric-ephemera:
A couple of the ML potentials/forcefields are currently untested (MatterSim, Allegro, and the Fairchem potentials) because of non-compatible dependencies (all three need a much newer version of `e3nn` than MACE supports, for example)

Breaking the tests apart to skip when a forcefield is not installed and launching separate environments of "dependency-compatible" forcefields would be needed to ensure implementation testing

This is on the dev to-do but help would be greatly appreciated!

PR #1388, opened 2026-02-18
Reorganize forcefield tests

esoteric-ephemera:
In the hopes of progress towards closing #1376
- Create separate forcefield dependency groups based on compatible install environments
- Skip forcefield tests when not installed (probably best practice anyways? we need to be more careful about this when developing)
- Separate forcefields by test group in CI 
- MatterSim tests

Docs:
- Expanding user forcefield docs with FAIRChem install instructions + more metadata
- Added user add-on info
- Added dev forcefield maintenance strategy guide

Tangential:
- Bump numpy<3 (except in a few edge cases)
- Fix APIs to Nequip, Allegro, and FAIRChem

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_language_version:
 exclude: ^(.github/|tests/test_data/abinit/)
 repos:
 - repo: https://github.com/charliermarsh/ruff-pre-commit
-  rev: v0.15.0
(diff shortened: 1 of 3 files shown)


JaGeo:
@esoteric-ephemera thanks again for this work! Could you maybe add a few comments directly to our pyproject yaml or elsewhere to make sure that we do keep this consistent in the future?

PR #1389, opened 2026-02-20
Inherit magmoms by default in VASP NSCF sets

### LINKED ISSUE #1377: BUG: MAGMOM removal in (HSE)BandStructureMaker
Hi there!

When using a (HSE)BandStructureMaker I noticed that the NSCF calculations (bands and dos) does not have the MAGMOM in the INCAR and I get the following warning in the OUTCAR:
```
|   You use a magnetic or noncollinear calculation, but did not specify
|   the initial magnetic moment with the MAGMOM tag. Note that a
|   default of 1 will be used for all atoms. This ferromagnetic setup
|   may break the symmetry of the crystal, in particular it may rule
|   out finding an antiferromagnetic solution. Thence, we recommend
|   setting the initial magnetic moment manually or verifying carefully
|   that this magnetic setup is desired.
```

And results, i.e. the DOS, are quite different between the cases with and without keeping the MAGMOM.

Now, looking at the VASP wiki https://www.vasp.at/wiki/MAGMOM I see the following:
"When restarting a magnetic calculation, MAGMOM is only used to determine the symmetry of the system and not to set the on-site magnetic moment. Therefore, if you remove the MAGMOM tag before restarting from a converged WAVECAR or CHGCAR, the magnetization is likely to be symmetrized away."

So, I suppose that MAGMOM should remain in the INCAR of a NSCF calculations from the previous static one.

I think the lines that remove the MAGMOM are
https://github.com/materialsproject/atomate2/blob/6b19e559c0d935f25b79e4b240c85ac396d4ed66/src/atomate2/vasp/sets/core.py#L262
https://github.com/materialsproject/atomate2/blob/6b19e559c0d935f25b79e4b240c85ac396d4ed66/src/atomate2/vasp/sets/core.py#L523

What seems odd is that those lines seem to be added a long time ago and nobody complained.
Could this warning be added in recent versions of VASP? I'm using the 6.4.2.

Let me know what you think.

Thanks!

esoteric-ephemera:
Closes #1377 by adding a `remove_magmoms` kwarg to the VASP bandstructure set generators which defaults to `False`. This will ensure MAGMOMs are inherited from a previous run.

The previous behavior is accessed only if set to `True`. Slightly breaking behavior but probably more in line with what users expect to happen

### DIFF
--- a/src/atomate2/vasp/sets/core.py
+++ b/src/atomate2/vasp/sets/core.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from copy import deepcopy
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
@@ -20,6 +21,16 @@ if TYPE_CHECKING:
 
 
 logger = logging.getLogger(__name__)
+
+
+def _emit_magmom_warning() -> None:
+    warnings.warn(
+        "Removing the MAGMOM tag is not recommended generally, "
+        "but is permitted to allow for previous behavior in atomate2. "
+        "See https://vasp.at/wiki/MAGMOM to understand how "
+        "magnetic initialization is affected by MAGMOM, CHGCAR, and WAVECAR.",
+        stacklevel=2,
+    )
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


esoteric-ephemera:
@fraricci if there's any other change you found necessary besides removing the `None` MAGMOM update, please let me know!

fraricci:
Thanks @esoteric-ephemera. For what I tested, that was enough to have the previous and correct behavior.

PR #1402, opened 2026-03-05
Update UPET default

frostedoyster:
Hello everyone and thanks for adding our UPET models (#1400)! I'm just submitting a small change to use "pet-mad-xs" (r2SCAN) instead of the "pet-omad-xs" model (PBEsol) that we might deprecate soon.

As a follow-up question: would it be possible to use our S model instead of XS? Our S model is significantly more accurate and I'm confident that it's faster than almost all the other default ML models in atomate2

### DIFF
--- a/tests/forcefields/flows/test_phonon.py
+++ b/tests/forcefields/flows/test_phonon.py
@@ -64,6 +64,7 @@ def test_phonon_maker_initialization_with_all_mlff(
         },
         MLFF.NEP: {"model_filename": f"{test_dir}/forcefields/nep/nep.txt"},
         MLFF.DeepMD: {"model": get_deepmd_pretrained_model_path},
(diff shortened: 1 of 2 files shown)


esoteric-ephemera:
How large is the S model vs. the XS model? My concern there is mostly on forcing the user to download large files / test execution time in CI

frostedoyster:
Thanks for the quick reply! XS is 20 MB, S is 110 MB

esoteric-ephemera:
OK S should be fine for the default, but for tests (really just ensuring the APIs/dependencies are aligned) I would prefer to do the XS model

Let me know if you need help with the corresponding test changes

JaGeo:
@esoteric-ephemera this sounds like a very good solution

esoteric-ephemera:
Thanks @frostedoyster!

PR #1196, opened 2025-05-06
Faster Phonon Forcefield Workflow By Introducing Batch ASEMakers (especially relevant for remote execution)

naik-aakash:
# Summary
This is an initial experimental code implementation for combining the phonon workflow in a single job when using forcefields. Its not very up to date with changes in over a year. Needs quite bit of modifications to make it work with updated API.

Will be adapted by @JaGeo further.

PR #1456, opened 2026-04-07
Alternative ASE/MLFF batch mode

esoteric-ephemera:
Possibly supersedes [#1196](https://github.com/materialsproject/atomate2/pull/1196). Tagging @naik-aakash and @JaGeo for thoughts.

The goal here is to make smaller structural changes overall to the jobs/workflows and ensure that each calculation maps to one output document

Changes:
- Allow for submission of a list of pymatgen Molecule or Structure to `AseRelaxMaker` and thus `ForceFieldRelaxMaker` / `ForceFieldStaticMaker`
- When submitting a list, a list of `TaskDocument`s are returned
- Cache ASE calculator once loaded to permit batch mode

### DIFF
--- a/src/atomate2/ase/jobs.py
+++ b/src/atomate2/ase/jobs.py
@@ -54,8 +54,7 @@ class AseMaker(Maker, ABC):
     class EMTStaticMaker(AseMaker):
         name: str = "EMT static maker"
 
-        @property
-        def calculator(self):
+        def _get_calculator(self):
             return EMT()
     ```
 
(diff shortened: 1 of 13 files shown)


JaGeo:
@esoteric-ephemera looks great to me! 

Did you think about a way to benchmark if there is indeed an advantage timewise, e.g. when executing in jobflow only. I think it will simplify execution with jobflow-remote!

esoteric-ephemera:
I can try to throw together a quick benchmark. Any concerns about the change in output documents? Not sure how much you've been using #1196 to do calculations, but this would change how the outputs are structured (here: a list of TaskDocument vs. 1196: TaskDocument having a list of structures/energies/etc.)

JaGeo:
We haven't used it much since the PR. There were some plans initially but we stopped this a while back. As long as the phonon workflow runs, I am happy.

JaGeo:
My hope is that we might build a thermal conductivity workflow in the future. Here, the batch mode might be really useful as well.

PR #1461, opened 2026-04-15
Forcefield roundtrip when specifying monty dict

### LINKED ISSUE #1460: BUG: serialization of ForceFieldMixin subclasses with dict `force_field_name`
In order to use a calculator for a Force field that is not included among the standard ones (e.g. the original chgnet) the suggested solution is to pass 
```
    {
        "@module": "chgnet.model.dynamics",
        "@callable": "CHGNetCalculator"
    }
```
to the `force_field_name` in the subclasses of `ForceFieldMixin`.
https://github.com/materialsproject/atomate2/blob/44fd9b1e739f779dbb8530127f07613c8d51bdc8/src/atomate2/forcefields/utils.py#L247

The problem is that, while this works with the generated Job object, it fails when the object is serialized. The following example:
```python
from atomate2.forcefields.jobs import ForceFieldRelaxMaker
from monty.json import MontyDecoder, MontyEncoder
import json

mgo_structure = Structure(
    species=["Mg", "O"],
)

flow = ForceFieldRelaxMaker(force_field_name = {"@module": "chgnet.model.dynamics", "@callable": "CHGNetCalculator"}).make(mgo_structure)
flow = json.loads(json.dumps(flow, cls=MontyEncoder), cls=MontyDecoder)
run_locally(flow)
```
fails with
> ValueError: Could not create ASE calculator for MLFF.Forcefield.

This leads to a failure in jobflow-remote, that needs to serialize the job in the DB, but I suppose the same would happen with Fireworks.

The primary reason is that the `calculator_meta` is set at creation time, but not serialized due to the presence of `field(init=False)`
https://github.com/materialsproject/atomate2/blob/44fd9b1e739f779dbb8530127f07613c8d51bdc8/src/atomate2/forcefields/utils.py#L178

Simply removing the `init=False` does not solve the issue though, because upon deserialization of the Job the `{"@module": "chgnet.model.dynamics", "@callable": "CHGNetCalculator"}` is automatically deserialized by the `MontyDecoder` and the code only expects either an `MLFF` or a `dict` for `calculator_meta`.

I managed to make it work by allowing `calculator_meta` to also be a `Calculator` subclass. I can open a PR if this seems a reasonable solution, but maybe there are better options?

esoteric-ephemera:
Closes #1460:
- Allows for specifying either:
  1. `force_field_name` as a `monty`-style dict
  2. `calculator_meta` as a dot-separated import str (preferred)
  3. `calculator_meta` as a `monty`-style dict
- Regression test for roundript, thanks to @gpetretto's example
- Updated docs accordingly

For the next minor release (v0.2.0), the forcefield interface could use some significant simplification. This will just serve as a patch

### DIFF
--- a/docs/user/codes/forcefields.md
+++ b/docs/user/codes/forcefields.md
@@ -48,16 +48,28 @@ Support is provided for the following models, which can be selected using `atoma
 
 ## Using custom forcefields by dictionary
 
-`force_field_name` also accepts a MSONable dictionary for specifying a custom ASE calculator class or function [^calculator-meta-type-annotation].
-For example, a `Job` created with the following code snippet instantiates `chgnet.model.dynamics.CHGNetCalculator` as the ASE calculator:
+`force_field_name` also accepts an import-like string, or MSONable dictionary to specify a custom ASE calculator class or function [^calculator-meta-type-annotation].
+For example, a `Job` created with the either of the following two code snippets instantiates a `chgnet.model.dynamics.CHGNetCalculator` as the ASE calculator.
 ```python
+# simple import string
 job = ForceFieldStaticMaker(
-    force_field_name={
+    calculator_meta="chgnet.model.dynamics.CHGNetCalculator",
+).make(structure)
+```
+
+or using `force_field_name` when
+
+```python
+# monty MSONable style
+job = ForceFieldStaticMaker(
+    calculator_meta={
         "@module": "chgnet.model.dynamics",
         "@callable": "CHGNetCalculator",
     }
 ).make(structure)
 ```
+Note that one can also specify `force_field_name = {"@module": ...,"@class": ...}` in the second example for backwards compatibility.
+However, this may not be preserved in future versions, and `calculator_meta` is preferred.
 
 [^calculator-meta-type-annotation]: In this context, the type annotation of the decoded dict should be either `Type[Calculator]` or `Callable[..., Calculator]`, where `Calculator` is from `ase.calculators.calculator`.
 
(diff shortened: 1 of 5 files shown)


gpetretto:
Hi @esoteric-ephemera, thanks for implementing this fix.

I think the solution may be fine, but I wonder if the fact of transforming the dict into an ad hoc string would not increase the likelyhood of ending up with backward incompatible updates later on. It seems that this string serialization may complicate any different use of dict or str values for `calculator_meta`. 
One concern that I have is that future changes may lead to serialized jobs in a DB that may not be deserialized. 

However, if you expect that v0.2.0 will introduce breaking changes anyway, it may not be worth to focus on potential future issues here.

esoteric-ephemera:
Thanks @gpetretto and good point. I added a bit more logic to the deserialization that should handle cases when `force_field_name` was stored as a `dict` in JSON. Also added a test for this

If you have some legacy docs where you could test that this correctly handles legacy docs, would appreciate it!

gpetretto:
Thanks for this update! I have tested and it seems to work fine deserializing this type of makers.

I should mention that in my previous message I was also refferring to potentially future incompatibilities. Serializing the dictionary with `calculator_meta=".".join(calculator_meta[k] for k in ("@module", "@callable"))`, if another kind of `str` or `dict` may need to be passed to calculator_meta in the future it may be difficult to preserve backward compatibility.

esoteric-ephemera:
Unrelated but `pyfhiaims` removed a parsing feature in `0.1.0` that now means the parsing in atomate2 no longer works. Pinning this until the `pyfhiaims` devs can chime in

esoteric-ephemera:
@gpetretto I agree it will hard to enforce strict backwards compatibility but I'm not sure what a better solution would be for now. The validation on import strings and monty-style dicts is pretty robust now

I'll try to release this as a patch by today just so this feature is working

PR #1495, opened 2026-06-16
Update FF dependencies.

shyuep:
Cleaning up some dependency specifications for force fields. In general, do not pin dependencies in pyproject.toml.


shyuep:
@JaGeo I have properly fixed the dependency definitions in this PR (including using uv.conflicts). I have no time to hunt down the rest of the test failures. So I will leave it to you to take it from here. You can close the other PR on upgrading matgl.

shyuep:
@JaGeo The rest of the failures are not related to what I did. So I will not handle them further. Thanks.

JaGeo:
@shyuep Thanks a lot!

JaGeo:
The problem here is connected to the test infrastructure and the required resources. Locally, the tests are passing. It seems that the run times and memory requirements have changed as part of the update.

I will try to switch to different forcefields in the test to balance the resource requirements out. As a first step, I will remove some tests and will try to figure out which of the tests is mainly responsible.

akwarii:
Hello, I just saw that this PR is updating to version v0.0.6+. In this version, torchsim decided to upstream most models to first party packages, which resulted in changes in some model API. I have an updated version of `atomate2.torchsim.core.pick_model` that account for these changes and add D3 support (as well as a foundation model dowloader). If you're interested I can open a PR after this one is merged or even contribute it here.

JaGeo:
@akwarii Thanks. Currently, the tests are running. How would this go beyond the current implemenation?

JaGeo:
Note for myself: I need to decide on whether I revert back to actually strict dependencies

Tests work otherwise.

Update: I have decided to switch back to strict dependencies. And I will remove mattersim for now, as phono3py should not be required to run a force field.

akwarii:
@JaGeo There are multiple points. First, about `pick_model`:
1. The current test suite for atomate2.torchsim only covers MACE and LJ models, that's why the test suite is still working
2. Fairchemv1 and graphPES were dropped by torchsim team. This PR also drops mattersim (anyway their torchsim wrapper isn't working, see [#163](https://github.com/microsoft/mattersim/issues/163#issue-4626457517))
3. Models such as orb are passing a model instance to torchsim wrapper instead of a model path
4. I also add the option to use D3 correction for any model through `torch_sim.models.dispersion.D3DispersionModel` and `torch_sim.models.interface.SumModel`

I also have a foundation model registry that download a foundation model giving its name, similar to `mace.calculators.foundations_models.download_mace_mp_checkpoint` but for MACE, NequIP, Allegro, and UPET (the others already have automated way to do so). For NequIP/Allegro models, the model compilation is also performed based on the available torch version, device, and accelerator. When the model is loaded, the correct device will be set and the accelarator library will be imported (if needed).

JaGeo:
@akwarii Happy to receive a PR about the torchsim update and a corresponding test for the different models that would not work without the PR.

JaGeo:
Thanks again @shyuep ! 

In general, I think the system requirements for running the matgl models have however changed as the tests were suddenly failing with the new matgl model. I am not sure why this is the case. Maybe, however, this is acceptable for users as they run the models on clusters.

PR #1517, opened 2026-07-21
Add SIESTA DFT code support (input sets, jobs, flows, schemas, docs, tutorials)

arsalan-akhtar:
## Summary

This PR adds first-class support for the [SIESTA](https://siesta-project.org/) DFT
code to atomate2, following the same architecture as the existing VASP/ABINit/FHI-aims
integrations. Everything lives under `src/atomate2/siesta/` and is opt-in via an extra
(`pip install atomate2[siesta]`); no existing code paths are modified.

SIESTA is a widely used, linear-scaling DFT package (numerical atomic orbitals,
norm-conserving/PSML pseudopotentials). This integration lets users drive SIESTA through
jobflow the same way they drive the other codes.

## What's included

- **Input sets** (`siesta/sets/`) — FDF input generation with a tiered parameter system,
  per-element basis handling, and k-point/band-path helpers.
- **Jobs & flows** (`siesta/jobs/`, `siesta/flows/`) — relaxation, static, band structure,
  DOS/PDOS, phonons (phonopy), EOS, elastic constants, NEB, surfaces & adsorption,
  point defects (+ charge corrections), and molecular/gas-phase calculations.
- **Schemas** (`siesta/schemas/`) — pydantic task/calculation documents parsed from SIESTA
  output (via `sisl`).
- **Custodian handlers/validators** (`siesta/custodian/`) — SCF/geometry/memory/wall-time
  error recovery.
- **Recipes** (`siesta/recipes/`) — high-level one-call workflow helpers.
- **CLI** (`siesta/cli/`) — structure manipulation, pseudopotential, database and
  jobflow-remote utilities.
- **Docs** (`docs/siesta/`) — nested under User Guide → Codes → SIESTA; builds with the
  main Sphinx site.
- **Tutorials** (`tutorials/siesta/`) — runnable scripts/notebooks (inputs only, no outputs).

## Dependencies

Adds a `siesta` optional-dependency group (`sisl`, etc.). SIESTA itself and
pseudopotentials are user-provided at runtime; the test suite ships a small packaged
pseudopotential set and is fully self-contained (no network, no external binaries).

## Testing

- ~2,100 unit/integration tests under `tests/siesta/`, all passing locally.
- Tests are hermetic: conftest fixtures provide packaged pseudopotentials and stub
  templates, so no SIESTA binary or network access is required.
- The full tutorial set has additionally been executed end-to-end aga
  5.4 build.

## Code quality

- `ruff check` and `ruff format` are clean across the entire `src/atomate2/siesta/` tree.
- `mypy` passes on the new `src/atomate2/siesta/` code.
- NumPy-style docstrings throughout.

## Notes for reviewers

- This is a large, **self-contained** addition — everything is under `siesta/` and behind
  the optional extra, so it can't affect existing codes.
- Happy to split into smaller reviewable pieces if preferred (e.g. sets+jobs first, then
  flows, then CLI/docs) — let me know what's easiest to review.

## Checklist

- [x] Code is formatted and linted (`ruff check`, `ruff format`).
- [x] Type annotations added; `mypy` passes on the new code.
- [x] NumPy-style docstrings on public API.
- [x] Unit tests added and passing.
- [x] Documentation and tutorials added.


esoteric-ephemera:
Hey @arsalan-akhtar , thanks for the contribution. While I think we would generally welcome an interface to siesta, I have a couple of serious concerns about this PR.

- Clear to me that this is agentically driven and while that's generally fine, I would prefer a narrow scope for a PR. You are more than welcome to use an agent to build scaffolding for a narrowly-scoped PR. A 444k positive diff is unreviewable
- The changes to the docs are not in keeping with how they are built
- Why are there functionalities like database interactions
- I see a CLI for siesta, which is not in keeping with the general atomate2 paradigm. This would need justification
- Any custodian-related changes should be made in a PR to [its repo](https://github.com/materialsproject/custodian), not `atomate2`'s
- What purpose do the very specialized workflows like "electrocatalysis" serve here?
- You need to explain which workflows, jobs, etc. are added
- Why are so many dataclasses needed to structure I/O with siesta?
- There are changes made to the `common` part of atomate2 that are unexplained

If you would like us to consider this for an addition to atomate2, I would need human answers to those questions per the Materials Project's AI policy, and I would also need this to be broken into smaller scopes that a human such as myself can review

arsalan-akhtar:
Hey @esoteric-ephemera  Thanks for the careful read... A bit of backstory on the size: I originally built this as a **"separate, standalone SIESTA package"** and worked on it for several months (well more than a year!) on its own. Well later did I decide it really belonged inside atomate2 — so I folded the whole thing back in at once, which is how you ended up with a ~500k diff landing in one shot 😅. Totally see and understand  not a reviewable unit. 😂

So: I'll restart with a small core-only PR — SIESTA input sets, StaticMaker + RelaxMaker, a task-document schema, and the run/files plumbing, with tests and a single docs page modeled on an existing code (aims). Everything else comes out and follows separately, if you're interested. 

Regarding  your specific questions:

1. **The changes to the docs are not in keeping with how they are built **: 
the reason I changed the build was to add togglable multilingual documentation, an RTL-aware language switcher with a full Persian/Farsi translation of the SIESTA docs extendable for any other langauge... . That's what the changes to conf.py, the page template, and the extra docs/siesta/ tree are for. I understand that's a departure from how atomate2 builds its docs. If it's not something you want in the project, no problem at all — I'll revert to the standard approach: a single siesta.md page (like vasp.md), with the API reference picked up by the existing central autodoc, and no changes to the shared docs infrastructure. Just let me know which you'd prefer.

2. **Why are there functionalities like database interactions  "AND" 3. I see a CLI for siesta, which is not in keeping with the general atomate2 paradigm. This would need justification**:
this is an extra convenience feature I added on top of atomate2 to make users' lives easier 😁: a small CLI (setup / test / status / query) that lets people who aren't already familiar with the jobflow/maggma stack check their MongoDB connection and browse stored calculations without writing any code. It lives entirely under cli/ and nothing in the core library (sets, jobs, flows, schemas) depends on it — it's purely a user-facing helper. I understand it overlaps with what jobflow-remote and maggma already provide, ...  The idea was zero-memorization access/guide for people that  not comfortable writing Python/ or lazy like me! : an interactive workflow/script generator, a tutorial browser, pseudopotential and structure-manipulation helpers, and remote-cluster setup.  so if it's not something you want in atomate2, no problem — I'll take it out of this contribution and release it separately as its own repo. Either way, the core SIESTA PR has no database code in it at all and all of the cli staff are self-contained .

4.**Any custodian-related changes should be made in a PR to its repo , not atomate2's**
well this also again is related to  backstory 🙈 🫠 ... I'll mirror atomate2 and refactor it and put it in custodian ... and atomate2 keeps only a thin run_siesta wrapper that imports them.

5.**What purpose do the very specialized workflows like "electrocatalysis" serve here? and 6.You need to explain which workflows, jobs, etc. are added**
 fair question, those things came out of my own research rather than out of any need for a general SIESTA interface.... so I'll take them out of this contribution; the domain-specific ones (electrocatalysis, adsorption, heterostructures) belong in a separate downstream package. For the generic ones that atomate2 already has common templates for (EOS, phonons, elastic, bands, NEB), if you're open to  them later I'd add them as small individual PRs that subclass the existing common flows rather than duplicating them — and nothing specialized goes in the core PR.

7. **Why are so many dataclasses needed to structure I/O with siesta?**
 the 33 dataclasses are a typed model of the SIESTA FDF input space, one class per section of the SIESTA User's Guide. I built them to do four things: validate FDF input against the manual (catch invalid/misspelled keywords and bad values before a run), give users a self-documenting, IDE-discoverable..., drive the tier-preset system and automatic activation of parameter groups, etc ...
That said, I agree it's far heavier than how atomate2 handles input elsewhere — VASP uses a dict-based INCAR with pymatgen's input sets, aims uses pymatgen-io-aims, and neither ships a model of its whole input manual. For the core PR I'll drop that layer entirely and use a conventional lean InputGenerator that produces an FDF parameter dict plus a small validation helper, in keeping with the other codes (and leaning on sisl for file I/O where it helps). The full typed model stays in the standalone tool for the users who want it; it isn't needed for the core SIESTA interface.  

8. **There are changes made to the common part of atomate2 that are unexplained**
And FINALLY  ... 
 you were right that these were bundled in without explanation, and they shouldn't have been part of a SIESTA PR at all. They're not SIESTA-specific; they're general changes that got swept in. There are four files, and they're two independent things:

1. Elastic expansion-order detection (common/analysis/elastic.py new, common/schemas/elastic.py) — the existing code chose 2nd- vs 3rd-order elastic fitting with:

 order = 2 if len(stresses) < 70 else 3  # TODO: Figure this out better.
 
I replaced that raw-count heuristic with estimate_expansion_order(strains), ... etc ... This changes behavior for every code that uses the elastic schema, so it clearly deserves its own PR and a discussion with you rather than riding along silently.

2. Grüneisen bug fix (common/flows/gruneisen.py) — the flow referenced a local bulk that only exists when bulk_relax_maker is set; ...This is an independent bug fix I think , unrelated to SIESTA.

The elastic change I'll propose separately (with the reasoning above and tests) so you can evaluate it on its own merits. The re-scoped SIESTA work will touch no common/ files.

Happy to hear your thoughts on any of the above , especially the docs question (whether you'd want the multilingual support or prefer I revert to the standard single-page setup). Once you've had a chance to comment, Soon I'll start refactoring and open a fresh, minimal SIESTA core PR that's actually reviewable!

PR #1527, opened 2026-08-17
Add run_stats calculations and corresponding tests for FHI-aims

TomatoTurtlez:
## Summary
* Add `_parse_run_stats` helper to `atomate2/aims/schemas/calculation.py` that parses timing, peak memory, and core-count stats directly from `aims.out` (CPU/elapsed time, min/max/avg tracked memory, core count), analogous to VASP's `Outcar.run_stats`.
* Add `run_stats: dict[str, Any] | None` field to `CalculationOutput` (`schemas/calculation.py`), populated in `Calculation.from_aims_files` via `_parse_run_stats`.
* Add `run_stats: dict[str, Any] | None` field to `AimsTaskDoc` (`schemas/task.py`), populated in `from_directory` from `calcs_reversed[-1].output.run_stats` — mirrors VASP's two-level `TaskDoc`/`CalculationOutput` pattern, no re-parsing.
* Fix `_parse_bandstructure` (`schemas/calculation.py`) raising `AttributeError: 'AimsOutput' object has no attribute 'band_structure'` whenever `parse_bandstructure` was truthy. Now guards with `hasattr` and warns + returns `None` if band structure parsing isn't yet supported by the installed `pymatgen.io.aims.outputs.AimsOutput`.

None. Uses `re` and `pathlib.Path`, both already in the standard library and already imported elsewhere in this codebase.

## TODO (if any)
* `run_stats` is parsed directly from `aims.out` text in atomate2 as a stopgap. This should eventually be upstreamed so `pyfhiaims.outputs.stdout.AimsStdout` and `pymatgen.io.aims.outputs.AimsOutput` expose `run_stats` natively (matching how VASP's `Outcar.run_stats` is parsed in pymatgen, not atomate2). Once that lands, `_parse_run_stats` here should be removed and `Calculation.from_aims_files` updated to read `aims_output.run_stats` directly.
* Band structure parsing (`AimsOutput.band_structure`) is not implemented upstream in pymatgen. This PR only prevents the crash when it's requested; actual band structure support is out of scope here and tracked as a separate follow-up.

## Checklist
- [ X] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). The easiest way to handle this is to run the following in the **correct sequence** on your local machine. Start with running [`ruff`](https://docs.astral.sh/ruff) and `ruff format` on your new code. This will automatically reformat your code to PEP8 conventions and fix many linting issues.
- [ X] Doc strings have been added in the [Numpy docstring format] (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html). Run [ruff](https://beta.ruff.rs/docs/rules/#pydocstyle-d) on your code.
- [ X] Type annotations are **highly** encouraged. Run [mypy] (http://mypy-lang.org) to type check your code.
- [ X] Tests have been added for any new functionality or bug fixes.
- [ X] All linting and tests pass.

### DIFF
--- a/src/atomate2/aims/schemas/calculation.py
+++ b/src/atomate2/aims/schemas/calculation.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import json
 import os
+import re
+import warnings
 from collections.abc import Sequence
 from datetime import datetime, timezone
 from pathlib import Path
@@ -85,6 +87,11 @@ class CalculationOutput(BaseModel):
         The valence band maximum in eV (if system is not metallic)
     atomic_steps: list[Structure or Molecule]
         Structures for each ionic step"
+    run_stats (dict[str, float | None]):
+        Various useful run stats including "System time (sec)",
+        "Total CPU time used (sec)", "Minimum memory used (kb)",
+        "Maximum memory used (kb)", "Average memory used (kb)",
+        "cores".
     """
 
     energy: float = Field(
@@ -130,6 +137,9 @@ class CalculationOutput(BaseModel):
     )
     atomic_steps: list[Structure | Molecule] = Field(
         None, description="Structures for each ionic step"
+    )
+    run_stats: dict[str, Any] | None = Field(
+        None, description="Summary of runtime statistics for this calculation"
     )
 
     @classmethod
@@ -345,6 +355,7 @@ class Calculation(BaseModel):
             aims_objects[AimsObject.BANDSTRUCTURE] = bandstructure  # type: ignore  # noqa: PGH003
 
         output_doc = CalculationOutput.from_aims_output(aims_output)
+        output_doc.run_stats = _parse_run_stats(aims_output_file)
 
         has_aims_completed = (
             TaskState.SUCCESS if aims_output.completed else TaskState.FAILED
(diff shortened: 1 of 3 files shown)


JaGeo:
@tpurcell90 would you mind taking a look or refer to a better suited reviewer? If not, I am happy to take over but I am not an aims expert, of course!

tpurcell90:
Ya I'll take a look once my laptop can log in. There send to be technical issues with the servers.

JaGeo:
@tpurcell90 sure!

src/atomate2/aims/schemas/calculation.py
```diff
@@ -390,6 +401,48 @@ def _get_output_file_paths(volumetric_files: list[str]) -> dict[AimsObject, str]
     return output_file_paths
 
 
+def _parse_run_stats(aims_output_file: Path | str) -> dict[str, Any]:
```
tpurcell90, line 404: Why is this function located here and not the pymatgen-io-aims plugin? I think it would be better for parsing of output flies to be done outside of atomate2 as much as possible.

src/atomate2/aims/schemas/calculation.py
```diff
@@ -390,6 +401,48 @@ def _get_output_file_paths(volumetric_files: list[str]) -> dict[AimsObject, str]
...
+    mem_match = re.search(
+        r"Peak values for overall tracked memory usage:\s*"
+        r"\|\s*Minimum:\s*([\d.]+) MB.*?"
+        r"\|\s*Maximum:\s*([\d.]+) MB.*?"
+        r"\|\s*Average:\s*([\d.]+) MB",
+        text,
+        re.DOTALL,
+    )
+
+    return {
```
tpurcell90, line 432: This will go for an entire calculation file or individual steps? This looks like it is for an entire calculation, but it would be good if this is done for each individual calculation

src/atomate2/aims/schemas/task.py
```diff
@@ -522,6 +530,7 @@ def from_directory(
             "tags": tags,
             "completed": calcs_reversed[-1].completed,
             "completed_at": calcs_reversed[-1].completed_at,
+            "run_stats": calcs_reversed[-1].output.run_stats,
```
tpurcell90, line 533: I would make this an optional tag since in high-throughput this can add extra disk space usage and not everyone will need to have this information.

tpurcell90:
Thanks for this pull request! I made some comments, but once resolved it should be good to go

src/atomate2/aims/schemas/calculation.py
TomatoTurtlez: Hi @tpurcell90! Agreed. The current `_parse_run_stats` was a stopgap so I could benchmark a running job before `pyfhiaims`/`pymatgen` upstream support existed (noted in the TODO section of this PR).

I've already extended `pyfhiaims.parsers` (`FinalMemoriesParser` for peak memory, feeding a new `Aims.Stdout.run_stats` property) to replicate this behavior at the source, and plan to open that PR next.

**Question:** Would it make sense for `atomate2` to depend on `pyfhiaims.stdout` directly, rather than going through `pymatgen.io.aims.outputs.AimsOutput`? That would let this land after just one upstream PR (`pyfhiaims`) instead of two (`pyfhiaims` then `pymatgen`).

I know the rest of the AIMS code in `atomate2` currently goes through `pymatgen` exclusively, so I wanted to check whether that's a hard convention here or if a direct `pyfhiaims` dependency would be acceptable.

src/atomate2/aims/schemas/calculation.py
TomatoTurtlez: Good catch. FHI-aims does print per-SCF-iteration timing/memory (`End self-consistency iteration #N`) as well as the SCF-initialization phase, so per-step stats are parseable.

Currently, `run_stats` only captures the final summary once per calculation.

I can add an opt-in flag (e.g., `parse_run_stats="all" | "final" | None`) to additionally capture per-iteration stats as a list, keeping the default behavior as calculation-level only so it doesn't add disk overhead for people who don't need it.

Let me know if that granularity is worth adding now or if the calculation-level summary is sufficient for this PR, and I can scope it as a fast follow-up either way.

src/atomate2/aims/schemas/calculation.py
tpurcell90: We designed the plugin such that the general form is all using the pymatgen framework as much as possible to reduce dependency. It creates problems if one mixes/matches different sources in my opinion.

src/atomate2/aims/schemas/calculation.py
tpurcell90: Adding in the opt-in would be good so it can all be consistent with the other parts of the code

esoteric-ephemera:
Hey @TomatoTurtlez are the review comments addressed?

tpurcell90:
@TomatoTurtlez Just checking to see if you need support in getting these changes done

PR #1530, opened 2026-08-21
Fix `structure` field shadowing `sort_structure` in `VaspInputGenerator`

### LINKED ISSUE #1529: BUG: `VaspInputGenerator.structure` field shadows `sort_structure` property, breaking MD (npt) LANGEVIN_GAMMA
**Describe the bug**
`VaspInputGenerator` (`atomate2/vasp/sets/base.py`) redeclares `structure` as a plain dataclass field. This shadows (overrides) the `structure` property pymatgen's `VaspInputSet` defines for that same name — the property whose setter applies `sort_structure`, `reduce_structure`, and `validate_magmom`. Because the child class's plain field is found first, `self.structure = ...` never reaches that setter, so structures are not sorted for every atomate2 VASP input-set generator (relax, static, MD, ...).

This is invisible for generators whose INCAR doesn't depend on POSCAR atom-type ordering. It is not invisible for `MDSetGenerator`'s `npt` ensemble: `LANGEVIN_GAMMA` is sized to the number of *distinct elements*, but VASP counts POSCAR "types" as *contiguous same-element runs*. If the input structure isn't already grouped by element, the two counts diverge and VASP aborts immediately with:
```
Error reading item LANGEVIN_GAMMA from file INCAR. IERR= 5
I REFUSE TO CONTINUE WITH THIS SICK JOB ... BYE!!!
```

**To Reproduce**
1. Build a `Structure` whose same-element sites are *not* contiguous (e.g. built cluster-by-cluster rather than grouped by element).
2. Pass it to `MDSetGenerator(ensemble="npt").get_input_set(structure, potcar_spec=True)`.
3. Compare `len(vasp_input["POSCAR"].natoms)` (atom-type count VASP will see) to `len(vasp_input["INCAR"]["LANGEVIN_GAMMA"])` (values VASP will get) — they differ.
4. Running the generated input through VASP produces the fatal error above.

A minimal, runnable repro is attached: [shadow_example.py](https://github.com/user-attachments/files/31317670/shadow_example.py). It asserts the two counts are equal — it fails today, and would pass if `sort_structure` actually ran.

**Expected behavior**
With `sort_structure=True` (the default), the structure should be sorted (grouping same-element sites into contiguous POSCAR blocks) before POSCAR/INCAR generation, exactly as pymatgen's `VaspInputSet` docstring promises. `LANGEVIN_GAMMA`'s length should then always match VASP's parsed atom-type count.

**Screenshots**
Not applicable (not a UI issue) — the VASP fatal error text above is the observable symptom.

ThomasWarford:
## Summary

* `VaspInputGenerator` redeclared `structure` as a plain dataclass field, overriding the `structure` property (with its `sort_structure`/`reduce_structure`/`validate_magmom` setter) inherited from pymatgen's `VaspInputSet`. As a result, `sort_structure` silently never ran for any atomate2 VASP input-set generator.
* This surfaces destructively in `MDSetGenerator`'s `npt` ensemble: `LANGEVIN_GAMMA` is sized to the number of distinct elements, while VASP counts POSCAR "types" as contiguous same-element runs. With an unsorted structure the two counts diverge and VASP aborts with `Error reading item LANGEVIN_GAMMA from file INCAR`.
* Fix: remove the redundant field declaration so the inherited property is no longer shadowed.
* Added regression tests (`tests/vasp/test_sets.py`) and a standalone repro (`example.py`).

**Alternatively, we modify MDSetGenerator something like:**
```python
class MDSetGenerator(VaspInputGenerator):
    def get_input_set(self, structure=None, prev_dir=None, potcar_spec=False):
        if structure is not None and self.sort_structure:
            structure = structure.get_sorted_structure()
        return super().get_input_set(structure, prev_dir=prev_dir, potcar_spec=potcar_spec)
```
**for a more targeted fix.**

* None.

## TODO (if any)

* None.

## Checklist

title.

  type check your code.

## Issue
Closes #1529. See this issue for bug reproduction instructions.

### DIFF
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -173,8 +173,6 @@ class VaspInputGenerator(VaspInputSet):
             from_prev_calc.
     """
 
-    # Do not redeclare `structure` here: VaspInputSet so setting structure
-    # uses pymatgen.io.vasp.sets setter which sorts the structure
     config_dict: dict = field(default_factory=lambda: _BASE_VASP_SET)
     files_to_transfer: dict = field(default_factory=dict)
     user_incar_settings: dict = field(default_factory=dict)
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -174,7 +174,8 @@ class VaspInputGenerator(VaspInputSet):
             from_prev_calc.
     """
 
-    structure: Structure | None = None
+    # Do not redeclare `structure` here: VaspInputSet so setting structure
+    # uses pymatgen.io.vasp.sets setter which sorts the structure
     config_dict: dict = field(default_factory=lambda: _BASE_VASP_SET)
     files_to_transfer: dict = field(default_factory=dict)
     user_incar_settings: dict = field(default_factory=dict)
--- a/src/atomate2/vasp/sets/base.py
+++ b/src/atomate2/vasp/sets/base.py
@@ -10,7 +10,6 @@ from pymatgen.io.vasp.sets import MPRelaxSet, MPScanRelaxSet, VaspInputSet
 from atomate2 import SETTINGS
 
 if TYPE_CHECKING:
-    from pymatgen.core import Structure
     from pymatgen.io.vasp import Kpoints
     from pymatgen.io.vasp.sets import UserPotcarFunctional
 
--- a/tests/vasp/test_sets.py
+++ b/tests/vasp/test_sets.py
@@ -1,7 +1,8 @@
 import pytest
 from pymatgen.core import Lattice, Species, Structure
-from pymatgen.io.vasp.sets import MPScanRelaxSet
+from pymatgen.io.vasp.sets import MPScanRelaxSet, VaspInputSet
 
+from atomate2.vasp.sets.base import VaspInputGenerator
 from atomate2.vasp.sets.core import (
     ElectronPhononSetGenerator,
     HSEBSSetGenerator,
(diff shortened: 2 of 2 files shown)


JaGeo:
@esoteric-ephemera do you have time to look into this? I don't have much experience with the MD workflow

src/atomate2/vasp/sets/base.py
```diff
@@ -174,7 +173,8 @@ class VaspInputGenerator(VaspInputSet):
             from_prev_calc.
     """
 
-    structure: Structure | None = None
+    # Do not redeclare `structure` here: VaspInputSet so setting structure
+    # uses pymatgen.io.vasp.sets setter which sorts the structure
```
esoteric-ephemera: remove the ai-generated comment please

esoteric-ephemera:
Thanks @ThomasWarford. I'm not sure the analysis is quite right, since `VaspInputSet` itself defines `structure` as an [init field](https://github.com/materialsproject/pymatgen-core/blob/3affe4635b1772421ef110c2e270b1e1ad282078/src/pymatgen/io/vasp/sets.py#L223), and then later defines it as a property with a setter. My guess is that there's some unhealthy handling/collision of fields in the upstream model that need to be changed

That being said, it is redundant to redeclare the field here so I'm fine with this change

Guessing `example.py` was an AI-generated MWE?

ThomasWarford:
`example.py` was AI generated yes. It's quite verbose sorry. Here's a simpler demonstration which I've read and cut down more closely.

It looks like redefining structure as a field in `VaspInputGenerator` means the setter function defined in `VaspInputSet` is no longer used.

```python
from pathlib import Path
from atomate2.vasp.sets.core import MDSetGenerator

defining_class = next(
    cls for cls in MDSetGenerator.__mro__ if "structure" in cls.__dict__
)
structure_attr = defining_class.__dict__["structure"]

print(f"`structure` is defined on: {defining_class.__module__}.{defining_class.__qualname__}")
print(f"`structure` is a property: {isinstance(structure_attr, property)}")
print(f"`structure` has a setter: {isinstance(structure_attr, property) and structure_attr.fset is not None}")
```

---

Before removing `structure` field definition from `VaspInputGenerator`:
```bash
(dft) twarford:~/dev/atomate2> python setter_example.py
`structure` is defined on: pymatgen.io.vasp.sets.VaspInputSet
`structure` is a property: True
`structure` has a setter: True
```
After removing `structure` field from `VaspInputGenerator`:
```bash
(dft) twarford:~/dev/atomate2> python setter_example.py
`structure` is defined on: atomate2.vasp.sets.base.VaspInputGenerator
`structure` is a property: False
`structure` has a setter: False
```


Without the setter function sorting based on elements, we get the bug in issue #1530.

src/atomate2/vasp/sets/base.py
ThomasWarford: Sorry that comment doesn't read well. I wanted to leave something there so it is clear `structure` can be given as an argument to the constructor, but I suppose the docstring does that. If there's another way you can think of to make it clear to the user, I'm happy to add it.

ThomasWarford: Comment removed.

esoteric-ephemera:
Gotcha, thanks for the analysis and fix @ThomasWarford! Merging asap

PR #1537, opened 2026-08-26
Amset interpolation convergence

Jackbt21:
`copy_amset_files()` currently looks for a file named `transport.json`. AMSET writes the transport files as `transport_{mesh}.json` by default. Because that file is never found, `transport.prev.json` is never created, so `check_converged()` has nothing to compare against on a resubmission, leaving jobs resubmitting weather the interpolation factor is converged or not. 

## Summary

- Replaces the exact match lookup of `transport.json` with the wildcard pattern `transport_*.json*`, matching the default file naming.
- `rename_files()` doesn't support wildcards, so the transport file is renamed with a plain pathlib glob + rename instead of going through `rename_files()`.
- Added `allow_missing=True` into `copy_files`, replacing what was in `rename_files`, to avoid crashing on the first run when no previous file exists.

### DIFF
--- a/src/atomate2/amset/files.py
+++ b/src/atomate2/amset/files.py
@@ -3,17 +3,13 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
-
-from monty.serialization import loadfn
-
-from atomate2 import SETTINGS
-from atomate2.common.files import copy_files, get_zfile, gunzip_files, rename_files
-from atomate2.utils.file_client import FileClient, auto_fileclient
-from atomate2.utils.path import strip_hostname
-
 from pathlib import Path
 
+from atomate2 import SETTINGS
+from atomate2.common.files import copy_files, get_zfile, gunzip_files
+from atomate2.utils.file_client import FileClient, auto_fileclient
+from atomate2.utils.path import strip_hostname
+from monty.serialization import loadfn
 
 logger = logging.getLogger(__name__)
 
--- a/src/atomate2/amset/files.py
+++ b/src/atomate2/amset/files.py
@@ -12,7 +12,6 @@ from atomate2.common.files import copy_files, get_zfile, gunzip_files, rename_fi
 from atomate2.utils.file_client import FileClient, auto_fileclient
 from atomate2.utils.path import strip_hostname
 
-
 from pathlib import Path
 
 
--- a/src/atomate2/amset/files.py
+++ b/src/atomate2/amset/files.py
@@ -5,11 +5,12 @@ from __future__ import annotations
 import logging
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


JaGeo:
@utf can you please check?

src/atomate2/amset/files.py
```diff
@@ -74,7 +71,9 @@ def copy_amset_files(
         file_client=file_client,
     )
 
-    rename_files({"transport.json": "transport.prev.json"}, allow_missing=True)
+    local_transport = next(Path().glob("transport_*.json*"), None)
```
esoteric-ephemera, line 74: Would a safer wildcard pattern be `"transport*.json"` so the previous matches would still be there (if applicable)?

src/atomate2/amset/files.py
```diff
@@ -55,17 +51,18 @@ def copy_amset_files(
         "band_structure_data.json",
         "wavefunction.h5",
         "deformation.h5",
-        "transport.json",
     ):
         found_file = get_zfile(directory_listing, file, allow_missing=True)
         if found_file is not None:
             files.append(found_file)
+    files.append(Path("transport_*.json*"))
```
esoteric-ephemera, line 58: Change to this? Calling `Path` with a regex interprets the symbol literally, AFAIK
```py
files.extend(list(Path(".").glob("transport*.json")))
```

src/atomate2/amset/files.py
Jackbt21: In any of the later iterations (from the 3rd onwards). The previous directory will contain `transport.prev.json*` and `transport_{mesh}.json*`. The underscore is here to differentiate between the two so the most recent transport file is selected and not the old one.

src/atomate2/amset/files.py
Jackbt21: This change was introduced to satisfy the `mypy` type checking in the test suite, which expects `files` to contain either `Path` or `None`. The path stored inside the `Path` object is later treated as a globbable string, rather than being subject to any of the class methods, so the `Path` is essentially just a wrapper.

Alternatively, I found that the `mypy` type checking can be satisfied by adding a type hint and appending `"transport_*.json*"` as a string:

```
    # find optional files
    files: list[str | Path] = []
    for file in (
        "settings.yaml",
        "vasprun.xml",
        "band_structure_data.json",
        "wavefunction.h5",
        "deformation.h5",
    ):
        found_file = get_zfile(directory_listing, file, allow_missing=True)
        if found_file is not None:
            files.append(found_file)

    files.append("transport_*.json*")
```

Either way, the wildcards are handled later in `common.files.find_and_filter_files`, as called by `copy_files` and `gunzip_files`.

I am happy to go with either solution.

PR #1544, opened 2026-09-10
Re-attach structure properties and support D3 dispersion in TorchSim

akwarii:
## Summary

- Preserve `Structure.properties` and `Structure.site_properties` across TorchSim runs. TorchSim's SimState conversions (`ts.io.structures_to_state()` / `state.to_structures()`) currently drops both, so properties set on the input structures are re-attached onto the output ones once the calculation finishes.

- Add support for a D3 dispersion correction on any TorchSim model. Passing `dispersion=True` (plus `a1`, `a2`, `s8`, and a `d3_params` `D3Parameters` instance, with `s6` optional) to a maker's model_kwargs now builds a `D3DispersionModel` and sums it with the base model via `SumModel`. The dispersion correction intentionally does not inherit the base model's cutoff, since D3's long-range tail needs a much larger cutoff than typical short-range ML potentials.

None of these changes are breaking. 

None.

## Note

The functional-dependent BJ damping parameters (a1, a2, s8) and the element reference data (d3_params) must currently be supplied explicitly by the caller; there is no bundled lookup table for common XC functionals (e.g. from https://github.com/dftd3/simple-dftd3/blob/main/assets/parameters.toml).

## Checklist

title.

  type check your code.


akwarii:
The test that failed seems to be related to the recent bump of phonopy to v4 and not to this PR.

JaGeo:
Could you restrict phonopy for now? Thanks a lot!

JaGeo:
@akwarii thanks again. Could you please check if the tests are really running as part of the CI? At least I couldn't find them in the torch-sim tests. Only flows seem to have run as part of this PR. I am sorry if I have overlooked anything.

akwarii:
The `test-torchsim` block of the CI is running ` pytest --durations=5 -n auto tests/torchsim --cov=atomate2 --cov-report=xml`. There is is also nothing that should make the tests being skipped, as the only skip condition is torch_sim being installed (and mace for some tests, which is also installed).

One thing that is currently skipped in `test-torchsim`, and must be added to the CI, is the models API compatibility. The tests are present but are currently being skipped. A new test matrix should probably be created. As it is out of the scope of this PR I believe it should be added in another one. That being said, I can open this new PR but I'm far from being a CI expert.

JaGeo:
@akwarii if you think you can open it, it would be highly appreciated. I am not sure when I can take care of it. This would ensure a timely merge

akwarii:
I opened #1549. It should consolidate the CI pipeline at a minimal cost but a few minimal changes might be needed in this PR once #1549 is merged.

JaGeo:
@akwarii Happy to merge once the conflicts are resolved!

akwarii:
@JaGeo everything should be ready now :)

JaGeo:
One point: you are referring to the fact in your text above that the neighbor cutoff of the D3 interaction is not inherited from the ML potential, right? This is fine. I mainly wanted to make sure I understood it.

akwarii:
Yes, I don't inherit from the model cutoff since we don't want D3 interactions to have a e.g., 6A cutoff. I didn't add a custom keyword that can be used to change the D3 cutoff but that's straightforward to implement if needed.

JaGeo:
@akwarii is there an explicit test of the new SumModel? I did not see this yet. I believe it would be imprtant.

akwarii:
You're absolutely right, it was removed when #1549 was merged since D3 was not implemented yet and I forgot to add it back - I will commit the test in a minute

JaGeo:
Thanks! The test is still skipped, I believe.

akwarii:
From what I see, 2 tests are being consistently skipped: `test_pick_model_mattersim` and `test_pick_model_fairchem`. The first one is because mattersim was disabled due to their dependency mess and the latter is due to access restrictions.

JaGeo:
I am sorry, but I don't see that your new test "test_pick_model_dispersion" ran. Are you missing an installation?
image

akwarii:
What you are showing here is the breakdown of the 5 slowest test and that 8 tests are being skipped in test-torchsim-generic. However, if you look at the logical OR of the set of tests that ran in -generic and -mace, then only 2 are never executed, namely fairchem and mattersim

JaGeo:
You are right. My mistake! Thanks!

JaGeo:
@akwarii thanks a lot!

PR #1549, opened 2026-09-15
Run TorchSim model backend tests in CI

akwarii:
## Summary

* The `test-torchsim` CI job only ever installed `mace-torch` alongside `torch-sim`, so every `test_pick_model_*` test for the other backends (fairchem, mattersim, metatomic, nequip, orb, sevenn) was silently skipped in CI.
* These backends pin conflicting `e3nn` versions (mace-torch requires `e3nn==0.4.4`, fairchem/nequip/sevenn require `e3nn>=0.5`) and can't all be installed in the same environment, so they're split into two new `pyproject.toml` extras (`torchsim-mace`, `torchsim-generic`) and run as a CI matrix, mirroring the existing `test-forcefields` dep-group split.
* Bump `torch-sim-atomistic` 0.6.0 -> 0.6.1.
* Fix `pick_model`'s `MATTERSIM` branch: it was passing the raw checkpoint path straight to `MatterSimModel`, which expects an already-loaded `mattersim.forcefield.potential.Potential` instance.
* mattersim is left out of `torchsim-generic` for now: `mattersim>=1.2.5` (the minimum version `torch-sim-atomistic` 0.6.1 pulls in) renamed `datasets.utils.convertor` to `datasets.utils.converter`, which breaks `torch_sim.models.mattersim` at import time. Can be added back once that's sorted out upstream.
* Consolidate `tests/torchsim/test_models.py` into `tests/torchsim/test_core.py`.
* Add a missing `_SKIP_MACE` marker to `test_torchsim_phonon_maker_integration`, which unconditionally uses MACE but was never exercised without `mace-torch` installed before.

`torch-sim-atomistic` extras used here (`fairchem`, `mace`, `metatomic`, `nequip`, `orb`, `sevenn`) pull in the corresponding MLIP packages, all of which were already optional dependencies for TorchSim model backends.

## TODO (if any)

* mattersim is supported but disabled until they clean up their dependencies.

## Checklist

* [x] Code is in the standard Python style (`ruff` / `ruff format` clean).
* [x] Type annotations included where changed.
* [x] Tests exercised locally against fresh installs of both new dependency groups.

Co-authored by Claude Code.


JaGeo:
@akwarii thanks a lot! I hope this will now merge automatically!

PR #1550, opened 2026-09-16
Enhance dev tutorial with Maker example and link missing tutorials

akwarii:
## Summary

* Link orphaned tutorial notebooks (`lammps_workflow`, `materials_project_workflows`, `torchsim_tutorial`) into the `tutorials/tutorials.md` ToC so they are discoverable from the built documentation.
* Resolve the two open TODOs in `docs/dev/workflow_tutorial.md` by replacing the ambiguous placeholder code with a concrete, generic `ExampleMaker` example that shows how input-set writing, running the underlying code, and parsing a `TaskDocument` all fit together inside a single `@job`-decorated `make()` method, with a pointer to `BaseVaspMaker.make` as a real, more elaborate reference implementation.

None.

## TODO (if any)

None — this is docs-only and complete.

## Checklist

* [x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/) (docs-only change, no Python code added).
* [x] Doc strings have been added in the [Numpy docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) (n/a — no new functions/classes).
* [x] Type annotations are **highly** encouraged (n/a — no new code).
* [x] Tests have been added for any new functionality or bug fixes (n/a — documentation-only change).


docs/dev/workflow_tutorial.md
```diff
@@ -8,45 +8,51 @@ In the context of computational materials science, `Flow` objects are most easil
...
-    def get_input_set(self, coordinates: Structure) -> InputSet:
-        # take the input coordinates, determine appropriate
+    def get_input_set(self, structure: Structure) -> InputSet:
+        # take the input structure, determine appropriate
         # input file contents, and return an `InputSet`
         return InputSet(...)
 ```
 
-`pymatgen` already contains `InputSet` for many common codes, so when developing a workflow `Maker` it is convenient to use the `InputGenerator` / `InputSet` to prepare your files. This is done in `atomate2` by making the `InputGenerator` a class parameter, e.g.,
+`pymatgen` already contains `InputSet` for many common codes, so when developing a workflow `Maker` it is convenient to use the `InputGenerator` / `InputSet` to prepare your files. This is done in `atomate2` by making the `InputGenerator` a class parameter, and writing the input files inside a `@job`-decorated function (rather than directly in `make()`), since it is the individual `Job` that actually gets executed, potentially on a remote compute resource.
```
JaGeo: That's potentially confusing and reads wrong. The make typcially returns a Job object or has a job decorator. It is def not part of the init of a Maker, yes. Maybe, one does want an additional job to run the calculation but then I would be more specific with an example.

JaGeo:
Thanks! I was a bit confused here. Potentially, adding the write_input method clearly to one of the classes would make it clearer. And there is def something weird in how it is phrased right now.

docs/dev/workflow_tutorial.md
akwarii: I fully agree, this part is quite confusing. I will update that.

akwarii:
I will try to improve the part you pointed out above and see how I can better explain the difference between `InputSet` and `InputGenerator` by adding a `ExampleInputSet` with a `write_input()`

JaGeo:
@akwarii thanks a lot!

docs/dev/workflow_tutorial.md
```diff
@@ -8,45 +8,64 @@ In the context of computational materials science, `Flow` objects are most easil
...
 ```py
 class ExampleMaker(Maker):
-    def make(self, coordinates: Structure) -> Flow:
-        # take the input coordinates and return a `Flow`
+    def make(self, structure: Structure) -> Flow:
+        # take the input structure and return a `Flow`
         return Flow(...)
 ```
 
 The `Maker` class usually contains most of the calculation parameters and other settings that are required to set up the calculation in the correct way. Much of this logic can be written like normal python functions and then turned into a `Job` via the `@job` decorator.
```
JaGeo:
```suggestion
The `Maker` class usually contains most calculation parameters and other settings needed to set up the calculation correctly. Much of this logic can be written within methods and then turned into a ' Job ' via the `make` method and the `@job` decorator, or into a `Flow` via the `make` method and by returning a `Flow` object.
```

docs/dev/workflow_tutorial.md
```diff
@@ -8,45 +8,64 @@ In the context of computational materials science, `Flow` objects are most easil
...
 
-One common task encountered in almost any materials science calculation is writing calculation input files to disk so they can be executed by the underlying software (e.g., VASP, Q-Chem, CP2K, etc.). This is preferably done via a `pymatgen` `InputSet` class. `InputSet` is essentially a dict-like container that specifies the files that need to be written, and their contents. Similarly to the way that `Maker` classes generate `Flow`s, `InputSet`s are most easily created by `InputGenerator` classes. `InputGenerator`
-have a method `get_input_set()` that typically takes atomic coordinates (e.g., a `Structure` or `Molecule` object) and produce an `InputSet`, e.g.,
+One common task encountered in almost any materials science calculation is writing calculation input files to disk so they can be executed by the underlying software (e.g., VASP, Q-Chem, CP2K, etc.). This is preferably done via a `pymatgen` `InputSet` class. `InputSet` is essentially a dict-like container that specifies the files that need to be written, and their contents. and has a `write_input()` method that writes those files to disk, e.g.,
```
JaGeo:
```suggestion
One common task in almost any materials science calculation is writing calculation input files to disk so the underlying software (e.g., VASP, Q-Chem, CP2K, etc.) can execute them. This is preferably done via a `pymatgen` `InputSet` class. `InputSet` is essentially a dict-like container that specifies which files to write and their contents. It has a `write_input()` method that writes those files to disk, e.g.,
```

docs/dev/workflow_tutorial.md
```diff
@@ -8,45 +8,64 @@ In the context of computational materials science, `Flow` objects are most easil
...
+    def __init__(self, inputs: dict[PathLike, InputFile]):
+        self.inputs = inputs
+
+    def write_input(self, directory: str | Path) -> None:
+        # write the files held by this `InputSet` to `directory`
+        for path, file in self.inputs.items():
+            file.write(directory / path)
+```
+
+Similarly to the way that `Maker` classes generate `Flow`s, `InputSet`s are most easily created by `InputGenerator` classes. `InputGenerator` classes have a method `get_input_set()` that typically takes atomic coordinates (e.g., a `Structure` or `Molecule` object) and produce an `InputSet`, e.g.,
```
JaGeo:
```suggestion
Similarly to the way that `Maker` classes generate `Flow`s and `Job`s, `InputSet`s are most easily created by `InputGenerator` classes. `InputGenerator` classes have a method `get_input_set()` that typically takes atomic coordinates (e.g., a `Structure` or `Molecule` object) and produces an `InputSet`, e.g.,
```

docs/dev/workflow_tutorial.md
```diff
@@ -8,45 +8,64 @@ In the context of computational materials science, `Flow` objects are most easil
...
-        # take the input coordinates, determine appropriate
+    def get_input_set(self, structure: Structure) -> ExampleInputSet:
+        # take the input structure, determine appropriate
         # input file contents, and return an `InputSet`
-        return InputSet(...)
+        return ExampleInputSet(...)
 ```
 
-`pymatgen` already contains `InputSet` for many common codes, so when developing a workflow `Maker` it is convenient to use the `InputGenerator` / `InputSet` to prepare your files. This is done in `atomate2` by making the `InputGenerator` a class parameter, e.g.,
+`pymatgen` already contains `InputGenerator` / `InputSet` pairs for many common codes, so when developing a workflow `Maker` it is convenient to reuse them to prepare your files. The `Maker` holds an `InputGenerator` as a class parameter and calls `get_input_set()` / `write_input()` from within its `make()` method.
```
JaGeo:
```suggestion
`pymatgen` already contains `InputGenerator` / `InputSet` pairs for many common codes, so when developing a workflow `Maker`, it is convenient to reuse them to prepare your files. The `Maker` holds an `InputGenerator` as a class parameter and calls `get_input_set()` / `write_input()` from within its `make()` method.
```

docs/dev/workflow_tutorial.md
```diff
@@ -8,45 +8,64 @@ In the context of computational materials science, `Flow` objects are most easil
...
 
-**TODO - the code block below needs refinement. Not exactly sure how write_inputs() fits into a`Job`**
+Putting this together, a `Maker`'s `make()` method is itself decorated with `@job`, which turns `make()` into the `Job` that actually gets executed, potentially on a remote compute resource — this is why any code that touches the filesystem or launches an external process must live inside `make()` (or another `@job`-decorated function it calls), rather than elsewhere in the `Maker`. A typical `make()` will (1) write the input files, (2) run the underlying code, and (3) parse the output directory into a `TaskDocument`, which is returned as the `Job`'s output via a `Response`:
```
JaGeo:
```suggestion
Putting this together for the specific case of executing a simulation code, a `Maker`'s `make()` method is itself decorated with `@job`, which turns `make()` into the `Job` that actually gets executed, potentially on a remote compute resource — this is why any code that touches the filesystem or launches an external process must live inside `make()` (or another `@job`-decorated function it calls), rather than elsewhere in the `Maker`. A typical `make()` for such a task will (1) write the input files, (2) run the underlying code, and (3) parse the output directory into a `TaskDocument`, which is returned as the `Job`'s output via a `Response`:
```

JaGeo:
@akwarii Can you check if this makes sense to you now?

docs/dev/workflow_tutorial.md
akwarii: Much better! Thanks! Just a small question about this part, 

> Much of this logic can be written within methods [...]

I read that part as "all the sub jobs of `make` should be methods of the `Maker` and not independent functions", which is not what we do. In practice, the jobs are almost always independent functions with a `job` decorator that are called in `make`.

What do you think?

JaGeo: Fine with me. I would say methods of the Maker or functions

akwarii:
```suggestion
The `Maker` class usually contains most calculation parameters and other settings needed to set up the calculation correctly. Much of this logic can be written within methods of the `Maker` or independent functions and then turned into a ' Job ' via the `make` method and the `@job` decorator, or into a `Flow` via the `make` method and by returning a `Flow` object.
```

akwarii:
I would say the current state of the PR is good and can be merged, what do you think @JaGeo ?

JaGeo:
@akwarii thank you!

